PR #24, opened 2024-01-24
Refactor Demo project to embed the package instead of using it as a local package

andrewdmontgomery:
## What this does

Attempting to make this project dependent on a local package presents some problems:
- New source files in the package aren't available without closing/opening the project
- Source files in the package cannot be deleted from within the project

This may be because Local Packages are meant to override existing Package dependencies, instead of existing as first-class citizens.

We are exploring Apple's recommended approach, which is to add the package as a regular dependency, and then add the local package (which would override the dependency for local development).  We've run into errors.

So while we explore that, this PR makes an adjustment that should work:
- Removes the local package
- Adds the package to the project
- Adds the "product" of the package (the Gravatar library) to each of the targets (which embeds it in the target and links it)

## Testing

- [ ] Open the Demo project, add a new source file (and a new type) to the Package `Sources > Gravatar`, and confirm that you can reference that new type from the Demo app
- [ ] Confirm that you can build you changes
- [ ] Confirm that you can delete files from the Package in Xcode

### DIFF
--- a/Demo/Gravatar-Demo.xcodeproj/project.pbxproj
+++ b/Demo/Gravatar-Demo.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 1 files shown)


pinarol (approved):
Thanks! Looks good. I didn't have the same issue with this config.

PR #100, opened 2024-03-06
Rename defaultImage as defaultImageOption

### LINKED ISSUE #99: Rename "defaultImage" as "defaultImageOption"
Rename "defaultImage" as "defaultImageOption". defaultImage sounds more like `UIImage`.

pinarol:
Closes #99 

`defaultImage` sounds like a `UIImage` type of parameter. I think it's better if we call this `defaultImageOption` since its type is actually `DefaultImageOption`. WDYT?

WordPressUI PR: https://github.com/wordpress-mobile/WordPressUI-iOS/pull/149


etoledom (approved):
Nice! Good call

PR #118, opened 2024-03-18
Profile service parity with Android

### LINKED ISSUE #112: Improve `ProfileService`
Public API should look like:

```swift
fetch(with email: Email) async throws -> UserProfile // Email is String alias
fetch(with hash: Hash) async throws -> UserProfile // Hash is String alias
fetch(with userName: UserName) async throws -> UserProfile // Username is String alias
```

etoledom:
Closes #112 

### Description

This PR changes the public API of `ProfileService` to be in parity with Android SDK.

I have added a new `protocol ProfileFetching`, which I hace kept internal since it doesn't have any practicality for users.
I have tested that the documentation declared in the protocol is shown on the docs of `ProfileService` even if the protocol is not public 👍 

CleanShot 2024-03-18 at 21 11 36@2x

#### String aliases

I have created the String aliases for `Email`, `Hash`, and `Username`.  The method signature though must have a different label, since the parameter is still basically a `String` and the compiler can't differentiate from them.

I did a fast test declaring these as `struct`s, conforming to `ExpressibleByStringLiteral`. It works well when the parameter given is a string literal, but not when it's a variable (it would look like: `fetch(with: Email(stringLiteral: emailVariable))` which lacks practicality).

And having Email, Hash, and UserName structs conforming to `ExpressibleByStringLiteral` will rise the same ambiguity issue, so at the end this approach is not practical.

At the end, we are not gaining anything more than a bit more explicit method signature, but I'm leaving them as we agreed already to have them, and I don't see any harm on it.

#### Deprecating `completionHandler`.

This is not a necessity, but I'd argue that we either give all methods a `completionBlock` counterpart, or we don't provide any.

WPiOS uses the completion block method, and that's why I've added a deprecation warning instead of removing it directly, but we would need to eventually change it for the concurrency counterpart.

I advocate to not support completion blocks, except when it's more necessary like in View classes.

### Testing Steps

- CI should be 🍏

### DIFF
--- a/Sources/Gravatar/Network/Services/Aliases.swift
+++ b/Sources/Gravatar/Network/Services/Aliases.swift
@@ -1,6 +1,0 @@
-/// A String representing an email.
-public typealias Email = String
-/// A String representing the hash of an email.
-public typealias Hash = String
-/// A String representing a user name.
-public typealias UserName = String
--- a/Sources/Gravatar/Network/Services/Aliases.swift
+++ b/Sources/Gravatar/Network/Services/Aliases.swift
@@ -1,0 +1,6 @@
+/// A String representing an email.
+public typealias Email = String
+/// A String representing the hash of an email.
+public typealias Hash = String
+/// A String representing a user name.
+public typealias UserName = String
--- a/Sources/Gravatar/Network/Services/ProfileFetching.swift
+++ b/Sources/Gravatar/Network/Services/ProfileFetching.swift
@@ -2,16 +2,16 @@ protocol ProfileFetching {
     /// Fetches a Gravatar user's profile information, and delivers the user profile asynchronously.
     /// - Parameter email: The user account email.
     /// - Returns: An asynchronously-delivered user profile.
-    func fetch(withEmail email: Email) async throws -> UserProfile
+    func fetch(withEmail email: String) async throws -> UserProfile
 
     /// Fetches a Gravatar user's profile information, and delivers the user profile asynchronously.
     /// - Parameter hash: The user's email sha256 hash which corresponds to their account. For more info, check [the gravatar
     /// documentation](https://docs.gravatar.com/general/hash/).
     /// - Returns: An asynchronously-delivered user profile.
-    func fetch(withHash hash: Hash) async throws -> UserProfile
+    func fetch(withHash hash: String) async throws -> UserProfile
 
     /// Fetches a Gravatar user's profile information, and delivers the user profile asynchronously.
     /// - Parameter userName: The user's account user name.
     /// - Returns: An asynchronously-delivered user profile.
-    func fetch(withUserName userName: UserName) async throws -> UserProfile
+    func fetch(withUserName userName: String) async throws -> UserProfile
 }
--- a/Sources/Gravatar/Network/Services/ProfileFetching.swift
+++ b/Sources/Gravatar/Network/Services/ProfileFetching.swift
@@ -1,0 +1,17 @@
+protocol ProfileFetching {
+    /// Fetches a Gravatar user's profile information, and delivers the user profile asynchronously.
+    /// - Parameter email: The user account email.
+    /// - Returns: An asynchronously-delivered user profile.
+    func fetch(withEmail email: Email) async throws -> UserProfile
+
+    /// Fetches a Gravatar user's profile information, and delivers the user profile asynchronously.
+    /// - Parameter hash: The user's email sha256 hash which corresponds to their account. For more info, check [the gravatar
+    /// documentation](https://docs.gravatar.com/general/hash/).
+    /// - Returns: An asynchronously-delivered user profile.
+    func fetch(withHash hash: Hash) async throws -> UserProfile
+
+    /// Fetches a Gravatar user's profile information, and delivers the user profile asynchronously.
+    /// - Parameter userName: The user's account user name.
+    /// - Returns: An asynchronously-delivered user profile.
+    func fetch(withUserName userName: UserName) async throws -> UserProfile
+}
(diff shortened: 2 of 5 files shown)


pinarol:
> I have created the String aliases for Email, Hash, and Username

I didn't understand the need behind this one. To me this makes the code more difficult to read. At the first glance it looks like different type like a struct but it is not, so now i need to make an effort to discover the real type. 🤔

pinarol:
> This is not a necessity, but I'd argue that we either give all methods a completionBlock counterpart, or we don't provide any.

I am ok with not providing any. Seems better than starting the framework with deprecated methods. Maybe we can keep the one in UIImageView extension though, I haven't thought about that one throughly.

etoledom:
> I didn't understand the need behind this one. To me this makes the code more difficult to read. At the first glance it looks like different type like a struct but it is not, so now i need to make an effort to discover the real type. 🤔

A bit more explanation of the reasoning: We agreed on doing this while working on iOS-Android parity, and on paper it seemed to be a good improvement over method signature. But now that I see it in practice, it does little to nothing good (or could even make it worse as you mention).

I have added it to this PR since we agreed on it, but as I mentioned on the description of the PR, I don't see any gains on it, and now that you mention the extra steps on the user side to figure out what's the type itself, it seems like a drawback more than anything good.

After seeing it, I'd pretty much agree to better keep `Strings` instead of using the aliases.

cc @andrewdmontgomery - since we talked about this.

etoledom:
> I am ok with not providing any. Seems better than starting the framework with deprecated methods. Maybe we can keep the one in UIImageView extension though, I haven't thought about that one throughly.

Do you mean to remove `fetchProfile(with:onCompletion:)` now also?
This is giving us a warning on Unit Tests, and is not letting passing CI 😬 
I'd be happy to remove it if we agree on it, though we will need to update the usage on WPiOS with `Task` (which is also fine)

**About the UI components:**

I'm not sure it's good idea to provide async functions on UI methods, though I honestly don't know if we shouldn't (I was looking for an example in UIKit, but haven't found any yet).

My reasoning is that UI classes should do their work on the main thread, and methods (in theory) on those clases should also be called on the main thread. 

Since we need to do tasks on background thread internally, the only ways to achieve this from the main thread is with either completion blocks or a delegate callback.

So based on this, I'd agree on keeping completion block on the `UIImageView` extension and on any future UI Component, regardless of how we do the work internally.

pinarol:
> Do you mean to remove fetchProfile(with:onCompletion:) now also?

I was thinking we would make separate PRs for those but I am fine if you wanna do it in this PR too. 

> So based on this, I'd agree on keeping completion block on the UIImageView extension and on any future UI Component, regardless of how we do the work internally.

Works for me.

etoledom:
> I was thinking we would make separate PRs for those but I am fine if you wanna do it in this PR too.

Ok makes sense. I'll remove de deprecated warning to pass CI, then we can focus on removing completion blocks from the services 👍

etoledom:
- https://github.com/Automattic/Gravatar-SDK-iOS/issues/120
- https://github.com/Automattic/Gravatar-SDK-iOS/issues/121

pinarol:
> A bit more explanation of the reasoning: We agreed on doing this while working on iOS-Android parity, and on paper it seemed to be a good improvement over method signature. But now that I see it in practice, it does little to nothing good (or could even make it worse as you mention).

I see. Thanks for explaining. In this particular case I think it's not worth the readability cost that we are introducing. Just using plain String would be much better imo.

andrewdmontgomery:
> > I didn't understand the need behind this one. To me this makes the code more difficult to read. At the first glance it looks like different type like a struct but it is not, so now i need to make an effort to discover the real type. 🤔
> 
> A bit more explanation of the reasoning: We agreed on doing this while working on iOS-Android parity, and on paper it seemed to be a good improvement over method signature. But now that I see it in practice, it does little to nothing good (or could even make it worse as you mention).
> 
> I have added it to this PR since we agreed on it, but as I mentioned on the description of the PR, I don't see any gains on it, and now that you mention the extra steps on the user side to figure out what's the type itself, it seems like a drawback more than anything good.
> 
> After seeing it, I'd pretty much agree to better keep `Strings` instead of using the aliases.
> 
> cc @andrewdmontgomery - since we talked about this.

Yeah, I fully agree.  It would be one thing if these were actual types.  At least then we could be validating the values, for example.  But this is really just another label for something that we are already labeling very clearly:

```
fetch(withEmail email: Email)
fetch(withHash hash: Hash)
fetch(withUserName userName: UserName)
```

100% redundant.  And as you pointed out, @pinarol, it also obscures the type.

hamorillo:
> Yeah, I fully agree. It would be one thing if these were actual types. At least then we could be validating the values, for example. 

@maxme is working on this on the Android side. We've discussed the `typealias` too and we agreed that it is not a good approach in our case, basically because of the issues you've already mentioned. We agreed to give it a try using real classes that can contain other functions as validation stuff. @maxme has just opened a PR. It is still a WIP, though.

etoledom:
> We agreed to give it a try using real classes that can contain other functions as validation stuff. @maxme has just opened a PR. It is still a WIP, though.

@hamorillo 
I've tried a concrete type also. Though it works for method overload, I think it has little practicality on the user side. 
Having to instantiate a type to pass a String has no practical value, and is unnecessary extra work for the user.

I have updated the PR to use `Strings` with the added labels to the function name to differentiate between them.

pinarol (approved):
LGTM 👍

etoledom:
Let's go with this for now. We can revisit the `Email` `Hash` `UserName` types afterwards.

PR #122, opened 2024-03-19
Add AvatarService and refactor accordingly

### LINKED ISSUE #111: Create new `AvatarService`
`AvatarService` Should be specific to user avatar images.

```swift
fetch(with email: String, options: ImageDownloadOptions) async throws -> ImageDownloadResult {}
upload(_ image: UIImage, email: String, accessToken: String) async throws  -> URLResponse {}
```

The current `ImageService` will keep the methods related URL `fetch(_ url: URL...)`

pinarol:
Closes #111 #120 

WP PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/22878

### Description

- Adds "AvatarService" to download & upload avatar via email.
- Splits ImageService into ImageDownloadService, ImageUploadService.
   - ImageUploadService is an internal type. Currently exists only for convenience and single responsibility.
   - ImageDownloadService is a public type that lets you download an image via URL.
- Removes the completion handler supporting alternatives of the existing async methods. I just left this one for internal usage (it's not public anymore):
```
 func fetchImage(
        with url: URL,
        forceRefresh: Bool = false,
        processingMethod: ImageProcessingMethod = .common,
        completionHandler: ImageDownloadCompletion?
``` 

### Testing Steps

Test image download & upload via demo app


Sources/Gravatar/Network/Services/ImageDownloader.swift
```diff
@@ -5,51 +5,36 @@ public typealias ImageDownloadCompletion = (Result<ImageDownloadResult, ImageFet
...
+}
 
-    /// Fetches an image from the given `URL`, and delivers the image asynchronously.
-    /// - Parameters:
-    ///   - url: The URL from where to download the image.
-    ///   - forceRefresh: Force the image to be downloaded, ignoring the cache.
-    ///   - processingMethod: Method to use for processing the downloaded `Data`.
-    /// - Returns: An asynchronously-delivered Result type containing the image and its URL.
+extension ImageDownloader {
+    @discardableResult
```
pinarol, line 22: Note: This is an internal method now.

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -0,0 +1,49 @@
...
+    /// `ImageUploadError`.
+    /// - Parameters:
+    ///   - image: The image to be uploaded.
+    ///   - accountEmail: The user email account.
+    ///   - accountToken: The authentication token for the user. This is a WordPress.com OAuth2 access token.
+    /// - Returns: An asynchronously-delivered `URLResponse` instance, containing the response of the upload network task.
+    @discardableResult
+    public func upload(_ image: UIImage, accountEmail: String, accountToken: String) async throws -> URLResponse {
+        try await imageUploader.uploadImage(image, accountEmail: accountEmail, accountToken: accountToken)
+    }
```
etoledom: Let's have the label names as mentioned in the de docs:

```swift
upload(_ image:email:accessToken:)
```

Sources/Gravatar/Network/Services/ImageDownloader.swift
etoledom: Will we refactor the ImageView extension later on to remove the internal use of the completion closure?

We can use concurrency internally in the extension while the public method `UIImageView.gravatar.setImage()` keeps the completion closure.

pinarol: My thinking was if we add support for more UI components (SwiftUI or UIKit) this can be a handy utility function. Because without this, we'll need to do this conversion in every UI component separately. WDYT?

Sources/Gravatar/Network/Services/AvatarService.swift
pinarol: Oops I apparently missed it. Updated now.

pinarol:
Docs are also updated.

pinarol:
Okay I converted the draft into a real PR. Ready for another check.

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+import UIKit
+
+/// AvatarService helps with uploading and downloading avatars.
```
etoledom:
```suggestion
/// A service to perform uploading and downloading of avatars.
///
/// An avatar is a profile image of a Gravatar user. See [the avatar docs](https://support.gravatar.com/profiles/avatars/) for more info.
```
To be more similar to the descriptions of `ProfileService` and `ImageDownloadService`.

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -0,0 +1,49 @@
...
+    /// Similarly, you can pass a custom type conforming to ``ImageCaching`` to use your custom caching system.
+    /// - Parameters:
+    ///   - client: A type which will perform basic networking operations.
+    ///   - cache: A type which will perform image caching operations.
+    public init(client: HTTPClient? = nil, cache: ImageCaching? = nil) {
+        self.imageDownloader = ImageDownloadService(client: client, cache: cache)
+        self.imageUploader = ImageUploadService(client: client)
+    }
+
+    /// Fetches a Gravatar user profile image using the user account's email, and delivers the image asynchronously. See also: `ImageDownloadService` to
```
etoledom:
```suggestion
    /// Fetches a Gravatar user profile image using the user account's email, and delivers the image asynchronously. See also: ``ImageDownloadService`` to
```

With this double quotation marks, the swift-docc will automatically generate a link to the docs page of this symbol :)

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -0,0 +1,49 @@
...
+    ) async throws -> ImageDownloadResult {
+        guard let gravatarURL = GravatarURL.gravatarUrl(with: email, options: options.imageQueryOptions) else {
+            throw ImageFetchingError.requestError(reason: .urlInitializationFailed)
+        }
+
+        return try await imageDownloader.fetchImage(with: gravatarURL, forceRefresh: options.forceRefresh, processingMethod: options.processingMethod)
+    }
+
+    /// Uploads an image to be used as the user's Gravatar profile image, and returns the `URLResponse` of the network tasks asynchronously. Throws
+    /// `ImageUploadError`.
```
etoledom:
```suggestion
    /// ``ImageUploadError``.
```

Sources/Gravatar/Network/Services/ImageDownloader.swift
etoledom: Makes sense, yes.

One thing though, by not using completion blocks internally, we can take full advantage of the concurrency model and offer thread-safety proven by the compiler. Which is not the case if we use completion blocks internally.

We can keep this for now and re-think this when we work on improved adoption of the concurrency model throughout the sdk 👍

etoledom (approved):
Changes look great 🎉 
Smoke-tested the demo app and all seems good.

I leaved a few comments, mostly regarding documentation, which are not blockers.

There's one more warning when building the docs here:

https://github.com/Automattic/Gravatar-SDK-iOS/blob/1e04be9ea692348b17cb61e14c18d3a9dcf1929f/Sources/Gravatar/Gravatar.docc/Gravatar.md?plain=1#L30

To see these warnings, is good to build the docs via xcode. You can do it by:
Product -> Build Documentation.

Feel free to merge when ready 👍 🚢

Sources/Gravatar/Network/Services/AvatarService.swift
pinarol: Aha :TIL:

pinarol:
I am about to open a WP side PR too, I thought we wouldn't need it for this change but we do. My bad I didn't notice earlier. So I am going to request an additional review for the WP side and hold this one until that's also approved.

pinarol:
> One thing though, by not using completion blocks internally, we can take full advantage of the concurrency model and offer thread-safety proven by the compiler. Which is not the case if we use completion blocks internally.

I am all up for doing those improvements and I would love to learn more on the subject. But here's what I didn't quite understand: If we are going to convert from async-awat to the completion block at some point, does it make a difference if we do this conversion here or in the UIImageView extension?

pinarol:
Okay I opened it: https://github.com/wordpress-mobile/WordPress-iOS/pull/22878 Could you check that too? @etoledom

pinarol:
> Product -> Build Documentation.

Great. I did this and then fixed the documentation warning.

etoledom:
> I am about to open a WP side PR too, I thought we wouldn't need it for this change but we do. My bad I didn't notice earlier. So I am going to request an additional review for the WP side and hold this one until that's also approved.

The big difference is that we will use a Task on the body of the method that is being called with a completion block, everything that happens inside that method is being checked by the concurrency model, and at some point we call the completion block on main thread when the await ends.

In contrast, if we create a completion block ourselves from within the body of the method, and then we call the user's completion block from within our internal completion block, we would lost the compiler-time guarantee that this method is thread safe.

In both cases though, the user will lose these compiler-time checks on their call to our method.

---

I was doing some tests about this yesterday and we could declare this method as `nonisolated` to make it run on background thread, even if it's a method declared in a MainActor class, In this case the user would need to create a Task for it, but with the benefit of removing the completion block, and we move thread safety one layer up to the caller, which I think is better.

I'd like to see how kingfisher goes around this, if they do it at all.

pinarol:
Okay I think best way to discuss this concurrency stuff would be on a separate PR with code examples. It's difficult for me to follow this way.

pinarol:
Thanks a lot for your help! Merging now. 🎉

PR #123, opened 2024-03-19
Migrate to separate modules for `Gravatar` and `GravatarUI`

### LINKED ISSUE #113: Create new package module to contain UI Components
Create `GravatarUIComponents` module.

andrewdmontgomery:
Closes #113 

### Description

This refactors the Swift Package and the CocoaPods podpsecs to support two separate libraries/modules:
- `Gravatar` ~GravatarCore~ (which was renamed from `Gravatar`)
- `GravatarUI` ~GravatarComponents~

## Questions?
- What do we think of the structure/architecture?
- What do we think of the naming?

## Architecture
This approach implies that `GravatarUI ` uses UIKit.  That's a safe implication for now.  I had originally structured this with the ability to dynamically load `GravatarUI ` based on the required platform.  The idea was that `GravatarUI` would have dependencies that were platform-specific:

```swift
        .target(
            name: "GravatarUI",
            dependencies: [
                .byName(name: "GravatarUIKit", condition: .when(platforms: [.iOS]))
            ]
        ),
        .target(
            name: "GravatarUIKit",
            dependencies: ["Gravatar"]
        ),
```

But even if we were ready to support more than one platform, it's not clear to me that this would be the best approach anyway.

So for multiple reasons, I've gone with the simpler approach.

## Why separate podspecs?
CocoaPods creates a module from each podspec it installs.  We need there to be a `Gravatar` module and a `GravatarUI` module, so that the import statements in the source files work.

--------
## SPM
The changes in the Package.swift are pretty simple:
- Added a library: `GravatarUI `
- Added two targets:
  - target: `GravatarUI ` (dependency: `Gravatar`)
  - testTarget: `GravatarUI Tests`

When a project wants to import Gravatar as a Swift Package, they will be presented with the following choices:
image

### SPM: Finding the Resource Bundle for tests
If we have defined a target with resources in the `Package.swift`, and if there are resources in that path, Swift Package Manager will autogenerate a static property on `Foundation.Bundle` named `module`.

We use this property via a helper in our test suites:
`Tests/GravatarUITests/Bundle+ResourceBundle.swift`

```swift
    public class var testsBundle: Bundle {
        #if SWIFT_PACKAGE
        return Bundle.module
        #else
        // [snip]
       #endif
    }
```

Note that this property is only generated if BOTH of the following are true:
- We have defined a target with resources in `Package.swift`
- Resources actually exist

If either of these are not true, the code in the block above will fail to compile.  That's why there is a `dummy.png` in the Resources folder of the new `GravatarUITests` folder.

------
## CocoaPods
The changes in CocoaPods require more of an explanation, because there's more going on:
- Removed all configurations related to source files, tests, etc. (because they will be handled by the two new dependencies instead)
- Added a podspecs for `GravatarUI`:
  - `GravatarUI.podspec`

### Local Pods
When importing locally, explicitly import each pod (optionally with tests):
```ruby
def gravatar
  pod 'Gravatar', path: '../../Gravatar/Gravatar-SDK-iOS', :testspecs => ['GravatarTests']
  pod 'GravatarUI', path: '../../Gravatar/Gravatar-SDK-iOS', :testspecs => ['GravatarUITests']
end
```

### Published Pods
We will publish all of our podspecs when the time comes.  Once they are published, then we will be able to install the `Gravatar` pod and have both dependencies installed as well:

```ruby
pod 'Gravatar', '~> 0.1.1'
pod 'GravatarUI', '~> 0.1.1'
```

Or we can always install them separately (specifying tests is optional):
```ruby
pod 'Gravatar', '~> 0.1.1`, :testspecs => ['GravatarTests']
pod 'GravatarUI, '~> 0.1.1`, :testspecs => ['GravatarUITests']
```

### CocoaPods: Naming and referencing the Resource Bundle for tests
The new podspec configures a "resource_bundle", like this:
```ruby
      swift_unit_tests.resource_bundles = {
          GravatarUITestsResources: [
              'Tests/GravatarUITests/Resources/**/*'
          ]
      }
```

This string `GravatarUITestsResources` is arbitrary, but it is referenced somewhere else, so it needs to match in both places.  The other place that uses it is in `Tests/GravatarUITests/Bundle+ResourceBundle.swift` (and the equivalent path for `Gravatar`:
```swift
    public class var testsBundle: Bundle {
        #if SWIFT_PACKAGE
        return Bundle.module
        #else
        let defaultBundle = Bundle(for: BundleFinder.self)
        // If installed with CocoaPods, resources will be in GravatarUITestsResources.bundle
        if let bundleURL = defaultBundle.resourceURL,
           let resourceBundle = Bundle(url: bundleURL.appendingPathComponent("GravatarUITestsResources.bundle"))
        {
            return resourceBundle
        }
        // Otherwise, the default bundle is used for resources
        return defaultBundle
        #endif
    }
```

The `#else... #endif` block is executed when this code is called via CocoaPods.  In that situation, CocoaPods will have created a resource bundle with the name specified in the podspec.  So this line needs to match:
```swift
bundleURL.appendingPathComponent("GravatarUITestsResources.bundle"))
```

### CocoaPods: Ensuring that the resource bundle is created
Note that configuring the resource bundle in the podspec is necessary, but it is not sufficient.  CocoaPods will only create the resource bundle if there are resources to bundle.  That's why the `Tests/GravatarUITestsTests/Resources` folder contains a `dummy.png`.  Without any resources, CocoaPods won't create the bundle, and the block of code above will fail.

------
## Testing Steps

- [ ] Ensure that `Gravatar` and `GravatarUI` can be added to WordPress-iOS as local Pods
  - [ ] Public types are accessible
  - [ ] Tests should run and pass

- [ ] Ensure that `GravatarCore` and `GravatarComponents` can be added as local packages
  - [ ] Public types are accessible


andrewdmontgomery:
NOTE: I bumped the version numbers in the podspecs from `0.1.0` to `0.1.1`.  But we will want to coordinate with Android when choosing the actual version number.

andrewdmontgomery:
Ah, interesting:
https://github.com/Automattic/Gravatar-SDK-Android/pull/76

I like their module names:
- `gravatar`
- `gravatar-ui`

I thought we had agreed to use the word `Core`, but I may have misunderstood.  I had originally changed the names to `GravatarCore` and `GravatarComponents`.  But if we don't change the name of the "Core" module, this becomes a much smaller change.  I am going to update this PR to rename `GravatarCore` back to `Gravatar`, and `GravatarComponents` to `GravatarUI`.

pinarol (approved):
LGTM 🎉

etoledom (approved):
Looking great! Thanks for doing this 🙏 

One small detail, I think to finish the setup on the Demo app, we still need to add GravatarUI to the linked Libraries.

CleanShot 2024-03-21 at 11 18 35@2x

andrewdmontgomery:
Great catch, thanks.

PR #125, opened 2024-03-21
Automate SDK Releases

### LINKED ISSUE #76: Setup Buildkite to handle releases
Update Buildkite pipeline to automate releases.  The outcome should be:

1. Developer makes a PR bumping the version number
2. Once merged, developer adds a `tag` for that commit
3. CI handles the rest
  - Create "Release" bundle in repo
  - Push PodSpec to CocoaPods trunk

```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/Gravatar-SDK-iOS/issues/77
- [ ] https://github.com/Automattic/Gravatar-SDK-iOS/issues/78
```

andrewdmontgomery:
Closes #76 #126 #127

### Description

This PR makes changes so that we can have CI reliably handle version releases:
- Sets up Buildkite to handle publishing of our podspecs
- Sets up Buildkite to build both versions of the Demo project (Swift and SwiftUI)
- Implements Fastlane so that we have easier and more consistent control over how the Package and the Demo project are built and tested

### Automated Releases

With this change in place, releases will work as follows:

1. A developer opens a PR to bump the versions in all the Podspecs to the appropriate version string
2. Once that PR merges, the developer will add a new release in the Github Repository.  The release will include a tag, and that tag will be the same version string we set in the Podspecs.

When that tag is created, Buildkite will be notified, and will run the pipeline for this repository.  All of the same steps will be run.  But this time, an extra step will also run:

```yaml
  #################
  # Publish the Podspec (if we're building a tag)
  #################
  - label: "⬆️ Publish Podspec"
    key: "publish"
    command: |
      .buildkite/commands/publish-pod.sh "Gravatar.podspec"
      .buildkite/commands/publish-pod.sh "GravatarUI.podspec"
    plugins: [$CI_TOOLKIT]
    depends_on:
      - "test"
      - "test_demo"
      - "validate"
      - "lint"
    if: build.tag != null
    agents:
      queue: "mac"
```

The key is the line `if: build.tag != null`.  This line causes this step to only run if the build was triggered by adding a new tag to the repository.

### Publishing Podspecs

This PR introduces a new script, which Buildkite uses when running the `Publish Podspec` job.  The script is simple, and uses three commands from the `CI_TOOLKIT` [repository](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/tree/trunk/bin): 

- [install_gems](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/install_gems)
- [publish_pod](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/publish_pod)
- [slack_notification_pod_published](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/slack_notify_pod_published)

Each of these can be seen in detail at the links above.  The most important part is the `publish_pod` helper.  This helper does a few things, two of which are important to us:

- It verifies that the `version` specified in each podspec matches the `tag` associated with the build.  If the versions don't match, it won't publish.
- It uses cocoapods to publish

Since this requires us to publish one at a time, and since the version verification only happens before each podspec is published, we run the risk of having some of our podspecs fail to publish.

So we've added some logic to verify the versions of all podspecs before publishing.  Since the `publish_pod` helper includes its own verification, this does mean that we are verifying the version of each podspec twice.  That second check is redundant, of course.  But its fine.

### Introducing Fastlane

This PR introduces Fastlane into our repository.  We are now building three different products in CI:
- The SDK Package and tests
- The Gravatar-Demo (Swift) app
- The Gravatar-SwiftUI-Demo app

Fastlane makes it much easier to be consistent about how we configure these builds.

### Testing Steps

- [ ] CI should be 🟢 
- [ ] Test the Makefile:
  - [ ] `make test` should build and test the SDK
  - [ ] `make build-demo-swift` should build the `Gravatar-Demo` target in the Demo project
  - [ ] `make build-demo-swiftui` should build the `Gravatar-SwiftUI-Demo` target in the Demo project
  - [ ] `make build-demo` should build both
  - [ ] `make validate-pod` should build the SDK for CocoaPods and run all testes


andrewdmontgomery:
Hey @jkmassel , would you (or someone from your team knowledgable in publishing podspecs) mind taking a look at this?

pinarol:
> A developer opens a PR to bump the versions in all the Podspecs to the appropriate version string

Should we always bump the versions in all the podspecs? What if we only change GravatarUI and don't change anything in Gravatar? I don't know the best practice here, just asking.

etoledom:
> Should we always bump the versions in all the podspecs? What if we only change GravatarUI and don't change anything in Gravatar? I don't know the best practice here, just asking.

I personally like when a repo with different modules have all their module version synced. It's easier to reason on which are the versions I should use for all the modules.

andrewdmontgomery:
> > A developer opens a PR to bump the versions in all the Podspecs to the appropriate version string
> 
> Should we always bump the versions in all the podspecs? What if we only change GravatarUI and don't change anything in Gravatar? I don't know the best practice here, just asking.

Thanks for pointing this out.  I had meant to call this out for discussion.

This is getting at a subtle difference between the way that Swift Package Manager and CocoaPods handle modules.  With SPM, all modules are defined in the same `Package.swift`.  But in CocoaPods, modules are defined in separate `podspec` files.  This is a limitation (or feature?) of CocoaPods.

CocoaPods supports `subspecs`.  But subspecs are just logical groupings.  When `pod install` creates the `Pod.xcodeproj` (where CocoaPod modules would be defined), `podspecs` are used to create modules.  But `subspecs`  become part of the podspec they are defined in, and therefore part of the _module_ that `pod install` will create from that podspec.

The problem is with the `import` statements in our code.  We need SPM and CocoaPods to agree on which modules are linked, and which symbols are in each of those modules.  If we use `subspecs` to define `GravatarUI` inside `Gravatar`, then with CocoaPods, there will only be one module, `Gravatar`, and all of the symbols will only be accessible via `import Gravatar`.  But SPM will see both modules, and symbols will only be available from the one they are compiled into.

So we can't use subspecs in CocoaPods.  We need to define a full podspec for each module.

This has a subtle implication for versioning:
- SPM: version applies to the whole `Package.swift` (and all modules)
- CocoaPods: version applies to each module individually

Since we support both types of dependency management, the easiest (only?) approach is to apply version numbers to the whole SDK.

For CocoaPods, that means we'll need to bump the version on each podspec.

The good news is that this will make it harder for us to miss pushing a Podspec during a release.  The release automation verifies that the `version` in each Podspec matches the new git tag.  If we forget to bump the version in a podspec, the entire release will fail.

If we were versioning individual podspecs, we wouldn't be able to use that failure mode.  If a podspec version didn't match the tag version, we would have assume that it didn't need to be published, and we would have to skip pushing that Podspec rather than failing the release.  If we made changes to the actual module, though, and we didn't push a new podspec, third parties wouldn't have access to that module's new version.  And that failure would be relatively silent.

> I personally like when a repo with different modules have all their module version synced. It's easier to reason on which are the versions I should use for all the modules.

It's also very common, probably for the reasons I mentioned.  Take Firebase, for example:
https://github.com/firebase/firebase-ios-sdk

andrewdmontgomery:
Hey @Automattic/apps-infrastructure : 

I tagged @jkmassel , and then realized he's AFK.  Would one of you be able to take a look at this?  I'd love any feedback you have.  But specifically, could you verify the Buildkite changes?  The goal of this PR is to automatically publish Podspecs when a new github tag/release is created.

I borrowed heavily from the setup in [WordPressUI-iOS](https://github.com/wordpress-mobile/WordPressUI-iOS)

- Have I set this up correctly?
  - Will it only run when a new tag/release is added to GitHub?
  - Is our Buildkite [pipeline](https://github.com/Automattic/buildkite-ci/blob/trunk/src/buildkite/pipelines/gravatar-sdk-ios.tf) configured for this to work?
- Since I'm using the `publish-pod` helper that you provide, I believe this will post the Slack:
  - Is that appropriate?
  - Is that the [#build-and-ship](https://a8c.slack.com/archives/C5ALGJYEP) channel?

We're hoping to merge this early next week, if possible.

.buildkite/commands/publish-pods.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash -eu
+
+if [ $# -eq 0 ]; then
+    echo "Error: CocoaPods publishing failed. All podspecs must be specified, separated by spaces."
+    exit 1
+fi
+
+SLACK_WEBHOOK=$PODS_SLACK_WEBHOOK
```
iangmaia: Reminder that you need to export a `PODS_SLACK_WEBHOOK` environment variable for this value to be available in Buildkite.

You need to add something like this to your `~/.mobile-secrets/CI/secrets/gravatar-sdk-ios/env` file:

```bash
export PODS_SLACK_WEBHOOK=${PODS_SLACK_WEBHOOK}
```

.buildkite/commands/publish-pods.sh
```diff
@@ -0,0 +1,35 @@
...
+PODSPEC_PATHS=("$@")
+
+# Verify that all versions are current before publishing any podspecs
+echo "--- :cocoapods: Verify version of all podspecs"
+invalid_pod=false
+for podspec_path in "${PODSPEC_PATHS[@]}"; do
```
iangmaia: Suggestion: I believe the script could be much cleaner if you dealt with a single parameter (the spec) instead of an array, and then you'd just call the script twice.

Correct me if I'm wrong, but in the near future, I don't think we need the flexibility of handling more than two parameters (when handling arrays would make more sense)?

.buildkite/commands/publish-pods.sh
```diff
@@ -0,0 +1,35 @@
...
+    pod_version=$(bundle exec pod ipc spec "$podspec_path" | jq -r '.version')
+    if [ -n "$BUILDKITE_TAG" ] && [ "$BUILDKITE_TAG" != "$pod_version" ]; then
```
iangmaia: I believe this check is already done by [publish_pod](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/f077728b25454cd7e727481096306603ae33d782/bin/publish_pod#L18)?

iangmaia:
Hey @andrewdmontgomery! I wrote a couple of comments in the PR.

---

To your questions:

> The goal of this PR is to automatically publish Podspecs when a new github tag/release is created.

> Will it only run when a new tag/release is added to GitHub?

> Is our Buildkite [pipeline](https://github.com/Automattic/buildkite-ci/blob/trunk/src/buildkite/pipelines/gravatar-sdk-ios.tf) configured for this to work?

This is looking good to me; the pipeline will [build tags](https://github.com/Automattic/buildkite-ci/blob/961c606b538b41983532ecf83d2eada505a77be0/src/buildkite/pipelines/gravatar-sdk-ios.tf#L37) and the job configuration [looks good](https://github.com/Automattic/Gravatar-SDK-iOS/pull/125/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53R61).

> Since I'm using the `publish-pod` helper that you provide, I believe this will post the Slack:

> Is that appropriate?

I think so, if we're following the pattern we have for other Pods.

> Is that the [#build-and-ship](https://a8c.slack.com/archives/C5ALGJYEP) channel?

Exactly 🎯 [this](https://a8c.slack.com/archives/C5ALGJYEP/p1710716189166459) is an example message.

pinarol:
Thanks for the detailed explanation! Sounds good to me. 👍

.buildkite/commands/publish-pods.sh
andrewdmontgomery: Good question.  Here's the problem I'm solving for:

1. We forgot to bump the version in one podspec
2. We add a tag to the repo
3. CI runs this automation
4. One podspec successfully publishes, the other fails.

Then our published podspecs would be out of sync.

With this approach, the script can verify that both podspecs have the expected version, and fail before any are pushed.

andrewdmontgomery: That said, I suppose in the scenario above, we could:
1. See that CI failed to push one pod
2. Bump the version
3. Remove the tag/release from the repo
4. Re-add the tag/release with the new commit

In fact, I think the steps will be the same regardless.  It's just that in the current setup, our published podspecs won't be out of sync while we sort this out.

Is that worth more complexity in the scripts?  Dunno.

.buildkite/commands/publish-pods.sh
andrewdmontgomery: Yep.  See above.

.buildkite/commands/publish-pods.sh
andrewdmontgomery: Thanks!  I've pushed that (and the CocoaPods trunk token).

.buildkite/commands/publish-pods.sh
iangmaia: > Then our published podspecs would be out of sync.

That's what I was guessing but in principle it didn't seem worth it for me. But on a second thought, the issue the code is solving is clearer for me now, thanks for the explanation.

> It's just that in the current setup, our published podspecs won't be out of sync while we sort this out.
> 
> Is that worth more complexity in the scripts? Dunno.

I think that's the question here.

My first instinct is that this might raise questions in the future and it's perhaps not clear enough why it is implemented this way. 🤔  Perhaps just rephrasing or adding more details to the existing comment is enough, if you rather keep the code with the array (which is fine with me as well).

andrewdmontgomery: Thank you for this!  I'm tempted to take your suggestion and simplify this.  If we were doing this in Ruby or Python, I'd have more confidence.  But `Bash + arrays` can really turn into a swamp of pain.

andrewdmontgomery: Even better: we can have CI verify that the versions in both podspecs match.  That way, we catch a version mismatch on every PR (including the one where we bump the versions).

Then this bash script can be trivial.

This is much better.  Thank you for taking your time with this!

AliSoftware: That's a good point to raise. But at this point, if the `podspec` versions are out of sync with the tag, it would be even better to avoid the tag to happen beforehand, or even better, avoid this to be possible in the first place 😉 

One option for checking before the tag would be to check that all `podspec` files use the same version during `validate-pods.sh`—which is run on any commit (even non-tag ones) and is a pre-requisite for `publish-pods.sh`—without even considering if there's a tag or not or comparing to it

But imho a better option to prevent this from even happening would be to have each `.podspec` require a `version.rb` file that contains a `VERSION = …` as a constant, and then have each `.podspec` have `s.version = VERSION` to reference it. That way not only there's a single source of truth and a single place to change when bumping the version, but it also prevents the problem from even existing.

andrewdmontgomery: Yes!  A `version.rb` is perfect.  That's what we'll do.  Thank you both.  Give me a couple mins.

andrewdmontgomery: @AliSoftware , @iangmaia 

Much cleaner, much less 🤞 with Bash arrays

.buildkite/commands/publish-pods.sh
andrewdmontgomery: Simplified substantially, see thread above.

version.rb
```diff
@@ -0,0 +1,3 @@
+module Gravatar
+    VERSION = '0.1.1'.freeze
```
iangmaia, line 2: 👍

iangmaia (approved):
Thank you for considering our comments.
The PR is looking good to me 👍 :shipit:

andrewdmontgomery:
Thanks all! I'm merging this for now. But if you think of anything else, don't hesitate to mention it.

PR #128, opened 2024-03-22
Split GravatarURL into AvatarURL and ProfileURL

### LINKED ISSUE #110: Split `GravatarURL` into `AvatarURL` and `ProfileURL`
## AvatarURL

```swift
// Fields:
let canonicalUrl: URL 
var hash: Hash 
// Methods:
init?(_ url: URL)
init?(_ email: Email)
init?(_ hash: Hash)
url(with options: AvatarQueryOptions) -> URL
static func isAvatarUrl(_ url: URL) -> Bool
```

## ProfileURL
```swift
// Fields:
let url: URL
let hash: Hash
var avatarURL: AvatarUrl { get }
// Methods:
init?(with email: Email)
init?(with hash: Hash)
```
`Hash`, `Email` are `String` `typealias`.

etoledom:
Closes #110 

### Description

This PR removes `GravatarURL` to be replaced by `AvatarURL` and `ProfileURL`.

WPUI PR: https://github.com/wordpress-mobile/WordPressUI-iOS/pull/153

### Testing Steps

- CI should be 🍏 
- Smoke-test the demo app
- Test the WPUI PR: https://github.com/wordpress-mobile/WordPressUI-iOS/pull/153


Sources/Gravatar/AvatarURL.swift
```diff
@@ -0,0 +1,85 @@
...
+
+    let options: ImageQueryOptions
+    var components: URLComponents
+
+    public var url: URL {
+        // When `AavatarURL` is initialized successfully, the `canonicalUrl` field is a valid URL.
+        // Adding query items from the options, which is controlled by the SDK, should never
+        // result in an invalid URL. If it does, something terrible has happened.
+        guard let url = canonicalUrl.addQueryItems(from: options) else {
+            fatalError("Internal error: invalid url with query items")
```
pinarol: We should never use fatalError() in PROD, unless it is something that is completely unrecoverable for an app, which is not the case here. We can maybe use assertionFailure() but in this case I think we don't need that either. 

Because we already have a failable initializer. Why not just move this code there?

```
    public init?(url: URL, options: ImageQueryOptions = ImageQueryOptions()) {
        guard
            Self.isAvatarUrl(url),
            let components = URLComponents(url: url, resolvingAgainstBaseURL: false)?.sanitizingComponents(),
            let sanitizedURL = components.url,
            let url = sanitizedURL.addQueryItems(from: options)
        else {
            return nil
        }

        self.canonicalUrl = sanitizedURL
        self.components = components
        self.hash = sanitizedURL.lastPathComponent
        self.options = options
        self.url = url
    }
```

Sources/Gravatar/AvatarURL.swift
```diff
@@ -0,0 +1,85 @@
+import Foundation
+
+public struct AvatarURL {
+    public let canonicalUrl: URL
+    public let hash: String
+
+    let options: ImageQueryOptions
+    var components: URLComponents
```
pinarol: should this be "let" instead of "var"?

Sources/Gravatar/ProfileURL.swift
```diff
@@ -0,0 +1,44 @@
...
+        AvatarURL(hash: hash)
+    }
+
+    static let baseUrl: URL = {
+        guard
+            let baseUrl = URL(string: .baseURL),
+            let components = URLComponents(url: baseUrl, resolvingAgainstBaseURL: false)?.sanitizingComponents(),
+            let url = components.url
+        else {
+            fatalError("A url created from a correct literal string should never fail")
```
pinarol: Same here. Let's not use fatalError(). This is not something that is totally unrecoverable for an app. So it's not justified to crash the whole app here. Even though we believe this would never happen in real life, the codebase is a living thing that can change so we'll never know. Let's stick with more safer principles.

A constructed URL is always an optional thing, it's totally OK to return an optional URL here.

Sources/Gravatar/ProfileURL.swift
```diff
@@ -0,0 +1,44 @@
+import Foundation
+
+public struct ProfileURL {
+    public let url: URL
+    public let hash: String
+    public var avatarURL: AvatarURL? {
+        AvatarURL(hash: hash)
+    }
+
+    static let baseUrl: URL = {
```
pinarol: nit:
instead of `baseUrl` we can maybe prefer `baseURL`, since URL is the official abbrevation.

Sources/Gravatar/AvatarURL.swift
etoledom: Right.

At first `public func replace(with options: ImageQueryOptions)` was supposed to be a mutating function, but then I decided to make a copy and I forgot to change `components`  back to `let` 👍

Sources/Gravatar/AvatarURL.swift
etoledom: Since now AvatarURL in immutable, we can do this also 👍 
This was mostly copy paste from `GravatarURL`. 

Good call.

Sources/Gravatar/ProfileURL.swift
etoledom: We can do this, but what's the reason for returning `nill` on the initializer? 

If this init fails, it will fail with any input from the user, and all their URL() will also fails and most apps will stop working, because it's a major bug on Fundation 🤔 

Again, we can make these inits failable, but we need to write down a reason for them to fail.

pinarol: I was actually suggesting to make only this "baseURL" static var optional. But I guess it has a domino effect of making the initializer also failable? I see.

Still, it is a known thing that URL construction from a String is a failable thing. And our initializer constructs a URL from String. So this makes our init also failable. Developers would understand this. They pass a string and we turn that into a URL(maybe that string contains an unacceptable char for a URL because maybe they passed the wrong string...). We can just write down something like "fails if 'self.url' initialization fails".

etoledom: >  But I guess it has a domino effect of making the initializer also failable?

Yes that's the problem.

`URL(string:)` says :
> This initializer returns nil if the string doesn’t represent a valid URL even after encoding invalid characters

Our url string is a literal `https://gravatar.com/` which is guaranteed to be a valid URL with valid characters, independent of the input of the user.

The email/hash the user might pass to the init is added to the url's path, which is non-failable, in which case the init should not be failable either.

But now giving it another thought, maybe by just having a `fatalError` in the code of the SDK, it could bring some warnings to developers and some might choose to not use it, so let's make inits failable anyway 👍 

I guess we can blame the inner works of URL for the failable init ¯\\\_(ツ)_/¯

andrewdmontgomery: I believe this fatalError came from me.  It was originally a force-unwrap.  The idea was just to take a forced unwrapped optional and give it an explicit error description.

My understanding, from many sources that I have come to trust (like [Hacking with Swift: When should you force unwrap optionals in Swift?](https://www.hackingwithswift.com/quick-start/understanding-swift/when-should-you-force-unwrap-optionals-in-swift)), is that force-unwrapping `URL(string: "https://www.gravatar.com")` is safe (and useful) to do in production code – as long as it is a string literal (or relies on a constant).

I think this is a perfectly valid place to force-unwrap.  If we don't like the `fatalError()` being public, we could revert.

Sources/Gravatar/ProfileURL.swift
```diff
@@ -0,0 +1,44 @@
...
+}
+
+extension URLComponents {
+    func sanitizingComponents() -> URLComponents {
+        var copy = self
+        copy.scheme = .scheme
+        copy.query = nil
+        return copy
+    }
+}
```
andrewdmontgomery: This is used by `AvatarURL`.  Can we either move it there (and make it private) or move it into its own file?

Tests/GravatarTests/AvatarURLTests.swift
```diff
@@ -0,0 +1,137 @@
...
+import XCTest
+
+final class AvatarURLTests: XCTestCase {
+    let verifiedAvatarURL = URL(string: "https://0.gravatar.com/avatar/205e460b479e2e5b48aec07710c08d50")!
+    let verifiedAvatarURL2 = URL(string: "https://gravatar.com/avatar/205e460b479e2e5b48aec07710c08d50")!
+
+    let exampleEmail = "some@email.com"
+    let exampleEmailSHA = "676212ff796c79a3c06261eb10e3f455aa93998ee6e45263da13679c74b1e674"
+
+    func testisAvatarUrl() throws {
```
andrewdmontgomery:
```suggestion
    func testIsAvatarUrl() throws {
```

Sources/Gravatar/ProfileURL.swift
andrewdmontgomery: That said, everyone expects a URL() to be optional.  I agree that it's reasonable to do here, too.

pinarol: I see 3 lines in that guard, only one of them is `URL(string: "https://www.gravatar.com")`. We still have 2 other lines that we force unwrap. And I can't be sure how that code could emerge in time. It also sets a certain example in the code base and it can spread.

If it was only this single line it could be ok, because we are sure there are no extra calculations here:

 `let baseURL = URL(string: "https://www.gravatar.com")!`

But unless we define the boundaries of force unwraps SUPER clearly, they can spread easily in the code base. 

Take this PR as an example. In the initial state of the PR we had the same fatalError in AvatarURL, but there, we were actually building the URL with some passed options. So in that case the URL was not constructed "only" from a static string. But the fatalError was there too. How do we draw the line? 

Maybe we can have limited number of URL definitions like this somewhere:

> static let baseGravatarURL = URL(string: "https://www.gravatar.com")!

But I would be careful to limit the force unwrap usage just there.

If we are 100% sure of the forced unwrap we can do it but if we are 99.999% sure we shouldn't do it. Both for safety reasons but also to prevent them from spreading in the code base.

etoledom: > Take this PR as an example. In the initial state of the PR we had the same fatalError in AvatarURL, but there, we were actually building the URL with some passed options. So in that case the URL was not constructed "only" from a static string

Is the same case than here. For the AvatarURL, the base URL was already a guaranteed valid URL since it passed the faliable init. And adding Query options is not a failable action. We are basically fighting the uncertainty of URL receiving a string that could not be a valid URL, with the certainty that we know we are working with a valid URL.

> I see 3 lines in that guard 

By apple docs, `URLComponents` will fail if `URL` is malformed (which is ensured to not be malformed since it's a literal string).

`URLComponents.url` is `nil` if:
> If the NSURLComponents has an authority component (user, password, host or port) and a path component, then the path must either begin with "/" or be an empty string. If the NSURLComponents does not have an authority component (user, password, host or port) and has a path component, the path component must not start with "//". If those requirements are not met, nil is returned.

And since this is a valid literal url string which won't change with any input of the user, we know and ensure that all the requirements are met.

If it success once, it will succeed always, independent of user input.
Now, I agree that the inner working of these structs might change in the future in ways we can't predict,  but what are the chances that a url like `https://gravatar.com/` will become a invalid url string?
And if it does become invalid, everything on the internet will need to be updated anyway.

etoledom: Updated to make inits failable 👍

etoledom:
Ready for another look 👀

Sources/Gravatar/ProfileURL.swift
pinarol:
```
    fileprivate func addQueryItems(from options: AvatarQueryOptions) -> URL? {
        guard var components = URLComponents(url: self, resolvingAgainstBaseURL: false) else {
            return nil
        }
        components.queryItems = options.queryItems

        if components.queryItems?.isEmpty == true {
            components.queryItems = nil
        }

        return components.url
    }
```

I don't think it's OK to force unwrap the result here. Not because it wouldn't work "today", but because it contains some logic in it and I can't guarantee that that logic won't change in ways that will make this risky.

What if we add a path there under some conditions? If we forget to add a / at the beginning of the path, [components.url will be nil](https://stackoverflow.com/questions/39852659/urlcomponents-url-is-nil). 

The more we add forced unwraps to the code base, the more difficult it gets to foresee the possibilities of how things can go wrong in the future.

And the forced unwraps will set an example in the code base. At some point the question of "how to draw the line?" will become too difficult to answer. "If we already have that one, why not add this one too? ".

If it's a SUPER well defined and limited usage like this, just creating a URL from a static string, no other logic tied to it, it could be fine. But once we start to add more logic around it, it becomes more difficult to manage. 

> "static let baseGravatarURL = URL(string: "https://www.gravatar.com/")!"

pinarol (approved):
👍

Sources/Gravatar/ProfileURL.swift
andrewdmontgomery: > But unless we define the boundaries of force unwraps SUPER clearly, they can spread easily in the code base.

This is one place where SwiftLint would have come in handy in a way that SwiftFormat doesn't.  SwiftLint can prevent force unwraps.  And then you can mark specific lines as exceptions.

That would pretty effectively address this:
- Make it clear via SwiftLint rules that force-unwrapping is strongly discouraged
- Forces developers to add a comment that disables the rule for a specific line of code

Then, as part of our code review, we would look for these disabling comments and discuss.  Eventually we could even have something like Danger check for new SwiftLint disabling comments and flag them in the PR.

The trick is in getting SwiftLint and SwiftFormat to play nice.  We'd want to:
- Allow SwiftFormat to handle anything it can handle
- Allow SwiftLint to handle only what cannot be handled effectively in SwiftFormat

That should be doable.

andrewdmontgomery: I created Draft PR #137 so we could investigate.

PR #132, opened 2024-03-25
Introduce ProfileIdentifier and AvatarIdentifier

### LINKED ISSUE #138: Implement ProfileIdentifier and AvatarIdentifier

andrewdmontgomery:
Closes #138 

### Description

This creates a `ProfileIdentifier` and an `AvatarIdentifer`, with supporting Types, in order to transform String values into concrete types that encapsulate and abstract away business logic.

I also added support in the Demo app for fetching a profile and an avatar with a hash

### Testing Steps

- [ ] CI should be 🟢 
- [ ] The Demo project should work for:
  - [ ] Fetching profiles by email and hash
  - [ ] Fetching avatars by email and hash
- [ ] Smoke tests in https://github.com/wordpress-mobile/WordPress-iOS/pull/22907


Sources/Gravatar/Identifiers/ProfileIdentifier.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+
+public enum ProfileIdentifier {
+    case username(Username)
```
pinarol: We don't know for sure if the new REST API will support querying with username. I mean they said "it could" but that's it. The docs currently only mention email & hash. So for now I suggest removing that case and adding it back in the future when this is certain.

pinarol:
This looks good overall thanks for putting effort into this. 

I like the naming "HashId" better than "Hash", it's more clear. The Id part could be capital maybe? Like HashID. I am not sure if we should respect to the camel case more, or, to the correct form of the abbreviation(ID) more. I tend to use camel case in a method name like "getByHashId" but when it comes to variable or type names I generally use HashID, baseURL. I guess I am not sure either :D 

I still have some doubts about whether HashId deserves to be a struct or not but I am up for trying. 

The Username struct on the other hand does not seem much necessary, it just wraps a string. But maybe we keep it just for  consistency in code? I guess it's ok... Also as I mentioned [here](https://github.com/Automattic/Gravatar-SDK-iOS/pull/132#discussion_r1538755465) we don't yet know if the new REST API will support querying with a username yet. So I would wait to be sure and then add it.

Sources/Gravatar/Identifiers/ProfileIdentifier.swift
andrewdmontgomery: Yeah, good call.  You had called this out in another thread.  I left it here just so we could see it in action, and to distinguish `ProfileIdentifier` from `AvatarIdentifier`.  With your note above, technically both of these would have the same cases.  But since their cases _could_ drift, I think it makes sense to make a distinction.

I'll pull it.

andrewdmontgomery:
> I like the naming "HashId" better than "Hash", it's more clear.

Thanks, I thought the same thing.

> The Id part could be capital maybe? Like HashID. I am not sure if we should respect to the camel case more, or, to the correct form of the abbreviation(ID) more. I tend to use camel case in a method name like "getByHashId" but when it comes to variable or type names I generally use HashID, baseURL. I guess I am not sure either :D

I feel exactly the same way.  I was on the fence about `baseURL` vs `baseUrl` (I was kinda leaning towards using `URL` when its type is `URL`, and `Url` when it's really a `String`, but that's another story, and I'm not even sure about that.)

I just tried making it `HashID`, and it's a lot more readable.  Good call, I think.  I made the change.

> I still have some doubts about whether HashId deserves to be a struct or not but I am up for trying.

The part that I like is that we can lean on the compiler.  We use that type all the way down the stack, until the last possible moment.  Then we pull out the String and use it.  This way, even the internal code is safe from mistakes.  We only have to worry about two phases, when we initialize it, and when we unwrap that String and use it.

pinarol (approved):
OK let's do this! 🎉 🍾 🥂

pinarol:
But we'll need to adapt WP side too. 

Edit: Now I realized we already have a PR for that: https://github.com/wordpress-mobile/WordPress-iOS/pull/22907

Sources/Gravatar/Identifiers/IdentifierProvider.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public protocol IdentifierProvider {
+    /// The string that the API expects when specifying a Gravatar type, such as an avatar or profile
+    var identifier: String { get }
+}
```
etoledom: I was thinking that `Identifiable` is a better name for this protocol, since the adopting type becomes "Identifiable".
But there's a native swift `Identifiable` protocol which does just that.

Maybe we can reuse `Identifiable` instead of declaring a new one?

etoledom: I made a fast test and it seems to work fine.

And since I had the changes already, I made a PR to show the proposal:
https://github.com/Automattic/Gravatar-SDK-iOS/pull/140

etoledom (approved):
This is super nice, thank you @andrewdmontgomery for this 🙏 

I just had one small comment [with a proposal](https://github.com/Automattic/Gravatar-SDK-iOS/pull/140), but either way, let's merge 👍

Sources/Gravatar/Identifiers/IdentifierProvider.swift
andrewdmontgomery: I'm looking now, and just from your comment above, this feels right.  It's funny, I almost called it `Identifiable`.  I'm surprised I didn't.  It would have been funny to have Xcode yell at me about re-declaring it. 😂

andrewdmontgomery:
There's one thing about this that stands out (a little).

These Identifiers work great for fetching.  Create an Identifier with whatever you have, and it will work.

But uploading seems to work only with an email address.  So those functions can't take one of these `Identifiers`.  So I've set them to explicitly require an `Email`.

If uploading and downloading could both use the same identifiers, that would be lovely.  But that's an API question, obviously.

pinarol:
> But uploading seems to work only with an email address. So those functions can't take one of these Identifiers. So I've set them to explicitly require an Email.
> If uploading and downloading could both use the same identifiers, that would be lovely. But that's an API question, obviously. 

That's fine. I think the ProfileIdentifier will be useful for the REST endpoints.

PR #133, opened 2024-03-25
Removing completion block from Profile Service

### LINKED ISSUE #121: Remove completion block methods from ProfileService
Also update WPiOS with the new async method

etoledom:
Closes #121 

### Description

Removing the completion-handler methods for fetching a profile.

Also updated WPiOS: 
https://github.com/wordpress-mobile/WordPress-iOS/pull/22894

### Testing Steps

- Smoke-test Demo app.

### DIFF
--- a/Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
+++ b/Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
@@ -66,29 +66,32 @@ class DemoFetchProfileViewController: UIViewController {
         }
         profileTextView.text = nil
         activityIndicator.startAnimating()
+        Task {
+            await fetchProfile(with: email)
+        }
+    }
+
+    nonisolated
+    func fetchProfile(with email: String) async {
         let service = Gravatar.ProfileService()
-        service.fetchProfile(with: email) { [weak self] result in
-            switch result {
-            case .success(let profile):
-                self?.setProfile(with: profile)
-            case .failure(let error):
-                self?.showError(error)
-            }
+        do {
+            let profile = try await service.fetch(withEmail: email)
+            await setProfile(with: profile)
+        } catch {
+            await showError(error)
         }
     }
 
     func setProfile(with profile: UserProfile) {
-        DispatchQueue.main.async { [weak self] in
-            self?.activityIndicator.stopAnimating()
-            self?.profileTextView.text = """
+        activityIndicator.stopAnimating()
+        profileTextView.text = """
 Profile URL: \(profile.profileUrl)
-Display name: \(profile.displayName)
-Name: \(profile.displayName)
+Display name: \(profile.displayName ?? "")
+Name: \(profile.displayName ?? "")
 Preferred User Name: \(profile.preferredUsername)
 Thumbnail URL: \(profile.thumbnailUrl)
 Last edit: \(String(describing: profile.lastProfileEditDate))
 """
-        }
     }
 
     func showError(_ error: Error) {
--- a/Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
+++ b/Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
@@ -71,14 +71,13 @@ class DemoFetchProfileViewController: UIViewController {
         }
     }
 
-    nonisolated
     func fetchProfile(with email: String) async {
-        let service = Gravatar.ProfileService()
+        let service = ProfileService()
         do {
             let profile = try await service.fetch(withEmail: email)
-            await setProfile(with: profile)
+            setProfile(with: profile)
         } catch {
-            await showError(error)
+            showError(error)
         }
     }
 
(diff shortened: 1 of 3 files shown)


Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
```diff
@@ -66,29 +66,32 @@ class DemoFetchProfileViewController: UIViewController {
...
-            case .success(let profile):
-                self?.setProfile(with: profile)
-            case .failure(let error):
-                self?.showError(error)
-            }
+        do {
+            let profile = try await service.fetch(withEmail: email)
+            await setProfile(with: profile)
+        } catch {
+            await showError(error)
```
pinarol: Q: Why "await" ? Those methods are just plain methods, no?

Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
```diff
@@ -66,29 +66,32 @@ class DemoFetchProfileViewController: UIViewController {
         }
         profileTextView.text = nil
         activityIndicator.startAnimating()
+        Task {
+            await fetchProfile(with: email)
+        }
+    }
+
+    nonisolated
```
pinarol: Q: Why is nonisolated necessary?

pinarol (commented):
Looks good. 🎉  I just have questions.

Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
etoledom: `UIViewController` is declared as `@MainActor`, so every method declared on it (or any subclass) is guaranteed to run on the main thread (when adopting and using the new concurrency model).

Whenever we want a method on a main actor insulated class to run its code on a different thread, we need to declare it as `nonisolated `, which is the case for `func fetchProfile(with:)`. In this case, the method will be executed outside of the main thread to execute and await for the network call.

`setProfile` and `showError` will run on the main thread (because they are methods on a UIViewController subclass) so, to call them from a method which is known at compile time that will run outside of the main thread, we need to await for them to complete their execution on main thread, and then we can continue execution on this method's thread if needed.

If we don't want to await for the execution of these methods in the main thread, we can also call them using Tasks, and you see that `await` is not needed any more:

```swift
nonisolated
func fetchProfile(with email: String) async {
    let service = Gravatar.ProfileService()
    do {
        let profile = try await service.fetch(withEmail: email)
        Task { @MainActor in
            setProfile(with: profile)
        }
        // Here any new code will run immediately, without waiting for  `setProfile` to finish its execution.
    } catch {
        Task { @MainActor in
            showError(error)
        }
    }
}
```

Previously, using completion block, the compiler wasn't able to understand completely which code would run in which thread, so `setProfile` could run in the background when called from a background queue, but when we adopt the concurrency model, we have the compiler enforcement  and guarantee of properly running the code in the threads they were meant to run, they are simply not able to be called and executed from the wrong thread 🌟

Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
etoledom: The full story is here: https://github.com/Automattic/Gravatar-SDK-iOS/pull/133#discussion_r1537619768 :)

etoledom: Though technically, `nonisolated `  is not mandatory here. We could spawn a Task from within `fetchProfile`:

```swift
func fetchProfile(with email: String) {
    let service = Gravatar.ProfileService()
    Task {
        do {
            let profile = try await service.fetch(withEmail: email)
            setProfile(with: profile)
        } catch {
            showError(error)
        }
    }
}
```

But a few different things happen here: 
- The code inside the Task will continue to run on the main thread (because now `fetchProfile` is also insulated to the main thread), and we will await on the main thread (but not blocking it).
- `let profile = try await service.fetch(withEmail: email)` will happen on a background thread, and when finished, we continue excecution on the main thread.
- See that now `setProfile(with: profile)` won't need to `await` even being inside a `Task`, because the task runs on the main thread.

I prefer `nonisolated ` because is easier to reason with, we can move more code outside of the main thread, and we have less `{ }` nesting.

pinarol: Thanks for the detailed explanation. It doesn't have to be on this PR but I think just making UserProfile Sendable results in a more simpler code. 

```
   func fetchProfile(with email: String) async {
        let service = Gravatar.ProfileService()
        do {
            let profile = try await service.fetch(withEmail: email)
            setProfile(with: profile)
        } catch {
            showError(error)
        }
    }
  ```

pinarol (approved):
👍

etoledom:
Updated according to https://github.com/Automattic/Gravatar-SDK-iOS/pull/133#discussion_r1537671243 👍 
Thanks for pointing it out 🙏

PR #137, opened 2024-03-26
[DO NOT MERGE] Add SwiftLint

andrewdmontgomery:
Closes #

### Description

This adds `SwiftLint` as a Build Tool plugin, and a basic configuration that enables only one rule: `force_unwrapping`.

### There's a problem
This doesn't currently play nicely with our implementation of SwiftFormat in our package.

SwiftFormat is giving some grief about SwiftLint.  SwiftLint technically supports macOS 12.  Our Package only specifies iOS .v15.  For some reason, SwiftFormat (which apparently looks at the Package.swift file) is throwing an error about this:
the library 'Gravatar' requires macos 10.13, but depends on the product 'SwiftLintPlugin' which requires macos 12.0; consider changing the library 'Gravatar' to require macos 12.0 or later, or the product 'SwiftLintPlugin' to require macos 10.13 or earlier.

I suspect this is because macOS 10.13 is the "default" when we don't specify in the Package.swift.
So when the plugin is added to our targets (so that it is run at build time for those targets), it's working ok because our Macs are running a newer OS than the plugin needs.  But SwiftFormat is detecting the theoretical difference.

We could mitigate this by updating our Package.swift:

```swift
    platforms: [
        .iOS(.v15),
        .macOS(.v12)
    ],
```

But we don't support macOS.  So I don't love this.

### SwiftLint vs SwiftFormat

SwiftFormat's ability to automatically make code changes is nice.  SwiftLint does have some limited capabilities for auto-correction.  But we can't use them yet.

The problem is with their SPM support.  At the moment, they only support a Build Tools version of the plugin.  These run on every build (which is very useful, and allows for in-line warnings and errors).  But Build Tool plugins cannot make changes to code.  They are read-only by design.

That's why SwiftFormat is a Command Line plugin (and requires manually running).

However, it looks like SwiftLint will be releasing a version of SwiftLint with support for both types of plugins: Build Tools and Command Line.

Ideally we would configure both:
- Build Tool Plugin: runs on every build, reports warnings and errors in-line
- Command Line Plugin: called manually by dev to autocorrect issues that can be corrected

### Waiting for the next major version of SwiftLint

Given the challenges above, I think it makes sense to strongly consider using either `SwiftLint` or `SwiftFormat` rather than both.

### Testing Steps

Try adding a force unwrapped variable somewhere, then try building.  You should see an build error.  Then try adding a comment to disable swiftlint on that line.  See the example in this PR.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,8 @@
+only_rules: # Rules to run
+  - force_unwrapping
+
+included:
+  - Sources
+
+# If true, SwiftLint will treat all warnings as errors.
+strict: true
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,12 +1,93 @@
 {
   "pins" : [
     {
+      "identity" : "collectionconcurrencykit",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/JohnSundell/CollectionConcurrencyKit.git",
+      "state" : {
+        "revision" : "b4f23e24b5a1bff301efc5e70871083ca029ff95",
+        "version" : "0.2.0"
+      }
+    },
+    {
+      "identity" : "cryptoswift",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/krzyzanowskim/CryptoSwift.git",
+      "state" : {
+        "revision" : "7892a123f7e8d0fe62f9f03728b17bbd4f94df5c",
+        "version" : "1.8.1"
+      }
+    },
+    {
+      "identity" : "sourcekitten",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/jpsim/SourceKitten.git",
+      "state" : {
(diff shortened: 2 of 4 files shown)


mokagio:
> SwiftFormat is giving some grief about SwiftLint. SwiftLint technically supports macOS 12. Our Package only specifies iOS .v15. For some reason, SwiftFormat (which apparently looks at the Package.swift file) is throwing an error about this:
the library 'Gravatar' requires macos 10.13, but depends on the product 'SwiftLintPlugin' which requires macos 12.0; consider changing the library 'Gravatar' to require macos 12.0 or later, or the product 'SwiftLintPlugin' to require macos 10.13 or earlier.

Just a note that it's possible to avoid this issue by constraining the dependency to be for the iOS platform only.

See https://github.com/wordpress-mobile/WordPressKit-iOS/pull/802

andrewdmontgomery:
Closing this for now.  We can revisit if we want to later.

PR #141, opened 2024-03-27
Adding `isGravatarURL` extension.

### LINKED ISSUE #117: Move `GravatarURL.isGravatarUrl()` to internal URL extension
As `GravatarURL` will be split into `AvatarURL` and `ProfileURL`, the generic check for `GravatarURL` will get insulated. 

`.isGravatarUrl(_ url: URL)` will check if a url is a Gravatar URL (it could be profile url, image, url, or any other).

etoledom:
Closes #117

### Description

This is a left-over from splitting `GravatarURL` struct.
There's no immediate benefit, but it could become handy in the future.

We might decide to not merge this and close the ticket as "won't do".

### Testing Steps

- 📗 CI


Sources/Gravatar/Extensions/URL.swift
```diff
@@ -9,4 +9,17 @@ extension URL {
             self.appendingPathComponent(path)
         }
     }
+
+    /// Wether this URL instance corresponds to a valid Gravatar URL.
```
pinarol: Just typo.
```suggestion
    /// Whether this URL instance corresponds to a valid Gravatar URL.
```

pinarol (approved):
LGTM 👍

PR #151, opened 2024-04-02
Move UIImageView extension and its depending types to GravatarUI

### LINKED ISSUE #149: Move UIImageView extension to GravatarUI
Move UIImageView extension to GravatarUI

pinarol:
Closes #149 

### Description

Move UIImageView extension and its depending types to GravatarUI

### Testing Steps

CI green


Demo/Demo/Gravatar-Demo/DemoUIImageViewExtensionViewController.swift
```diff
@@ -8,6 +8,7 @@
 import Foundation
 import UIKit
 import Gravatar
+import GravatarUI
```
etoledom, line 11: It would be great to only need to import `GravatarUI` and get everything from `Gravatar` also.
Like importing `UIKit` implicitly imports `Foundation`.

I guess we can add an @_exported import on `GravatarUI` if we want this to happen, though we can discuss it further without blocking this PR 👍

etoledom (approved):
Looking great!

A couple of details:

When building the docs, there are two warnings of broken links:
CleanShot 2024-04-02 at 22 08 28@2x

We need to restructure this whole description anyway, so I suggest only removing the links to get rid of the warnings, and we add a new ticket to update the docs and have a good doc base to continue the components work.

The other detail is on the code review. Not a blocker though 👍 

Feel free to merge 🎉

Demo/Demo/Gravatar-Demo/DemoUIImageViewExtensionViewController.swift
pinarol: Good idea. I created a new issue to track this. https://github.com/Automattic/Gravatar-SDK-iOS/issues/154

pinarol:
This is to track documentation update: https://github.com/Automattic/Gravatar-SDK-iOS/issues/155

pinarol:
I removed the links from GravatarWrapper. I removed the IdentifierProvider altogether because it doesn't exist in our code base. It must be a leftover from a PR.

PR #168, opened 2024-04-03
Add a design system to use when building the UI components

### LINKED ISSUE #150: Add a design system to Gravatar UI to define colors, fonts, spacings
Add a design system to Gravatar UI to define colors, fonts, spacings

pinarol:
Closes #150 

### Description

Defines colors, fonts, padding values to be used in the UI components.

Adds a public "Palette" struct and a `PaletteType` enum, this way it will be possible for 3rd party devs to inject their own colors. 

### Testing Steps

CI green.

### DIFF
--- a/Sources/GravatarUI/DesignSystem/Palette.swift
+++ b/Sources/GravatarUI/DesignSystem/Palette.swift
@@ -1,0 +1,94 @@
+import Foundation
+import UIKit
+
+public struct ForegroundColors {
+    public let primary: UIColor
+    public let primarySlightlyDimmed: UIColor
+    public let secondary: UIColor
+}
+
+public struct BackgroundColors {
+    public let primary: UIColor
+}
+
+public struct Palette {
+    public let name: String
+    public let foreground: ForegroundColors
+    public let background: BackgroundColors
+    public let avatarBorder: UIColor
+}
+
+public enum PaletteType {
+    case light
+    case dark
+    case system
+    case custom(() -> Palette)
+
+    public var palette: Palette {
+        switch self {
+        case .light:
+            Self.lightPalette
+        case .dark:
+            Self.darkPalette
+        case .system:
+            Self.systemPalette
+        case .custom(let paletteProvider):
+            paletteProvider()
+        }
+    }
+
+    public var name: String {
+        palette.name
+    }
+
+    static var systemPalette: Palette {
+        .init(
+            name: "System Default",
+            foreground: .init(
+                primary: UIColor(
+                    light: lightPalette.foreground.primary,
+                    dark: darkPalette.foreground.primary
+                ),
+                primarySlightlyDimmed: UIColor(
+                    light: lightPalette.foreground.primarySlightlyDimmed,
+                    dark: darkPalette.foreground.primarySlightlyDimmed
+                ),
+                secondary: UIColor(
+                    light: lightPalette.foreground.secondary,
+                    dark: darkPalette.foreground.secondary
+                )
+            ),
+            background: .init(primary: UIColor(
+                light: lightPalette.background.primary,
+                dark: darkPalette.background.primary
+            )),
+            avatarBorder: .porpoise
+        )
+    }
+
+    public static var lightPalette: Palette {
+        .init(
+            name: "Light",
+            foreground: .init(
+                primary: .black,
+                primarySlightlyDimmed: .gravatarBlack,
+                secondary: .dugong
+            ),
+            background: .init(primary: .white),
+            avatarBorder: .porpoise
+        )
+    }
+
+    static var darkPalette: Palette {
+        .init(
+            name: "Dark",
+            foreground: .init(
+                primary: .white,
+                primarySlightlyDimmed: .white,
+                secondary: .snowflake60
+            ),
+            background: .init(primary: .gravatarBlack),
+            avatarBorder: .porpoise
+        )
+    }
+}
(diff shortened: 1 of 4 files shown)


Sources/GravatarUI/DesignSystem/UIFont+DesignSystem.swift
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+
+        static var footnote: UIFont { DynamicFontHelper.fontForTextStyle(.footnote, fontWeight: .regular) } // 13pt
+        static var caption: UIFont { DynamicFontHelper.fontForTextStyle(.caption1, fontWeight: .regular) } // 12pt
+    }
+}
+
+private enum DynamicFontHelper {
+    static func fontForTextStyle(_ style: UIFont.TextStyle, fontWeight weight: UIFont.Weight) -> UIFont {
```
etoledom: Small detail: Apple APIs usually break the function name into name(label in cases like this:
```suggestion
    static func font(forTextStyle style: UIFont.TextStyle, weight: UIFont.Weight) -> UIFont {
```
It basically changes where the `(` is, bit it makes it nicer to split it in different lines when needed:

```swift
DynamicFontHelper.font(
    forTextStyle: .body,
    weight: .bold
)
```

We are also in the context of font creation (by the function name) so, I'd say that the font `fontWeight` is not really needed 🤔

Sources/GravatarUI/DesignSystem/UIColor+DesignSystem.swift
```diff
@@ -0,0 +1,27 @@
...
+}
+
+extension UIColor {
+    // Generated names are from https://colornamer.robertcooper.me/
+    static let gravatarBlack: UIColor = .rgba(16, 21, 23)
+    static let snowflake: UIColor = .rgba(240, 240, 240)
+    static let porpoise: UIColor = .rgba(218, 218, 218)
+    static let snowflake60: UIColor = snowflake.withAlphaComponent(0.6)
+    static let dugong: UIColor = .rgba(112, 112, 112)
+
```
etoledom: I like the idea of naming the colors, but some names gaves me zero clue about how the color looks like :( 

`porpoise` and `dugong` in particular. I couldn't say if they are light, dark, reddish, bluish.. etc...

Maybe we can add a sufix with a general color to which it approximates the most?

Like `gugongGrey` (if it's near to a grey color. I checked it but I don't know. I'm color-blind so I'm the worst person to name colors 😅 )

Sources/GravatarUI/DesignSystem/Palette.swift
```diff
@@ -0,0 +1,94 @@
...
+            paletteProvider()
+        }
+    }
+
+    public var name: String {
+        palette.name
+    }
+
+    static var systemPalette: Palette {
+        .init(
```
etoledom: I'd add all these pre-defined instances as extensions to `Palette` itself.

`Palette.light`
`Palette.system`
etc...

Sources/GravatarUI/DesignSystem/UIFont+DesignSystem.swift
```diff
@@ -0,0 +1,46 @@
...
+        var fontDescriptor = UIFontDescriptor.preferredFontDescriptor(withTextStyle: style)
+        let traits = [UIFontDescriptor.TraitKey.weight: weight]
+        fontDescriptor = fontDescriptor.addingAttributes([.traits: traits])
+        let font = UIFont(descriptor: fontDescriptor, size: CGFloat(0.0))
+
+        if UIFont.isGravatarDynamicFontSizeEnabled {
+            return UIFontMetrics.default.scaledFont(for: font)
+        }
```
etoledom, line 43: Pretty nice option 🎉

etoledom (commented):
Looking great! Just a few comments.

I guess that the values for the colors and paddings were taken from the designs, right?
Also, I like that the user can style the components with their colors. I'm just thinking if this was agreed to do the same on Android?

Sources/GravatarUI/DesignSystem/UIColor+DesignSystem.swift
pinarol: That can be useful I thought that too actually. I had to search for dugong and porpoise first. But once I saw what they are I thought they were really good words to describe these colors. But I agree, it can still useful to add the color.

pinarol:
> I guess that the values for the colors and paddings were taken from the designs, right?

Yes.

> Also, I like that the user can style the components with their colors. I'm just thinking if this was agreed to do the same on Android?

We talked about giving all supported text related options. 3rd party devs should be able to set the color, font, etc of the label as they want. We didn't talk about such a palette structure, it doesn't have to be this way I think. There are different ways to achieve the same thing.

Sources/GravatarUI/DesignSystem/Palette.swift
pinarol: yes actually that would be better.

Sources/GravatarUI/DesignSystem/UIFont+DesignSystem.swift
pinarol: Yep, much better.

pinarol:
Ready for another check

etoledom:
> We didn't talk about such a palette structure

We can share this on a P2 to coordinate after merging. I think it's a good way to go, and I don't think we need to be super tight with Android as long as we both give the options for customization.

etoledom (approved):
Thank you for the changes 🙏 

Looking great

PR #169, opened 2024-04-03
Add label decorators for "About Me", "Personal Info", "Display Name"

pinarol:
Closes part of https://github.com/Automattic/Gravatar-SDK-iOS/issues/173

### Description

- ~~Adds UILabel decorators under the `gravatar` namespace for profile card fields: "About Me", "Personal Info", "Display Name"~~

Usage:

```
let label = UILabel(frame: frame)
Configure(label)
   .asDisplayName()
   .content(model)
   .palette(palette)

// model is anything that conforms to  "DisplayNameModel". Our UserProfile struct conforms to DisplayNameModel. 
// So in real life we'll pass UserProfile here.
```

More examples can be found in the unit tests.

- Adds snapshot-testing.

### Testing Steps

In this PR I just added some snapshot tests to test. In the next PR I am planning to add a new screen to the demo app.


Sources/GravatarUI/ProfileFields/PersonalInfoField.swift
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    var label: UILabel
+    init(label: UILabel) {
+        self.label = label
+    }
+
+    public func update(
+        with model: PersonalInfoModel,
+        lines: [PersonalInfoLine] = Self.defaultPersonalInfo,
```
pinarol: We can remove this `lines` param for now to match with Android. And make PersonalInfoLine internal.

pinarol:
This is ready for another look after updating with UILabel builders.

Sources/GravatarUI/ProfileFields/Model/AvatarIdentifierProvider.swift
```diff
@@ -0,0 +1,12 @@
...
+
+public protocol AvatarIdentifierProvider {
+    var avatarIdentifier: AvatarIdentifier { get }
+}
+
+extension UserProfile: AvatarIdentifierProvider {
+    public var avatarIdentifier: Gravatar.AvatarIdentifier {
+        .hashID(HashID(hash))
+    }
+}
```
etoledom, line 12: Maybe we can use `Identifiable` instead?

```swift
extension UserProfile: Identifiable {
    public var id: Gravatar.AvatarIdentifier {
        .hashID(HashID(hash))
    }
}
```

Or if the idea is to have this specific field name `avatarIdentifier`, I think the name `AvatarIdentifiable` is a bit more suited for the protocol.

Though having UserProfile Identifiable, would also be a good thing in case we want it to be a model on Diffable Snapshots for example.

Sources/GravatarUI/ProfileFields/Model/PersonalInfoModel.swift
```diff
@@ -0,0 +1,51 @@
...
+            model.currentLocation
+        case .separator(let string):
+            string
+        }
+    }
+
+    static var defaultSeparator: PersonalInfoBuildingBlock {
+        .separator("・")
+    }
+}
```
etoledom: Something I noticed about this:




If the content is empty, we still add the separators, and it doesn't look right.

etoledom (approved):
Looking great! 

Just one small detail that I found while testing. But let's continue moving forward with the rest 👍

Sources/GravatarUI/ProfileFields/Model/PersonalInfoModel.swift
pinarol: Ah yes, let me fix it. It was in my mind to add this case to unit tests apparently I forgot.... :(

PR #171, opened 2024-04-03
Remove test-spec from the podspecs

pinarol:
Closes #

### Description

Our CI already uses spm configured test targets. Having the test-spec in the podspec is becoming limiting to us. 

For example: We want to add a common spm target to share code between Gravatar Tests and GravatarUI Tests. podspec doesn't let us have a local folder as a dependency. validate-podspecs CI job fails.

And it's also a pain every time we want to add a normal dependency to the test target.

### Testing Steps

CI green.


etoledom (approved):
❤️

AliSoftware (approved):

PR #172, opened 2024-04-03
Add basic profile card

### LINKED ISSUE #173: Add basic profile card - (1st iteration of the profile card)
Add basic profile card with:

- avatar
- display name
- personal info
- about me

Screenshot 2024-04-04 at 17 21 42

Figma: https://www.figma.com/file/AzU51EzjGYFJPtLKG2GBIr/SDK-Components?type=design&node-id=8013-48577&mode=design&t=NRR6Q3rPKJoMi0Tp-0

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/173

### Description

This is just the initial version of the profile card view. I think we'll do many iterations on this.

- Adds the basic profile card
- Adds snapshot tests
- Adds demo page

<image src="https://github.com/Automattic/Gravatar-SDK-iOS/assets/5032900/b90a8658-cd8e-4a95-9e7a-c3c2c494e7ec" width=300/>

### Testing Steps

Demo app > profile card


Sources/GravatarUI/ProfileCard/ProfileCardView.swift
```diff
@@ -0,0 +1,131 @@
...
+        return label
+    }()
+
+    public var paletteType: PaletteType {
+        didSet {
+            refresh(with: paletteType)
+        }
+    }
+
+    public let containerLayoutGuide = UILayoutGuide()
```
pinarol: This is just to help with adding paddings around this view without the need for creating another content view. I put a sample usage in the demo app.

Sources/GravatarUI/ProfileCard/ProfileCardView.swift
```diff
@@ -0,0 +1,131 @@
...
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    public func update(with model: ProfileCardModel) {
+        aboutMeLabel.gravatar.aboutMe.update(with: model, paletteType: paletteType)
+        displayNameLabel.gravatar.displayName.update(with: model, paletteType: paletteType)
+        personalInfoLabel.gravatar.personalInfo.update(with: model, paletteType: paletteType)
+    }
```
etoledom: About this pattern:
`aboutMeLabel.gravatar.aboutMe.update(...`
While I was reviewing[ the previous PR](https://github.com/Automattic/Gravatar-SDK-iOS/pull/169), I wasn't sure if this patter was that needed. It seems nice but kinda an overuse of `.gravatar` for something that seems a style helper only.

Now checking it here, It does look pretty comfortable to use, so I'm happy to keep it and use it like this internally, but not everyone likes these extensions over UIKit types. Even KingFisher offers [an alternative](https://github.com/onevcat/Kingfisher?tab=readme-ov-file#method-chaining).

Maybe we could also offer an alternative, something like: `DisplayNameField.update(label, with: model, paletteType: palette)`

etoledom: Another different topic regarding `public func update(with model: ProfileCardModel)`

On `ProfileCardView` it would be amazing to adopt the `UIContentView` protocol.
By adopting it, this method might be replaced by setting a configuratio instance (which could also be a `ProfileCardModel`

etoledom (commented):
Looks pretty good!

I moved forward with this PR to understand better how the implementation of https://github.com/Automattic/Gravatar-SDK-iOS/pull/169 looks in action.

Independent of the dicussion about the `gravatar` extension, just a few things to note from this PR:

- We could try to implement UIContentView, see if it matches our pattern, and decide early if we want our components to adopt it. Personally I think it's a helpful pattern to adopt in modern UIKit.

- The layout looks different from the one from the design:

CleanShot 2024-04-04 at 13 02 49@2x

It's fine to iterate on this in future PRs, but maybe in this case is not needed to do the snapshots test just yet, since we know they are going to change soon.

- On the example screen, it would be good to set the controller's background to a different color than the card's background, to notice better the boundaries. Some very soft gray would do (maybe system's secondary background)

pinarol:
- Updated with label builders.

- This is one of the existing designs from [figma](https://www.figma.com/file/AzU51EzjGYFJPtLKG2GBIr/SDK-Components?type=design&node-id=8013-48577&mode=design&t=NRR6Q3rPKJoMi0Tp-0). 

- Let's explore UIContentView adoption together. I don't have any clear picture about how it could look like. We can do it in a different PR I think.

Sources/GravatarUI/ProfileCard/ProfileCardView.swift
etoledom: Nice!

etoledom (approved):
Looks good!

My only comment is probably related to the previous PR, I left the comment there: 
https://github.com/Automattic/Gravatar-SDK-iOS/pull/169#discussion_r1553433221

PR #180, opened 2024-04-04
Adding customURL option to DefaultAvatarOption

### LINKED ISSUE #152: Add new customURL options to `DefaultAvatarOption`
1. Custom: A URL of user’s choice, in case they want to use their own fallback image. (Android already has [such option](https://github.com/Automattic/Gravatar-SDK-Android/blob/48c8659948683ddeef3c8147d53b6332d02843ac/gravatar/src/main/java/com/gravatar/DefaultAvatarOption.kt#L68), and our backend supports it, so we can)

~~2. Gravatar logo, it fallbacks to https://www.gravatar.com/avatar/00000000000000000000000000000000?s=200 (“s” parameter to be adjusted) This will be our default in the profile card component.~~

Update: I completely missed that the gravatar logo is already the default behavior when one doesn't set this parameter. Although I like the idea of making this an explicit option. Let's discuss if this is a good idea.

etoledom:
Closes #152 

### Description

Adding a custom URL option to DefaultAvatarOption.

### Testing Steps

Using the Gravatar Demo app:

- Go to `ImageDownload - Networking` page
- Enter an invalid string as `Email`
- Search on internet any image and get its URL
- Paste the URL on the field `Set custom avatar URL`
- Press Enter to save the change
- Check that the Default Avatar Option now says `Custom URL` on the button abov.
- Tap on `Fetch Avatar`.
  - Check that the image from the URL is displayed


pinarol:
Looks and works good but I wonder if we should be encoding the URL as described in the [docs](https://docs.gravatar.com/general/images/).

> The URL should be URL-encoded to ensure that it carries across correctly, for example:

Having said that, I see it working both ways 🤷‍♀️ 

https://gravatar.com/avatar/00000000000000000000000000000000?d=https://3.img-dpreview.com/files/p/E~C83x0S500x500T1200x1200~articles/0896048182/jpegmini-logo.jpeg


https://gravatar.com/avatar/00000000000000000000000000000000?d=https%3A%2F%2F3.img-dpreview.com%2Ffiles%2Fp%2FE~C83x0S500x500T1200x1200~articles%2F0896048182%2Fjpegmini-logo.jpeg

pinarol (approved):
👍

etoledom:
> Looks and works good but I wonder if we should be encoding the URL as described in the [docs](https://docs.gravatar.com/general/images/).

When I tried that yesterday, the query get's double encoded and looks like this:
`http%253A%252F%252F....` and the request failed.

Now I gave it a second look and implemented a solution for it. Better to do what the docs say 👍

Sources/Gravatar/AvatarURL.swift
```diff
@@ -55,10 +55,10 @@ extension URL {
         guard var components = URLComponents(url: self, resolvingAgainstBaseURL: false) else {
             return nil
         }
-        components.queryItems = options.queryItems
+        components.percentEncodedQueryItems = options.queryItems
```
etoledom: This is the way to avoid double encoding, but we will need to be careful since URLComponents won't encode other options.
So far we control everything that goes here except from the custom URL, which we encode manually, but we'll need to keep this in mind if we ever add new options.

pinarol: Actually, if it already gets encoded then we don't need to worry. I thought there was no encoding at all. We can revert this imo, this would be difficult to remember in the future...

PR #183, opened 2024-04-08
Adding `LargeProfileSummaryView`

### LINKED ISSUE #184: Create `LargeProfileSummaryView`

etoledom:
Closes #184 

### Description

Implementing `LargeProfileSummaryView`.

**Note**: This is a simple copy/paste of the `LargeProfileView`, and changing the layout as needed. I tested a different approach later on to allow sharing this common code between components. This PR is not intended to try this code sharing approach.

Design | Implementation


Sources/GravatarUI/ProfileFields/PersonalInfoField.swift
```diff
@@ -37,4 +39,10 @@ public struct PersonalInfoBuilder {
         label.textColor = paletteType.palette.foreground.secondary
         return self
     }
+
+    @discardableResult
+    public func alignment(_ alignment: NSTextAlignment) -> PersonalInfoBuilder {
+        label.textAlignment = alignment
+        return self
+    }
```
pinarol, line 47: I think it's better if we don't provide "public" methods for things that are NOT Gravatar related in these builders. UILabel has many other properties like this... We can easily end up with many public methods... And the more public types we provide more things to worry when changing something. So how about we keep this one internal, WDYT?

Sources/GravatarUI/ProfileView/LargeProfileSummaryView.swift
```diff
@@ -0,0 +1,120 @@
+import Gravatar
+import UIKit
+
+public protocol ProfileCardSummaryModel: DisplayNameModel, PersonalInfoModel, AvatarIdentifierProvider {}
```
pinarol, line 4: I wonder if we should drop "card" from these model names as well? 🤔 Like:

> ProfileSummaryModel

And the other one that LargeProfileView uses:

> ProfileModel

WDYT?

Demo/Demo/Gravatar-Demo/DemoLargeProfileViewController.swift
```diff
@@ -32,10 +32,16 @@ class DemoLargeProfileViewController: UIViewController {
...
     }()
-    
+
+    lazy var largeProfileSummaryView: LargeProfileSummaryView = {
+        let view = LargeProfileSummaryView(frame: .zero, paletteType: preferredPaletteType)
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.avatarImageView.gravatar.activityIndicatorType = .activity
+        return view
+    }()
+//
```
pinarol, line 42: `//` can be removed

Demo/Demo/Gravatar-Demo/DemoLargeProfileViewController.swift
```diff
@@ -44,7 +50,7 @@ class DemoLargeProfileViewController: UIViewController {
 
         return stack
     }()
-
+//
```
pinarol, line 53: `//` can be removed

Sources/GravatarUI/ProfileView/LargeProfileSummaryView.swift
```diff
@@ -0,0 +1,120 @@
...
+    public func update(with model: ProfileCardSummaryModel) {
+        Configure(displayNameLabel).asDisplayName().content(model).palette(paletteType)
+        Configure(personalInfoLabel).asPersonalInfo().content(model).palette(paletteType).alignment(.center)
+    }
+
+    public func loadAvatar(
+        with avatarIdentifier: AvatarIdentifier,
+        placeholder: UIImage? = nil,
+        rating: Rating? = nil,
+        preferredSize: CGSize? = nil,
```
pinarol, line 86: Again I notice this here but it's same in `LargeProfileView`. I think we should remove `preferredSize` parameter from this method. The caller party don't control the size of the UImageView so they don't have a reason to pass a `preferredSize`.

Sources/GravatarUI/ProfileView/LargeProfileSummaryView.swift
```diff
@@ -0,0 +1,120 @@
...
+        self.paletteType = .system
+        super.init(frame: frame)
+        addSubview(rootStackView)
+        layoutMargins = UIEdgeInsets(top: .margin, left: .margin, bottom: .margin, right: .margin)
+        NSLayoutConstraint.activate([
+            layoutMarginsGuide.topAnchor.constraint(equalTo: rootStackView.topAnchor),
+            layoutMarginsGuide.leadingAnchor.constraint(equalTo: rootStackView.leadingAnchor),
+            layoutMarginsGuide.trailingAnchor.constraint(equalTo: rootStackView.trailingAnchor),
+            layoutMarginsGuide.bottomAnchor.constraint(equalTo: rootStackView.bottomAnchor),
+        ])
```
pinarol, line 63: I know this is taken from `LargeProfileView` but I think we should also add `refresh(with: paletteType)` here like we do in the convenience init, just for sake of consistency... 

Same for `LargeProfileView`.

pinarol (commented):
Looks good. 🎉  I dropped some small comments, some of them also apply to `LargeProfileView`. We can talk about those quickly and then merge this.

pinarol (approved):

Sources/GravatarUI/ProfileFields/PersonalInfoField.swift
etoledom: Will do on a next PR 👍

Sources/GravatarUI/ProfileView/LargeProfileSummaryView.swift
etoledom: Will do on a next PR 👍

Sources/GravatarUI/ProfileView/LargeProfileSummaryView.swift
etoledom: Will do on a next PR 👍

PR #188, opened 2024-04-10
Adding `ProfileSummaryView`

### LINKED ISSUE #185: Create `ProfileSummaryView`

etoledom:
Closes #185 

### Description

Adding `ProfileSummaryView`.

On this PR I tested a new approach to share code between components.

A new base class which holds all needed subviews for any component, plus the root stack view which is added as subviews.

Subclasses of this base class will choose which subviews will be added to the hierarchy and will build the layout.
Subclasses can also have their own stack views to aid building the layout.

### Testing Steps

- An instance of this new view was added to the Components example in the app


Sources/GravatarUI/ProfileView/ProfileComponentView.swift
```diff
@@ -0,0 +1,130 @@
+import Gravatar
+import UIKit
+
+open class ProfileComponentView: UIView {
```
pinarol, line 4: ProfileComponentView > BaseProfileView

pinarol (approved):
LGTM 👍

Sources/GravatarUI/ProfileView/ProfileComponentView.swift
etoledom: Will do on a next PR 👍

PR #192, opened 2024-04-12
SwiftUI exploration with Profile view.

### LINKED ISSUE #191: Add Profile View to SwiftUI

etoledom:
Closes #191 

CleanShot 2024-04-12 at 15 55 24@2x

### DIFF
--- a/Demo/Demo/Gravatar-SwiftUI-Demo/ContentView.swift
+++ b/Demo/Demo/Gravatar-SwiftUI-Demo/ContentView.swift
@@ -1,21 +1,58 @@
-//
-//  ContentView.swift
-//  Demo
-//
-//  Created by Andrew Montgomery on 1/19/24.
-//
-
 import SwiftUI
+import GravatarUI
+import Gravatar
 
 struct ContentView: View {
+    @State var model: ProfileCardModel?
+    @State var isLoading: Bool = false
+    @State var email: String = ""
+    @State var error: String? = nil
+
     var body: some View {
-        VStack {
-            Image(systemName: "globe")
-                .imageScale(.large)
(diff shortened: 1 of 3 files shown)


Demo/Demo/Gravatar-SwiftUI-Demo/ContentView.swift
```diff
@@ -1,21 +1,58 @@
...
+                .textFieldStyle(.roundedBorder)
+                .keyboardType(.emailAddress)
+
+                if isLoading {
+                    VStack {
+                        ProgressView()
+                    }
+                } else {
+                    Profile(model: $model).frame(height: 220)
+                }
```
etoledom, line 31: This is a problem I ran into:

Here we need to specify the profile view height. If height is not given, the view will expand to all the available vertical space.

One way to fix this (as I read) is to manually calculate intrinsic content size (which is (-1, -1) as for now). I did a fast test of this and it works, but to properly calculate the height of the view is cumbersome, error-prone, and defeats the purpose of auto-layout.

Another way I tested was to `profileView.translateAutoresizingMaskIntoContrains = false` on the `UIViewRepresentable` definition.

This works on calculating the height automatically, but we lose the width expansion. The rootStackView, in this case, will use as minimum width as possible, which is currently given by the bottom bar (all the width of the buttons together), and it will ignore the constraints to its superView (for some reason I don't understand).



As far as I understand, SwiftUI will use frames, so it's not good to set masks translation into constraints to false on the main view. But we still need to figure out a way to automatically calculate the view height.

etoledom:
Closed via https://github.com/Automattic/Gravatar-SDK-iOS/pull/678

PR #189, opened 2024-04-10
Adding account buttons `ProfileView`

etoledom:
Closes #179 

### Description

This PR adds the account buttons to `ProfileView`
- Using the shared base class, it's simple also to implement the profile button to finish the design.

Design | Implementation

### Testing Steps

- Using the demo app

PR #196, opened 2024-04-15
Set icon size programmatically so that we don't need to manually adjust pdf icon sizes

### LINKED ISSUE #197: Set Account button size programatically

pinarol:
Follow up work for https://github.com/Automattic/Gravatar-SDK-iOS/pull/189
Closes: #197

### Description

We had to manually adjust icon size according to the designs before. After this PR we won't need that.

By default the UIButton's UIImageView has an intrinsic size equal to its image's size. But we don't always want to rely on icon's size, especially if it's a vector image that should be able to be rendered in any size. So we explicitly set UIImageView size. I think there should be a better way to do this but 🤷‍♀️.

### Testing Steps

- Change `accountIconLength` to any size in the code and check if the icon size adjusts fine and the icon quality is good.


etoledom (approved):
Works as described! 🎉 

Sad the need to constraint the button's image, but it does seem to be needed.
Let's go with this 👍

PR #214, opened 2024-04-17
Adding profile view buttons action delegate

### LINKED ISSUE #199: Delegate Account and Profile buttons action

etoledom:
Closes #199 

### Description

Adding API to react to user button taps on the Profile views.

Added to all Profile views and to the Profile configuration.

### Testing Steps

- On the Profile Card demo screen:
  - Search for a profile with some verified account
  - Test that all buttons (view profile and account buttons) open the corresponding link.
- On the Profile Card **Configuration** demo screen:
  - Add an account with some verified accounts to the list
  - Test that all buttons (view profile and account buttons) open the corresponding link.


Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -211,3 +223,8 @@ open class BaseProfileView: UIView, UIContentView {
         }
     }
 }
+
+public protocol ProfileViewDelegate: NSObjectProtocol {
+    func didTapOnProfileButton(with style: ProfileButtonStyle, profileURL: URL?)
+    func didTapOnAccountButton(with accountModel: AccountModel)
```
pinarol: Can we pass the originating BaseProfileView instance in the first param? That's usually the convention for delegate methods. (That way if there are multiple objects they can be distinguishable from each other.)

pinarol (approved):
LGTM 👍 I am just dropping 1 comment.

PR #219, opened 2024-04-19
Adding edit profile button style

### LINKED ISSUE #218: Add edit profile button style

etoledom:
Closes #218 

### Description

This PR adds support for the `Edit profile ->` button style.

### Testing Steps

- The Profile Card demo screen now shows the last two profile card with the edit profile button, while the two first cards show the view profile  button.
- Tap on `Edit profile` and check that it redirects to the Gravatar login page.
- Tap on `View profile` and check that it redirects to the profile page.

**Note:** I wasn't able to login in the webview. This is something important to check and test later on. Plus the experience wasn't good.


pinarol (approved):
LGTM 👍

PR #220, opened 2024-04-19
Fix extra dot separator with pronouns only model

### LINKED ISSUE #212: Extra dot separator with only pronoun on profile
When the profile has only a pronoun to display in personal info, the UI draws an extra dot separator:

Image

etoledom:
Closes #212 

### Description

This PR solves the issue of having an extra separator when there's just one personal info field in the model.

I propose a different approach, in which a separator is not a building blocks, but we add the separator in between building blocks using `Array.joined()` function.

In this way we don't need to worry about edge cases, and much of the code needed for checking such cases is not needed any more.

Users can still add a custom separator using:

```swift 
public func content(
    _ model: PersonalInfoModel,
    lines: [PersonalInfoLine],
    separator: String?
) -> PersonalInfoBuilder
```

### Testing Steps

- Added new unit test cases.
- Smoke test the demo app.


pinarol (approved):
This is nice. "Array.joined(...)", of course, why didn't i think that...

Issue #211, opened 2024-04-17
Add LargeProfileSummaryView ghost UI


pinarol:
<img width="126" alt="Screenshot 2024-04-17 at 13 38 04" src="https://github.com/Automattic/Gravatar-SDK-iOS/assets/5032900/5f474554-a255-4b1c-a569-a18e85ce9433">

https://www.figma.com/file/AzU51EzjGYFJPtLKG2GBIr/SDK-Components?type=design&node-id=8870%3A13260&mode=design&t=lAQrgXEvgO4IJUl3-1

Issue #210, opened 2024-04-17
Add LargeProfileView ghost UI


pinarol:
<img width="129" alt="Screenshot 2024-04-17 at 13 36 55" src="https://github.com/Automattic/Gravatar-SDK-iOS/assets/5032900/ca2c7259-65bb-4022-94d4-d7f5279ca320">

https://href.li/?https://www.figma.com/file/AzU51EzjGYFJPtLKG2GBIr/SDK-Components?type=design&node-id=8870%3A13260&mode=design&t=lAQrgXEvgO4IJUl3-1

Issue #209, opened 2024-04-17
Add ProfileSummaryView ghost UI


pinarol:
<img width="224" alt="Screenshot 2024-04-17 at 13 35 44" src="https://github.com/Automattic/Gravatar-SDK-iOS/assets/5032900/021a78f0-6b19-404a-ab11-34839789fc4e">

https://href.li/?https://www.figma.com/file/AzU51EzjGYFJPtLKG2GBIr/SDK-Components?type=design&node-id=8870%3A13260&mode=design&t=lAQrgXEvgO4IJUl3-1

Issue #208, opened 2024-04-17
Add ProfileView ghost UI


pinarol:
<img width="204" alt="Screenshot 2024-04-17 at 13 34 50" src="https://github.com/Automattic/Gravatar-SDK-iOS/assets/5032900/1fcbf3fb-b261-4f11-b3b4-59beac0b7b3f">

https://href.li/?https://www.figma.com/file/AzU51EzjGYFJPtLKG2GBIr/SDK-Components?type=design&node-id=8870%3A13260&mode=design&t=lAQrgXEvgO4IJUl3-1

PR #221, opened 2024-04-19
Add Placeholder UI & loading state & profile view controller 

### LINKED ISSUE #213: Add a view controller for profile views

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/213 https://github.com/Automattic/Gravatar-SDK-iOS/issues/211 https://github.com/Automattic/Gravatar-SDK-iOS/issues/210 https://github.com/Automattic/Gravatar-SDK-iOS/issues/209 https://github.com/Automattic/Gravatar-SDK-iOS/issues/208 

### Description

- Adds ghost ui. I designed the placeholder and the activity indicator as separate things. Placeholder can be viewed without the loading animations(for example if the service call fails, the placeholder will remain but the animations will stop). One can also customize the placeholder & activity indicator.
- I didn't set the bottom sheet to use intrinsic size in this PR. It uses predefined size.

- Adds a ProfileViewController to test the standalone usage of our profile views. ProfileViewController does the service call and emits the view model state changes & updates the view.

### Testing Steps

Check these 2 pages in the demo app:

- Profile View Controller Presentations
- Profile Card


pinarol:
I am splitting these changes into smaller pieces.

PR #222, opened 2024-04-22
Add placeholder displayers for profile fields

pinarol:
Closes part of ghost ui tasks #208 #209 #210 #211

### Description

- Adds `PlaceholderDisplaying` protocol to define a placeholder displayer for profile fields.
- Adds various implementations for `PlaceholderDisplaying`
- Adds `ProfileViewPlaceholderDisplayer`. In this PR I didn't use it in `BaseProfileView` yet.
- This is just the beginning of Ghost UI changes, more PRs are coming.

### Testing Steps

- CI green


Tests/GravatarUITests/TestPlaceholderDisplayers.swift
```diff
@@ -0,0 +1,101 @@
+import Gravatar
+@testable import GravatarUI
```
pinarol, line 2: I wasn't sure to add tests for internal api, but I felt like they can be useful to describe what these new types are doing. Not sure if we would want to keep these in the longer term.

Sources/GravatarUI/ProfileView/Placeholder/ProfileViewPlaceholderDisplaying.swift
```diff
@@ -0,0 +1,87 @@
+import UIKit
```
pinarol, line 1: I didn't add the part where we use `ProfileViewPlaceholderDisplayer` inside `BaseProfileView` in this PR because then the PR started to get really big. This is coming with the next PR.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -2,16 +2,27 @@ import Gravatar
...
 
     open var avatarLength: CGFloat {
         Constants.avatarLength
     }
 
+    public enum PlaceholderColorPolicy {
+        /// Gets the placeholder colors from the current palette.
+        case currentPalette
+        /// Custom colors. You can also pass predefined colors from any ``Palette``. Example:  `Palette.light.placeholder`.
+        case custom(PlaceholderColors)
```
pinarol, line 20: Our designer requested that the placeholder colors can be customizable independent of the current palette type. Because if the 3rd party app applies a custom background (like the pocket casts one in the design) the placeholder colors can look ugly. So even though the palette is dark, we may want to use colors from the light palette, or any other color. So i did it this way.

Sources/GravatarUI/ProfileView/LargeProfileSummaryView.swift
```diff
@@ -4,6 +4,7 @@ import UIKit
 public class LargeProfileSummaryView: BaseProfileView {
     private enum Constants {
         static let avatarLength: CGFloat = 132.0
+        static let displayNamePlaceholderHeight: CGFloat = 32
```
etoledom, line 7: This is fine for now but at some point we might want to look into a proper height depending on preferred content size category.

Sources/GravatarUI/ProfileView/Placeholder/ProfileViewPlaceholderDisplaying.swift
```diff
@@ -0,0 +1,87 @@
...
+                height: 14,
+                widthRatioToParent: 0.8
+            ),
+            ProfileButtonPlaceholderDisplayer(
+                baseView: view.profileButton,
+                color: color,
+                cornerRadius: 8,
+                height: 16,
+                widthRatioToParent: 0.2
+            ),
```
etoledom, line 60: I'm just wondering how were those `widthRatioToParent` calculated?

etoledom (approved):
Looking good!
Let's keep rollin' 🚢

Sources/GravatarUI/ProfileView/Placeholder/ProfileViewPlaceholderDisplaying.swift
pinarol: They are not calculated, they are just how designs look (more or less).

Sources/GravatarUI/ProfileView/LargeProfileSummaryView.swift
pinarol: Yeah that can be a good idea 🤔.

PR #223, opened 2024-04-22
Adding docs to branch directory

etoledom:
Closes #

### Description

Build and publish documentation for GravatarUI, alongside Gravatar SDK.

We will keep building the docs on trunk, but it won't be versioned.

Next step is to publish the packages on [Swift Package Index](https://swiftpackageindex.com/), where it will build and publish a versioned page for both our targets. For this is the new `.spi.yml` file.

### Testing Steps

These pages should exist:
- https://automattic.github.io/Gravatar-SDK-iOS/gravatarui/documentation/gravatarui/
- https://automattic.github.io/Gravatar-SDK-iOS/gravatar/documentation/gravatar/


pinarol (approved):
LGTM 👍 Thanks a lot for looking into this.

PR #224, opened 2024-04-22
Add placeholder mechanism to BaseProfileView

pinarol:
Closes part of ghost ui tasks #208 #209 #210 #211

### Description

Adds placeholder mechanism to BaseProfileView. The placeholders shows when the given model is nil, hidden when the model has data.
 
### Testing Steps

Mostly done by CI but also "Demo app > Profile card"


Sources/GravatarUI/ProfileView/LargeProfileView.swift
```diff
@@ -31,32 +28,48 @@ public class LargeProfileView: BaseProfileView {
         return stack
     }()
 
-    override public init(frame: CGRect) {
-        super.init(frame: frame)
-        [topStackView, displayNameLabel, personalInfoLabel, aboutMeLabel, bottomStackView].forEach(rootStackView.addArrangedSubview)
+    override public func arrangeSubviews() {
+        super.arrangeSubviews()
+        [topStackView, displayNameLabel, personalInfoLabel, aboutMeLabel, aboutMePlaceholderLabel, bottomStackView, .spacer()]
```
pinarol: I added a spacer at the end here and in other profile views. Otherwise it's not predictable what happens if the container view is bigger than our profile view, it can put a space in a random place and it doesn't look well. At least this way the space is only at the bottom and the content stays same. 🤔 (I noticed this when testing things in a fixed height bottom sheet)

Sources/GravatarUI/DesignSystem/Palette.swift
```diff
@@ -5,10 +5,18 @@ public struct ForegroundColors {
     public let primary: UIColor
     public let primarySlightlyDimmed: UIColor
     public let secondary: UIColor
+    public init(primary: UIColor, primarySlightlyDimmed: UIColor, secondary: UIColor) {
+        self.primary = primary
+        self.primarySlightlyDimmed = primarySlightlyDimmed
+        self.secondary = secondary
+    }
```
pinarol, line 12: I noticed that we didn't make any initializers public so there was no way to create a custom palette from outside the project. So I am adding public inits.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -170,6 +185,13 @@ open class BaseProfileView: UIView, UIContentView {
             layoutMarginsGuide.bottomAnchor.constraint(equalTo: rootStackView.bottomAnchor),
         ])
         refresh(with: paletteType)
+        placeholderDisplayer?.setup(using: self)
+        arrangeSubviews()
+        togglePlaceholder() // We should call this after subviews are added (which means after `arrangeSubviews()`)
+    }
+
+    open func arrangeSubviews() {
+        // Subclasses should override and add fields to the `rootStackView`.
```
etoledom: We can add an assertion here with a "Must be overridden" message

Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -285,6 +308,34 @@ open class BaseProfileView: UIView, UIContentView {
...
+        displayNameLabel.text = nil
+        personalInfoLabel.text = nil
+        aboutMeLabel.text = nil
+        updateAccountButtons(with: nil)
+        avatarImageView.image = nil
+    }
+
+    func shouldShowPlaceholder() -> Bool {
+        model == nil
+    }
```
etoledom: Maybe this could be a `var`?

Sources/GravatarUI/ProfileView/LargeProfileView.swift
etoledom: > (I noticed this when testing things in a fixed height bottom sheet)

A good approach to avoid this is to not pin the bottom anchor to the controller's view bottom. In this case the view's height is its intrinsic content height.

Having this spacer here could potentially mess up the content height bottom sheet or the popover. Though there are chances that it will just work with the spacer resulting to zero height, but it could also become ambiguous 🤔 Would need testing.

etoledom (approved):
Looking great!
A couple of comments but nothing blocking.

Sources/GravatarUI/ProfileView/LargeProfileView.swift
```diff
@@ -30,31 +28,51 @@ public class LargeProfileView: BaseProfileView {
...
+        update(with: config.model)
+    }
+
+    override public func showPlaceholders() {
+        super.showPlaceholders()
+        rootStackView.setCustomSpacing(.DS.Padding.single, after: displayNameLabel)
+        rootStackView.setCustomSpacing(.DS.Padding.double, after: personalInfoLabel)
+        rootStackView.setCustomSpacing(.DS.Padding.single, after: aboutMeLabel)
+        rootStackView.setCustomSpacing(.DS.Padding.double, after: aboutMePlaceholderLabel)
+    }
```
etoledom, line 67: A bit sad that we need to change spaces, but I see how it is necessary 👍

Sources/GravatarUI/ProfileView/LargeProfileView.swift
pinarol: Actually you are right, I was confused by the usage within the scroll view but I experimented further and i see that i didn't have to worry. I'll remove those spacers at the bottom. They don't cause anything odd but there's no reason to keep them.

Sources/GravatarUI/ProfileView/LargeProfileView.swift
pinarol: Yeah unfortunately.

pinarol:
All the comments have been addressed. 👍

PR #227, opened 2024-04-23
Update .spi.yml with platform specification.

etoledom:
### Description
Adding platform specification to .spi.yml


pinarol (approved):
LGTM 👍

PR #228, opened 2024-04-23
Add ProfileViewController & ProfileViewModel

### LINKED ISSUE #213: Add a view controller for profile views

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/213

### Description

- Adds a simple ProfileViewController & ProfileViewModel. ProfileViewController works based on `ProfileViewConfiguration` so we don't need to worry about which profile view it should use inside 🎉 .
- Adds a demo screen to open the ProfileViewController in a bottom sheet. 
- Updated the demo screens a bit to share some code.
- Added some avatar related options to the `ProfileViewConfiguration`.

### Testing Steps

- Demo app > Profile view controller


Demo/Demo/Gravatar-Demo/DemoBaseProfileViewController.swift
```diff
@@ -0,0 +1,121 @@
+import UIKit
+import Gravatar
+import GravatarUI
+
+open class DemoBaseProfileViewController: UIViewController {
```
pinarol, line 5: I moved some common code here.

Demo/Demo/Gravatar-Demo/DemoProfileViewsViewController.swift
```diff
@@ -0,0 +1,122 @@
+import UIKit
+import Gravatar
+import GravatarUI
+import SafariServices
+
+class DemoProfileViewsViewController: DemoBaseProfileViewController {
```
pinarol, line 6: I renamed "DemoLargeProfileViewController" as "DemoProfileViewsViewController" and made it inherit from `DemoBaseProfileViewController`. No functionality change.

Demo/Demo/Gravatar-Demo/DemoProfileViewControllerPresentingViewController.swift
```diff
@@ -0,0 +1,87 @@
+import UIKit
+import Gravatar
+import GravatarUI
+
+@MainActor
+class DemoProfileViewControllerPresentingViewController: DemoBaseProfileViewController, UISheetPresentationControllerDelegate {
```
pinarol: > DemoProfileViewControllerPresentingViewController

This is a horrible name 😢. Maybe we can find something better.

Sources/Gravatar/Identifiers/ProfileIdentifier.swift
```diff
@@ -4,7 +4,7 @@ import Foundation
 ///
 /// The `ProfileService` uses an `ProfileIdentifier` to specifiy which avatar to fetch.  The `ProfileService` allows specifying a profile
 /// by either its email address or the hashID
-public enum ProfileIdentifier {
+public enum ProfileIdentifier: Equatable {
```
pinarol, line 7: oh these `Equatable`s are some leftovers from something I tried. But i think we can keep them, because these are identifiers and they indeed should be `Equatable`.

Demo/Demo/Gravatar-Demo/DemoProfileViewControllerPresentingViewController.swift
etoledom: What about something like:
DemoProfilePresentationStylesViewController

We can add other presentation styles here later also.

Sources/GravatarUI/ProfileViewController/ProfileViewController.swift
```diff
@@ -0,0 +1,104 @@
+import Combine
+import Gravatar
+import UIKit
+
+@MainActor
```
etoledom: Is this needed?
I'd think that subclasses of a MainActor class are also MainActor 🤔

Sources/GravatarUI/ProfileViewController/ProfileViewController.swift
```diff
@@ -0,0 +1,104 @@
...
+    override public func viewDidLoad() {
+        super.viewDidLoad()
+        view.addSubview(scrollView)
+        scrollView.addSubview(profileView)
+        self.profileView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            scrollView.topAnchor.constraint(equalTo: view.topAnchor),
+            scrollView.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+            scrollView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            scrollView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
```
etoledom: Why do we want to use a scroll view?

And if that's fine, why not to set the scroll view as the root view? (via `loadView()`). It's simpler with the same final result 🤔

Sources/GravatarUI/ProfileViewController/ProfileViewModel.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+import Gravatar
+
+@MainActor
+public class ProfileViewModel {
```
etoledom, line 5: Nice!

etoledom (approved):
Looking good!

Do we want to make the bottom-sheet show only the profile view content hight?
Can be in a next PR.

We can have a flag `presentContentHeightOnBottomsheet` or similar.

Sources/GravatarUI/ProfileViewController/ProfileViewController.swift
pinarol: It's a copy paste leftover i think. Removing.

Demo/Demo/Gravatar-Demo/DemoProfileViewControllerPresentingViewController.swift
pinarol: Better!

Sources/GravatarUI/ProfileViewController/ProfileViewController.swift
pinarol: Because we or someone else may still want to limit the height of wherever this is displayed, to make sure it's doesn't get too bigger than intended. 
Maybe someone wouldn't want the bottom sheet get bigger than `.medium` detent... This can happen with a bigger font size in a small screen... In general, for a dynamic content like this, I think it's better to have a scroll view than not having it. 

> why not to set the scroll view as the root view?

I think it's more or less a personal taste and a habitual thing. Let me update.

etoledom: > I think it's more or less a personal taste and a habitual thing

I guess is personal taste for me also, in the other way 😅

etoledom: > In general, for a dynamic content like this, I think it's better to have a scroll view than not having it.

Oh I see! If the custom detent is smaller than the profile 🤔 
Ok as a safeguard

PR #229, opened 2024-04-24
GravatarUI docc bundle

### LINKED ISSUE #155: Restructure documentation for the Gravatar & GravatarUI split
We need to restructure the docs following the Gravatar & GravatarUI split.

etoledom:
Closes #155 

### Description

This PR implements the basics to add docs and pages to GravatarUI using Swfit-docc.

I've added a new snapshot test case to generate an example profile view to use in the docs. For now I've decided to use the same account that is used in https://docs.gravatar.com/general/profile/ , though this this is subject to change. More important than the account itself, is to check the strategy chosen to generate and use these snapshots.

I've created a make command to copy the snapshots from the test target to the docs bundle:
`make update-example-snapshots`

This is not hook to anything yet, so it need to be triggered manually.
Sadly in SPM we don't have a way to trigger post-test commands. 
We can add a post-action to the test target in the Xcode scheme, but it's a local setting.

### Testing Steps

- Checkout the banch
- Build the project
- On Xcode, go to Product -> Build Documentation
- Though the content is also subject to change, is enough to see that the main GravatarUI page renders correctly.

CleanShot 2024-04-24 at 18 33 32@2x


maxme:
I remember Beau asking somewhere to be removed from https://docs.gravatar.com/general/images/, might be worth changing the default screenshots here.

Sources/GravatarUI/GravatarUI.docc/GravatarUI.md
```diff
@@ -0,0 +1,44 @@
...
+
+```swift
+// 1. Get an instance of a ProfileService
+let service = ProfileService()
+// 2. Get the user's profile:
+let profile = try await service.fetch(with: .email("user@email.com"))
+// 3. Get the instance of a ProfileView:
+let profileView = ProfileView()
+// 4. Set the profile to the view:
+profileView.update(with: profile)
```
maxme, line 32: Kind of related to this documentation, as a newbie Swift dev, I'd like to know how you handle errors here :)

We're changing the network API on Android and had a discussion about errors on this PR: https://github.com/Automattic/Gravatar-SDK-Android/pull/130/files#r1579020468

etoledom:
> I remember Beau asking somewhere to be removed from https://docs.gravatar.com/general/images/, might be worth changing the default screenshots here.

Oh that's good to know.
We were talking with @pinarol about this also. We don't really want to keep this example account either. I just couldn't decide on which another image/information (even from a [random generator](https://randomuser.me/) 😅 )

But yes, the idea is to change it for an account of a not-real person.

Sources/GravatarUI/GravatarUI.docc/GravatarUI.md
etoledom: We will have some more-in-detail implementation examples in other pages. I plan a `Getting Started` page for example, and a tutorial for making lists of profile cards. Also probably another for the presentation styles.

This one is meant to be only introductory 👍 

But still from this example, if someone copy paste this code, error handling will be enforced by the compiler from the `try` clause. So, though it's not explicit, it won't be missed.
```
let profile = try await service.fetch(with: .email("user@email.com"))
```

pinarol: In Swift all errors are like checked exceptions so the compiler forces you to handle them. However, they are not declared in the method signature(you only add "throws"), so if you want to handle a specific error you need to do a type casting first. :( And you never know what sorts of errors will be thrown just by looking at the signature. So to compensate for that we try to keep a catalog of errors that we are throwing, that way 3rd party devs can have an idea. 

https://github.com/Automattic/Gravatar-SDK-iOS/blob/8dd5a4a4a48e7966514b627cc8bc428df5417828/Sources/Gravatar/Network/Errors.swift

maxme: 👍 thanks for the explanation!

pinarol:
I think we can generate one from [random generator](https://randomuser.me/) to unblock this PR. It would be good to have these changes in trunk since we might want to add more docs at the top of this. We can always update our snapshots in the future. WDYT? @etoledom

etoledom:
> I think we can generate one from [random generator](https://randomuser.me/) to unblock this PR. It would be good to have these changes in trunk since we might want to add more docs at the top of this. We can always update our snapshots in the future. WDYT? @etoledom

Cool! Let's do that

etoledom:
I've chosen an AI generated image and description, plus personal information from the simulator's contact app.

pinarol (approved):
Looks good 👍

pinarol:
> personal information from the simulator's contact app

Good choice. This is pretty familiar for any iOS dev.

PR #231, opened 2024-04-25
Removing `UIScreen.main` due to deprecation warning

### LINKED ISSUE #156: [Concurrency] Tackle the `UIScreen.main.scale` error
Screenshot 2024-04-03 at 12 01 05

The solution I come up for this one is to define a constant var and use it throughout the SDK. Example code here: https://github.com/Automattic/Gravatar-SDK-iOS/pull/136/files

```
import UIKit

@MainActor
enum UI {
    static let scaleFactor: CGFloat = UIScreen.main.scale
}
```

Although maybe there are better options.

etoledom:
Closes #156 

### Description

`UIScreen.main` has a deprecation warning:

```swift
@available(iOS, introduced: 2.0, deprecated: 100000, message: "Use a UIScreen instance found through context instead: i.e, view.window.windowScene.screen")
open class var main: UIScreen { get }
```

The proposed method to access it `view.window.windowScene.screen` won't work for us since we don't have direct access to a view on this context.

An alternative method to get the screen scale is using `UITraitCollection`. It will have the correct `displayScale` as long as we are running in a UIKit app. (This is the method KingFisher uses).

I've also added an alternative `ImageProcessingMethod` for a custom scale factor, in case the user requested a custom image size independent of screen scale.

With this change, we don't have the build error when building with strict concurrency.

### Testing Steps

- Gravatar demo app: Try fetching image demo screen.

### DIFF
--- a/Sources/Gravatar/Options/AvatarQueryOptions.swift
+++ b/Sources/Gravatar/Options/AvatarQueryOptions.swift
@@ -27,7 +27,7 @@ public struct AvatarQueryOptions {
         forceDefaultAvatar: Bool? = nil
     ) {
         self.init(
-            scaleFactor: UIScreen.main.scale,
+            scaleFactor: UITraitCollection.current.displayScale,
             rating: rating,
             forceDefaultAvatar: forceDefaultAvatar,
             defaultAvatarOption: defaultAvatarOption,
--- a/Sources/Gravatar/Options/ImageProcessorOption.swift
+++ b/Sources/Gravatar/Options/ImageProcessorOption.swift
@@ -11,20 +11,14 @@ public enum ImageProcessingMethod {
 
     /// A processing method which will directly transform the `Data` to an `UIImage` and return it.
     ///
-    /// This method uses UITraitCollection's displayScale to create a UIImage instance.
-    case common
-
-    /// A processing method which will directly transform the `Data` to a `UIImage` using the given scale factor.
-    /// - Parameter scaleFactor: The scale factor to use to create the `UIImage`.
-    case scaleFactor(_ scaleFactor: CGFloat)
+    /// - Parameter scaleFactor: The scale factor to use to create the `UIImage`. If nil,  UITraitCollection's displayScale is used.
+    case common(scaleFactor: CGFloat = UITraitCollection.current.displayScale)
 }
 
 extension ImageProcessingMethod {
     var processor: ImageProcessor {
         switch self {
-        case .common:
-            DefaultImageProcessor(scaleFactor: UITraitCollection.current.displayScale)
-        case .scaleFactor(let scaleFactor):
+        case .common(let scaleFactor):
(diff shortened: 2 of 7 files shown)


Sources/Gravatar/Options/AvatarQueryOptions.swift
```diff
@@ -27,7 +27,7 @@ public struct AvatarQueryOptions {
         forceDefaultAvatar: Bool? = nil
     ) {
         self.init(
-            scaleFactor: UIScreen.main.scale,
+            scaleFactor: UITraitCollection.current.displayScale,
```
pinarol, line 30: I can't be sure if this is the correct replacement. Are you sure about this? 

Screenshot 2024-04-26 at 13 52 06


Value `3.0` was mentioned in the docs of UIScreen.main.scale but not in UITraitCollection.current.displayScale. So I am really confused.

Screenshot 2024-04-26 at 14 01 57


Maybe we should just default to a constant value but expect the correct value from outside. 🤔 

I checked Kingfisher too, they use "UITraitCollection.current.displayScale" in a different place. Here they just default to 1 and expect the correct value from outside. Which is unfortunate I think but maybe it's the right thing to do.

etoledom: > Value 3.0 was mentioned in the docs of UIScreen.main.scale but not in UITraitCollection.current.displayScale. So I am really confused.

I checked that we get the correct value of 3.0. I also saw the docs and it mentions that this value specifies the screen density. As they mention "Retina display" my guess is that it's outdated documentation on their part.

etoledom: It's also mentioned [here](https://forums.developer.apple.com/forums/thread/733516), though on a different context.

pinarol: Okay thanks.

Sources/Gravatar/Options/ImageProcessorOption.swift
```diff
@@ -10,15 +11,21 @@ public enum ImageProcessingMethod {
...
+    case scaleFactor(_ scaleFactor: CGFloat)
 }
 
 extension ImageProcessingMethod {
     var processor: ImageProcessor {
         switch self {
         case .common:
-            DefaultImageProcessor.common
+            DefaultImageProcessor(scaleFactor: UITraitCollection.current.displayScale)
+        case .scaleFactor(let scaleFactor):
```
pinarol: I was thinking instead of adding a new case for this we can also change the "common" case as:

>     case common(scaleFactor: CGFloat = UITraitCollection.current.displayScale)

that way if you call it as `.common()` it'll have this default value.

wdyt?

etoledom: Yeah sounds good 👍

pinarol (approved):
Nice 🎉

PR #232, opened 2024-04-25
Refactor image cache

### LINKED ISSUE #157: [Concurrency] Fix `ImageCache.cache` issue
There's an error related to ImageCache.cache usage.

We can make this change to fix it.

Screenshot 2024-04-03 at 12 33 21

The whole diff is here: https://github.com/Automattic/Gravatar-SDK-iOS/pull/136/files

etoledom:
Closes #157 

### Description

This PR is an experiment implementing the cache system described [here.](https://developer.apple.com/tutorials/app-dev-training/caching-network-data).

There are two main changes:

- `ImageCache` is now an actor.
This serialises access to the internal NSCache instance.

- ImageCache will also cache the tasks of getting an image from remote.
This will avoid executing the same network call many times while waiting for a response.

Currently on the demo app `Image Download` screen, if you tap on `Fetch Avatar` many times, all those taps will turn into a network call, until any of the created network calls return.
With this change, there will be just one network call, and all other taps will await for this network call to return.

### Testing Steps

- Setup Proxyman or similar.
- Run the demo app and access `Image download - Networking` screen.
- Insert a valid Gravatar email account.
- Tap the fetch image button many times fast.
- Check on Proxyman that there's only one network call excecuted.


Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,92 @@
...
+
+    public func getImage(for key: URL) async throws -> UIImage? {
+        switch cache[key] {
+        case .ready(let image):
+            image
+        case .inProgress(let task):
+            try await task.value
+        case .none:
+            nil
+        }
```
pinarol: I think this operation of "awaiting the network task" should stay in the ImageDownloaderService.  Both because of the reasons I mentioned on the ImageCaching protocol but also throwing an error that belongs to an "external" task is causing a total ambiguity on the error handling side, how do you know how to handle this? Is the error from the external task, or did sth go wrong inside the cache? Should i move on with the network call? 

Also this would make it easier for 3rd party devs to implement their own cache based on our protocol.

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,92 @@
...
+    }
+}
+
+private enum CacheEntry: Sendable {
+    case inProgress(Task<UIImage, Error>)
+    case ready(UIImage)
```
pinarol: Not sure if it would make things more complex but this looks like a good fit for generic...

```
private enum CacheEntry<T: Sendable>: Sendable {
    case inProgress(Task<T, Error>)
    case ready(T)
}
```

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,92 @@
...
+/// Represents a cache for images
+///
+/// An ImageCaching will cache the task of obtaining an image from remote.
+/// Requesting an image from cache should await for any task cached and return the image from this task instead of creating a new task.
+public protocol ImageCaching: Sendable {
+    /// Saves an image in the cache.
+    /// - Parameters:
+    ///   - image: The image to set
+    ///   - key: The Image's URL.
+    func setImage(_ image: UIImage, for key: URL) async
```
pinarol: Can we keep the key as string? The cache key doesn't have to be a URL, no need to put this limitation.

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,92 @@
...
-        cache.setObject(image, forKey: key as NSString)
+    public func setImage(_ image: UIImage, for key: URL) {
+        cache[key] = .ready(image)
     }
 
-    public func getImage(forKey key: String) -> UIImage? {
-        cache.object(forKey: key as NSString)
+    public func setTask(_ task: Task<UIImage, Error>, for key: URL) async {
+        cache[key] = .inProgress(task)
+    }
```
pinarol: I think it would be more simpler if these methods start to take in/return `CacheEntry` from now on, then we wouldn't need this "setTask(...)" method. It's not clear how to implement this "setTask(...)" anyway and it doesn't do anything important.

I think `CacheEntry` is much more understandable than this. "I am passing you either the image or the "inProgress" task if you are interested to hold." 🤔

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,92 @@
...
+
+/// Making `setTask` optional.
+extension ImageCaching {
+    public func setTask(_ task: Task<UIImage, Error>, for key: URL) async {}
 }
 
-public class ImageCache: ImageCaching {
-    private let cache = NSCache<NSString, UIImage>()
+/// The default `ImageCaching` used by this SDK.
+public actor ImageCache: ImageCaching {
```
pinarol: NSCache is already thread safe. If I remember right we can just make the "shared" instance "let" instead of making this an actor, so this can just be another "Sendable" I think?

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,92 @@
...
+    ///   - task: The task to set.
+    ///   - key: The image's URL.
+    func setTask(_ task: Task<UIImage, Error>, for key: URL) async
+
+    /// Gets an image from cache for the given URL, or nil if none is found.
+    ///
+    /// If a task is found for the given URL, this  will await until the task returns the image.
+    /// - Parameter key: The image URL.
+    /// - Returns: The image cached for the given URL, or nil if none is found.
+    func getImage(for key: URL) async throws -> UIImage?
```
pinarol: Semantically, a cache get/set is known as a "synchronous" operation. 

I don't expect the cache to make me wait/await and then give me the value. 

It should give me the value "synchronously" if it has it "ready". Otherwise it shouldn't give me that value, and it's fine. But a cache shouldn't wait until that value gets "ready". Same for the set operation. I just pass the value to cache but I may not know when exactly that value will actually be saved to the actual datasource. 

It doesn't mean such a cache is not thread safe. 

It doesn't also mean it can't do any async operations inside.

There might be some async cache implementations where the real datasource is updated sometime after the cache entry is added, so the real datasource and cache entries can get out of sync sometimes. Which is fine and expected. But it's not a problem. We are prioritizing a different thing here. 

So I think these methods should not be async.

Sources/Gravatar/Network/Services/ImageDownloadService.swift
```diff
@@ -41,17 +47,18 @@ public struct ImageDownloadService: ImageDownloader {
         forceRefresh: Bool = false,
         processingMethod: ImageProcessingMethod = .common
     ) async throws -> ImageDownloadResult {
-        if !forceRefresh, let result = cachedImageResult(for: url) {
+        if !forceRefresh, let result = try await cachedImageResult(for: url) {
```
pinarol: There's no way of properly handling this cache error. It depends on the implementation. Should i go on with the network request or should i let it be thrown? 
This is another reason to keep the cache simple. I don't think we should be throwing from imageCache.getImage(...).

Sources/Gravatar/Cache/ImageCaching.swift
etoledom: Since this is tailored for ImageCache, to be used to cache images I don't see benefits on making it generic.
If we find that we can use this in the future for something else, we can do an easy refactor then.

Sources/Gravatar/Cache/ImageCaching.swift
etoledom: > Semantically, a cache get/set is known as a "synchronous" operation.

I guess this is true for internal, in-memory caches.

Users could use a database as cache, and you will need to await and throw. Or using an encrypted file. If a user implements gravatar in a Swift server (let's dream it's gonna be possible some day 🥲) they could even use Redis or some other external microservice.

This is an extract of Swift Vapor cache Protocol
```swift
    /// Sets an encodable value into the cache. Existing values are replaced. If `nil`, removes value.
    func set<T>(_ key: String, to value: T?) async throws where T: Encodable {
        try await self.set(key, to: value).get()
    }

    /// Gets a decodable value from the cache. Returns `nil` if not found.
    func get<T>(_ key: String) async throws -> T? where T: Decodable {
        try await self.get(key).get()
    }
```
And though this is a server use-case, it's not crazy to think that a user would want to use their database based cache system.

So I guess we can decide for practicality and simplify our implementation, or to decide to give users more posibilities.

Honestly at this point, I prefer to go for practicality and making it synchronous anyway. ¯\_(ツ)_/¯ 

The point of the ImageCaching being an `actor` is that NSCache is not Sendable, though it's mentioned that it is thread safe. So to make ImageCaching Sensable, ImageCache will need to be an actor. And the cache is a shared instance which will be passed across threads, so my guess is that it should be Sendable, and if it is, access to cache will have to await (though not needed to throw).

Not sure if this is the best way to go with this, but I'll try other options also.

Sources/Gravatar/Cache/ImageCaching.swift
etoledom: `Stored property 'cache' of 'Sendable'-conforming struct 'ImageCache' has non-sendable type 'NSCache<NSString, CacheEntryObject>'`

:( 

```
@available(*, unavailable)
extension NSCache : @unchecked Sendable {
}
```

NSCache is explicitly marked as non-sendable, event though the docs says it's thread safe. 

Seems like we need this to be an actor, or we will just be moving the problem somewhere else.

Sources/Gravatar/Cache/ImageCaching.swift
pinarol: > I guess this is true for internal, in-memory caches.

> Users could use a database as cache, and you will need to await and throw. Or using an encrypted file. If a user implements gravatar in a Swift server (let's dream it's gonna be possible some day 🥲) they could even use Redis or some other external microservice.

Nope I said that considering those cases also. As I mentioned a cache can do async jobs inside. It can write to a file or to a DB, and these jobs should be done asyncly, there's no doubt in that. But I would still expect that the Cache's get/set functions to be synchronous. 

Because the set() function should not wait for the DB operation to be completed to return. It can just put the operation to a queue to be done later. Or the get() function should not wait for DB writing to end to return. (Here we are waiting for a network call to end to return from a cache.get() which is really odd.) Cache's get() should return synchronously/immediately. 

In practice, these kinds of persistant caches also maintain an in-memory cache to be able to return immediately. That in-memory cache and the persisted cache(DB or a file system) can go out of sync sometimes. They get in sync later. Which is fine. And it doesn't concern us when designing our cache protocol.

pinarol: I tackled this compiler problem here before https://github.com/Automattic/Gravatar-SDK-iOS/pull/136/files#diff-5861f07091f878c705ab76ad62a0d214ad6f23a8a9f2dd28768d616bc9dd2ca0

It was enough to just turn "ImageCache" into a struct, and "shared" instance to a let. Not sure why we need to do sth else.

etoledom: > Not sure why we need to do sth else.

Because NSCache is not Sendable.

Those changes remove the warnings from this file, but those warnings are just moved somewhere else.

pinarol: OK I think it's because you made ImageCaching protocol Sendable. Not sure if it needs to be Sendable. The compiler doesn't seem to complain when we revert that.

Sources/Gravatar/Cache/ImageCaching.swift
pinarol: Do we need to make ImageCaching Sendable?

etoledom:
Ok, I managed to silence all warnings with the minimum of changes to the `ImageCaching` struct.

Overall, I like this approach of caching the image request task to avoid multiple ongoing network calls, but it was more troublesome than just fixing the original concurrency warnings.

It's ready for another look.

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,75 @@
...
+    /// - Parameters:
+    ///   - image: The cache entry to set.
+    ///   - key: The entry's key, used to be found via `.getEntry(key:)`.
+    func setEntry(_ entry: CacheEntry, for key: String)
+
+    /// Gets a `CacheEntry` from cache for the given key, or nil if none is found.
+    /// .
+    /// - Parameter key: The key for the entry to get.
+    /// - Returns: The cache entry which could contain an image, or a task to retreive the image. Nill is returned if nothing is found.
+    func getEntry(with key: String) -> CacheEntry?
```
pinarol, line 21: > or a task to retreive the image.

This a bit sounds like a call to action, it's like as a 3rd party dev I need to use that task to retrieve the image my own. I suggest saying "the task that is retrieving the image" instead. Small change but less indication of an action imho.

We can also add a little more insight for other devs to understand, "`.inProgress(task)` state is used by the image downloader to check if there's already an ongoing download task for the same image. If yes, it awaits that task instead of starting a new one.". 

Also, typos at "retreive" and "Nill".

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,75 @@
-import Foundation
 import UIKit
 
+/// Represents a cache for images
+///
+/// An ImageCaching will cache an instance of an image, or the task of retriving an image from remote.
+/// Requesting an image from cache should await for any task cached and return the image from this task instead of creating a new task.
```
pinarol: This sentence seems outdated now.

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,75 @@
-import Foundation
 import UIKit
 
+/// Represents a cache for images
+///
+/// An ImageCaching will cache an instance of an image, or the task of retriving an image from remote.
```
pinarol: Suggestion:

> An ImageCaching represents a cache for ``CacheEntry`` elements. Each ``CacheEntry`` is either an  instance of an image, or the task of retrieving an image from remote.

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -1,24 +1,75 @@
...
+    /// Gets a `CacheEntry` from cache for the given key, or nil if none is found.
+    /// .
+    /// - Parameter key: The key for the entry to get.
+    /// - Returns: The cache entry which could contain an image, or a task to retreive the image. Nill is returned if nothing is found.
+    func getEntry(with key: String) -> CacheEntry?
+}
+
+/// Making `setTask` optional.
+extension ImageCaching {
+    public func setTask(_ task: Task<UIImage, Error>, for key: URL) async {}
```
pinarol: `setTask(...)` can be removed now. Plus the related code in the unit tests.

pinarol:
Looking good just some finishing touches! 🎉

etoledom:
> Looking good just some finishing touches! 🎉

@pinarol - Updated with the latest review comments

pinarol (approved):
Looks good 🎉 :shipit:

PR #233, opened 2024-04-26
Add RemoteSVGView to display an SVG via URL

### LINKED ISSUE #198: Download unknown account icons from URL

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/198

### Description

Adding RemoteSVGView to display an SVG via URL. This is just a fallback mechanism in case we don't have the icon locally. We should still add new icons to local when they come up. 

Using a WKWebView inside and embedding the downloaded `<SVG>` to an html where we can set its size & color. 

### Testing Steps

1. I added a new screen to the demo app > "Display Remote SVG"

2. You can temporarily change the code to use RemoteSVGView. That way you can use profile cards in the demo app to test it.

Screenshot 2024-04-26 at 18 16 24


Sources/GravatarUI/ProfileFields/AccountButtonBuilder.swift
```diff
@@ -3,6 +3,7 @@ import UIKit
 
 @MainActor
 public struct AccountButtonBuilder {
+    static let fallbackIcon: UIImage? = UIImage(systemName: "link")
```
pinarol: Now that I fix this the link icon it become visible in the snapshot tests. So I regenerated the snapshots. However I think that icon looks a bit large for this purpose. I opened a thread in slack to discuss this further.

pinarol: Okay, we are moving fwd with the link icon from the WP design library.

Sources/GravatarUI/ProfileView/AccountIconWebView/RemoteSVGView.swift
```diff
@@ -0,0 +1,153 @@
+import Gravatar
+import UIKit
+import WebKit
+
+class RemoteSVGView: UIView, WKNavigationDelegate, UIGestureRecognizerDelegate {
```
etoledom: I'd suggest making this class a subclass of `UIControl`. 
In this way we get taps and all the different control action events handling (`touchUpInside` or any other) by default, plus all the control states (highlighted, selected, etc...) and we don't need to worry about `tapHandler` closure or gesture recognizers.

Users could use it as a normal `UIButton` adding a UIAction or a targer:action:

```swift
svgButton.addAction(UIAction { _ in print("Icon Tapped!") }, for: .touchUpInside)
```

We can also do something fancy like:
```swift
override var isHighlighted: Bool {
    didSet {
        webView.alpha = isHighlighted ? 0.8 : 1
    }
}
```
to have a very simmilar response to taps as the other account buttons.

If we go this way, the only other change needed is:
`.isUserInteractionEnabled = false` on `fallbackImageView` and `webView` to let the event pass through.
And maybe we could call this class `RemoteSVGButton`.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -298,6 +301,35 @@ open class BaseProfileView: UIView, UIContentView {
...
+            guard let self else { return }
+            self.delegate?.profileView(self, didTapOnAccountButtonWithModel: model)
+        }
+        if UIImage(named: model.shortname) != nil {
+            return createAccountButton(model: model, tapHandler: tapHandler)
+        } else if let iconURL = model.iconURL { // If we have the iconURL try downloading the icon
+            return createAccountWebView(url: iconURL, tapHandler: tapHandler)
+        } else { // This will show the local fallback icon
+            return createAccountButton(model: model, tapHandler: tapHandler)
+        }
```
etoledom: Having `RemoteSVGView` being a UIControl subclass, we can share and pass around a `UIAction` instance instead of a closure.

etoledom (commented):
Looking pretty good!

I added a suggestion on code about `UIControl`.

Apart from that, it would be great if we could add a small fade-in to the icon when it loads.
We could add this small css to the webview
```css
@keyframes fadeIn {
    0% { opacity: 0; }
    100% { opacity: 1; }
}
.fade-in { animation: fadeIn 0.2s; }
```

and: `<div class="icon fade-in">` on the HTML side.

Just an option but the effects improve quite a bit in my opinion.

Sources/GravatarUI/ProfileView/AccountIconWebView/RemoteSVGView.swift
pinarol: Great suggestion! Done.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
pinarol: > Apart from that, it would be great if we could add a small fade-in to the icon when it loads.

Yes it looks better this way. Thanks a lot for the css snippet.

pinarol:
Inheriting from UIControl makes perfect sense here. Thanks a lot. Ready for another look.

Sources/GravatarUI/ProfileView/AccountIconWebView/RemoteSVGView.swift
```diff
@@ -0,0 +1,153 @@
...
+            webView.widthAnchor.constraint(equalTo: widthAnchor),
+            webView.heightAnchor.constraint(equalTo: heightAnchor),
+            webView.centerXAnchor.constraint(equalTo: centerXAnchor),
+            webView.centerYAnchor.constraint(equalTo: centerYAnchor),
+            imageView.widthAnchor.constraint(equalTo: widthAnchor),
+            imageView.heightAnchor.constraint(equalTo: heightAnchor),
+            imageView.centerXAnchor.constraint(equalTo: centerXAnchor),
+            imageView.centerYAnchor.constraint(equalTo: centerYAnchor),
+        ])
+        return imageView
```
etoledom: One last thing:
Any special reason to do the layout inside the lazy var init block? 🤔

etoledom (approved):
Looking good! Thank you for the update.

Just one more question on code review, but not blocking so ✅

Sources/GravatarUI/ProfileView/AccountIconWebView/RemoteSVGView.swift
pinarol: No good reason. This is the result of trying too many different things, and forgetting to revert some of them. Let me move these and the addSubview() calls to the init(). Thanks!

pinarol: Fixed.

PR #237, opened 2024-05-01
Adding profile view empty state

### LINKED ISSUE #234: Empty state to use for the missing Gravatar account
We need to add the empty state to use when no Gravatar profile exists.

Figma:

https://www.figma.com/file/AzU51EzjGYFJPtLKG2GBIr/SDK-Components?type=design&node-id=9308%3A3707&mode=design&t=7IyupvAU7qFGeK1C-1

etoledom:
Closes #234 

### Description

This PR adds an empty state to the profile cards.

### Testing Steps

- Smoke-test the demo app, both cards and configuration demos.
- By entering an email without an account, the empty state should show.


Sources/GravatarUI/DesignSystem/Palette.swift
```diff
@@ -25,12 +25,21 @@ public struct Palette {
     public let background: BackgroundColors
     public let avatarBorder: UIColor
     public let placeholder: PlaceholderColors
-    public init(name: String, foreground: ForegroundColors, background: BackgroundColors, avatarBorder: UIColor, placeholder: PlaceholderColors) {
+    public let preferredUserInterfaceStyle: UIUserInterfaceStyle
```
pinarol, line 29: We should document and explain what this is used for i think... This can be confusing, like, we already have PaletteTypes and light, dark options, and there's also this... It feels like duplication...

Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -326,12 +327,19 @@ open class BaseProfileView: UIView, UIContentView {
         return button
     }
 
+    public func updateAsEmpty(userName: String? = nil) {
+        guard let config = (configuration as? ProfileViewConfiguration) else { return }
+        configuration = config.emptyProfile(userName: userName)
+    }
+
     open func update(with config: ProfileViewConfiguration) {
+        clearFields()
```
pinarol: I feel like this `clearFields()` belongs to `updateAsEmpty()` instead. 

Sometimes it doesn't look good to see the placeholders between each avatar load. So we have an option like`ImageSettingOption.removeCurrentImageWhileLoading` for our UIImageView extension. While we offer that option, we shouldn't override that behavior here(see: `ProfileViewConfiguration.avatarSettingOptions`). It should be a more deliberate choice to clear the current image rather than doing it every time we update the config. 

What we can do is I think, if avatarID is nil then we can clear the image(if there's a placeholder in the config we can use it otherwise set the image to `nil`.). Otherwise we shouldn't touch the image. This way one can be more in control about this behavior. 

For the case where there's already an avatarID, our imageView extension will set the placeholder considering the empty state and the option `.removeCurrentImageWhileLoading`.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -326,12 +327,19 @@ open class BaseProfileView: UIView, UIContentView {
...
         paletteType = config.palette
         padding = config.padding
         isLoading = config.isLoading
         avatarActivityIndicatorType = config.avatarActivityIndicatorType
-        if let avatarID = config.avatarID {
+        avatarImageView.image = config.avatarPlaceholder
```
pinarol: As I mentioned  above I think we shouldn't override the current image here. The placeholder should be set only when avatarID is nil.

Sources/GravatarUI/ProfileFields/Model/EmptyProfileModel.swift
```diff
@@ -0,0 +1,40 @@
...
+    )
+
+    var displayName: String? = NSLocalizedString(
+        "GravatarEmptyProfileDisplayNameKey",
+        value: "Your Name",
+        comment: ""
+    )
+
+    let accountsList: [any AccountModel] = [GravatarAccountModel(accountURL: URL(string: "https://gravatar.com/profile"))]
+    let avatarIdentifier: Gravatar.AvatarIdentifier = .hashID("")
```
pinarol: I think we should make `AvatarIdentifierProvider.avatarIdentifier` optional and set to `nil` here. It's to be able to show the placeholder(or clear the image) in a deliberate way. Otherwise we have a problem about it. (I put more comment below about this). AvatarIdentifierProvider is a protocol we invented to use for Profile views, so it should be fine to do this.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -326,12 +327,19 @@ open class BaseProfileView: UIView, UIContentView {
         return button
     }
 
+    public func updateAsEmpty(userName: String? = nil) {
```
pinarol: `updateAsEmpty()`, this function doesn't do what the name promises. The name sounds like we are setting the fields as empty. But it doesn't. I think we should be more descriptive about it, something like: "showProfileClaim()". And we should document it preferably with a screenshot.

Sources/GravatarUI/ProfileFields/AccountButtonBuilder.swift
```diff
@@ -33,8 +33,8 @@ public struct AccountButtonBuilder {
 }
 
 extension UIImage {
-    convenience init?(localName: String) {
-        self.init(named: localName, in: Bundle.module, with: nil)
+    convenience init?(named name: String) {
```
pinarol, line 36: Nit: Not a big deal at all but with the old naming it was more clear what it's doing.

Sources/GravatarUI/ProfileView/ProfileViewConfiguration.swift
```diff
@@ -81,4 +81,14 @@ extension ProfileViewConfiguration {
     public static func largeSummary(model: ProfileSummaryModel? = nil, palette: PaletteType = .system) -> ProfileViewConfiguration {
         self.init(model: model, palette: palette, profileStyle: .largeSummary)
     }
+
+    public func emptyProfile(userName: String? = nil) -> ProfileViewConfiguration {
```
pinarol: Again, I think we should update the name `emptyProfile`. This is not an empty profile. We can maybe use `profileClaim(...)`.

pinarol (commented):
Just dropping some small comments

1. I wasn't able to see the dark mode of the empty states in the demo app. They look light even if I select dark.



2. In [figma](https://www.figma.com/file/AzU51EzjGYFJPtLKG2GBIr/SDK-Components?type=design&node-id=9308-3707&mode=design&t=75AjbgAxl9aXCgIo-0) there are some dashed lines around the "Tell the world who you are..." section. So unfortunately it's not sufficient to make this change only by passing new texts. Maybe we can put this border when `ProfileButtonStyle == .create`. 🤔 

Screenshot 2024-05-02 at 12 42 06

Sources/GravatarUI/ProfileFields/AccountButtonBuilder.swift
etoledom: Yes, but we forgot to use this init for the account buttons with local images, and them all were loading from remote SGV, (except from the G with loaded the icon as default), so I decided to use the same name we normally use to avoid this error again in the future.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
etoledom: Makes sense. I'd still prefer a more generic name like `updateAsNotFound`.
Now we decided to show the `claim profile` button but that could change in the future.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
etoledom: I saw errors with the profile image showing where it shouldn't in the table view.
Will double-check 👀

Sources/GravatarUI/DesignSystem/Palette.swift
etoledom: Yes, definitely will need an explanation. I didn't find another way to load the correct version of the image.

There was one other option though: To make the profile image a symbol and we can tint it programatically. 
It has two colors which we can tinti separatelly, if this is a symbol. It would be great for customisation, but I'm not sure how much extra work will it entail.

There's one example in figma of diferent tints:
CleanShot 2024-05-02 at 13 45 05@2x

etoledom:
Thanks for the review, I'll check out the comments.

About this one:

> 1. I wasn't able to see the dark mode of the empty states in the demo app. They look light even if I select dark.

That's odd, it works well for me:



CleanShot 2024-05-02 at 14 00 41

Sources/GravatarUI/ProfileFields/AccountButtonBuilder.swift
etoledom: https://github.com/Automattic/Gravatar-SDK-iOS/pull/239

That's is what I meant 😅 
It went pass both of us in the SVG review.

It does say that the SVG WebView as button looks convincible good 😃

etoledom: `init?(named:)` 
does the same than 
`init?(named:in:compatibleWith)`
They both search for the image locally, but are specifying the `module` bundle instead of the `main` bundle as an internal convenience.
That's why I think this name is better, and it will avoid future issues like this one with the svg buttons.

pinarol: I see. Let's try then.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
pinarol: But this is, by nature, not just a generic "not found", it is intended for the "current user" to claim their own profile. We put pre-defined texts and everything is made specific for this use case. 

And the more I look at this method the more think like it shouldn't exist at all. It doesn't do anything meaningful inside. It's just applies the right config. 🤔

Sources/GravatarUI/DesignSystem/Palette.swift
pinarol: > tint it programatically

I am not sure if this is worth it.

pinarol: > There's one example in figma of diferent tints:

😬

Sources/GravatarUI/ProfileView/BaseProfileView.swift
etoledom: This is basically there because I worked on the configuration later.

> It's just applies the right config.

I kept it as a convenience for the user to call when they create and configure the view directly without the configuration.

Otherwise this is needed:
```swift
if let config = largeProfileView.configuration as? ProfileViewConfiguration {
    largeProfileView.configuration = config.emptyProfile()
}
```

Creating a new configuration is complicated because the user will have to know and set the correct view style and model, and that's a source of possible errors.

Other possible solutions could be:

- We can make a `.profileConfiguration` field to retrieve the typed configuration. 
  - UIButton does this
- Every Profile View subclass can offer a pre-configured initial configuration appropriate for their style.
  - UITableViewCell does something like this with `.defaultContentConfiguration`.

pinarol: One might be using one of the profile views without passing any configuration and it should be fine. We offer public methods for this to happen. Using the UIContentConfiguration is optional. So for this to happen I think our public methods shouldn't assume that a valid configuration exists. 

This one assumes that a valid config exists, and derives the new config based on it. So there's no guarantee this will work fine. I might have made changes in the UI directly without using the configuration, after this call it won't just change things related to the empty profile, it will also bring other fields from the existing configuration(because it creates a copy) which i didn't intend to use. 

To be consistent I would update this method to offer the same functionality without using the configuration in case we want to keep it. (in this case we need a different one for each subview because we need to call update(with: model))

```
        self.isLoading = false
        update(with: EmptyProfileModel(userName: userName))
        self.profileButtonStyle = .create
        self.avatarPlaceholder = UIImage(named: "empty-profile-avatar")
```

When we change other things like "isLoading, model, profileButtonStyle etc." we don't do it via updating the configuration first and applying the new configuration. We directly update the UI element. So I think this method creates a sort of inconsistency about how our public methods work.

The configuration should just be a sort of proxy in front of what our public API is capable of. If we offer functionality that may not work properly unless ones uses the profile views via a configuration we can lead to bugs. 

--- 
I think the key here is not copying the existing config. If we create the configuration from scratch we should be fine. So a static method that creates a new config at each "subview" can also work fine, for example:

```
extension LargeProfileView {

     public static func claimProfileConfiguration(userName: ...) {
           let config = // create new config suitable for LargeProfileView
            config.model = EmptyProfileModel(userName: userName)
            config.profileButtonStyle = .create
            config.avatarPlaceholder = UIImage(named: "empty-profile-avatar")
            return config
     }
}

extension ProfileView {

     public static func claimProfileConfiguration(userName: ...) {
           let config = // create new config suitable for ProfileView
            config.model = EmptyProfileModel(userName: userName)
            config.profileButtonStyle = .create
            config.avatarPlaceholder = UIImage(named: "empty-profile-avatar")
            return config
     }
}
```

(We can create an internal convenience method for the repeating lines I think.)

(In this case we may not need to offer an additional separate public method I think since these configs should be designed in a way that works fine even if I didn't use any configuration on this view before.) 

(We can as well define these near the other configs, I put them here just as an example.)

etoledom:
Updated with review comments.

> In [figma](https://www.figma.com/file/AzU51EzjGYFJPtLKG2GBIr/SDK-Components?type=design&node-id=9308-3707&mode=design&t=75AjbgAxl9aXCgIo-0) there are some dashed lines around the "Tell the world who you are..." section.

I totally missed this one. 

I spent some time tinkering about this one. The dash lines plus the padding makes me think that a good path to go is either UIView wrapper of the label or using a UITextView.

As our public API is a UILabel, and the builders work based on UILabel, ideally we keep `UILabel` as the public interface, and we can keep using it with the `AboutMeBuilder`, so I tried to keep it as is as much as possible, but this forced the usage of a UILabel subclass.

Some extra thoughts:

With the UIView wrapper, it's simple to add the dashes and also the padding with constraints. Drawbacks: 
- More difficult to reach UILabel properties.
- Probably `DashedView` would need to be public.
- Confusion to create BaseProfileView subclasses (aboutMeLabel already has a superview).
  - Will we change the interface to a  `aboutMeView`? 
  - Should we keep a `aboutMeLabel` for easy access? 
    - This would make it confusing to subclass `BaseProfileView`.
  - Modify the Builder and make the UILabel more difficult to reach?
  - Make accessors in the UIView to access internal UILabel properties? 😬 
  - ... etc.  


Then I thought about a protocol composition, but some overrides are needed anyway, so though it would help to share this code and make a DashedView, DashedButton, etc... as for now we don't need them, and the refactor is simple, and it won't help overcoming the drawbacks of the UILabel subclass.

Finally, the UILabel subclass.
I like that we can keep the public api as UILabel, the `AboutMeBuilder` keeps working without modifications and the `BaseProfileView` subclasses don't need modifications either.
Drawbacks:
- override of:
  - `layoutSubviews`. Here we don't do any layout modification, so there are no circular calls to the layout engine because of this override.
  - `drawText`. Needed to make the padding in the UILabel.
  - `preferredMaxLayoutWidth` To help the layout system know when is best to jump to new lines (there's less space available when we have the padding.

I try to avoid those overrides, but in this case it feels better than the UIView wrapper option.

Still, I'm happy to reconsider.

pinarol (approved):
Looks good ! 🎉 🥳

pinarol:
> I like that we can keep the public api as UILabel, the AboutMeBuilder keeps working without modifications and the BaseProfileView subclasses don't need modifications either.

Exactly. Even though the internal implementation gets a bit more complex, subclassing makes sense here. I thought similar things.

Sources/GravatarUI/DesignSystem/UIColor+DesignSystem.swift
```diff
@@ -34,6 +34,7 @@ extension UIColor {
     static let boatAnchorGray: UIColor = .rgba(107, 107, 107)
     static let spanishGray: UIColor = .rgba(151, 151, 151)
     static let dugongGray: UIColor = .rgba(112, 112, 112)
+    static let orchidBlack: UIColor = .rgba(80, 87, 94)
```
pinarol: nit: I am not sure if this is a good name. The color looks more like a gray. And orchid color is more similar to purple. 

If you ask me, the color naming site got it right again:

Screenshot 2024-05-06 at 14 49 33

`bovineGray`

pinarol: OR `powderedSlateGray`

etoledom: Sure, I'm a zero about colors 😆 
Though I took the name from that site ¯\_(ツ)_/¯ 
It was one of the options.

etoledom:
Thank you for the review 🙏

PR #239, opened 2024-05-02
Use `UIImage(localName:)` to check if the account icon exists.

pinarol:
Closes #

### Description

I noticed we have an unintended bug in account icons. It fallbacks to the webview unnecessarily. This should fix it.

### Testing Steps

In any demo screen with a profile card, the account icon should not be shown in the webview.


pinarol:
Merged in https://github.com/Automattic/Gravatar-SDK-iOS/pull/241

PR #241, opened 2024-05-02
Fix: UILabel placeholder animations doesn't work in ProfileViewController

### LINKED ISSUE #240: Bug: UILabel placeholder animations doesn't work in ProfileViewController
The placeholder animations for UILabel elements doesn't seem working:

Whereas it works well in this screen:

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/240

### Description

This happens when animations begin and then for some valid reason we re-display the placeholders(which basically means we update the colors and etc.). 

UILabel is different in the sense that its `backgroundColor` is not animatable. So for UILabel we shouldn't let backgroundColor to be updated while an animation is in progress. 

### Testing Steps

Preferably use a network conditioner and lower your network speed

Open the Demo app > Profile View Controller > Enter an invalid email and show the bottom sheet

Observe: UILabels animate.


etoledom (approved):
Looking good!

Just to note that this will be merged to [wppinar/fix-account-icon](https://github.com/Automattic/Gravatar-SDK-iOS/tree/wppinar/fix-account-icon).

Should we merge to trunk?

PR #244, opened 2024-05-06
Set a background color for avatar

### LINKED ISSUE #243: Add a background color for the avatar
I feel like this is missing. You don't always want to set a placeholder, but sometimes just want a background color. I think we should allow doing that. We can have such a color entry in the Palette.

pinarol:
Closes #243 

### Description

Avatar looks bad without a background color until the image loads. So I think it would be good to have an option to set a background color for it. 

- Adds a new color to the palette for the avatar background. (I used the placeholder color by default.)
- Adds "AvatarColors" to wrap avatar related colors. Not sure if it's worth it but it seemed tidy this way. 

### Testing Steps

Smoke test with the demo profile cards. You should see a background color until the avatar loads.


etoledom (approved):
Nice addition 🙏

PR #247, opened 2024-05-07
Open `claimProfile(...)` from `ProfileViewConfiguration` as well

pinarol:
Closes # Continuation of the empty profile work.

### Description

I noticed that we didn't offer the `claimProfile(...)` config directly via `ProfileViewConfiguration`. In pocket-casts I use `ProfileViewConfiguration` so the code is unaware of the underlying view type. It would be good to offer this via `ProfileViewConfiguration` as well.

I am not sure if this is the best but I created a new `Templates` enum to separate these type of "prefilled" configurations from other static functions that create an empty profile. 

So the usage will look like this:

> ProfileViewConfiguration.Templates.claimProfile(profileStyle: .summary, ...)

Let me know what you think.

### Testing Steps

CI green.

### DIFF
--- a/Sources/GravatarUI/ProfileFields/Model/ClaimProfileModel.swift
+++ b/Sources/GravatarUI/ProfileFields/Model/ClaimProfileModel.swift
@@ -53,24 +53,24 @@ extension ProfileViewClaimProfileConfigurable where Self: BaseProfileView {
 
 extension LargeProfileView: ProfileViewClaimProfileConfigurable {
     public static func claimProfileConfiguration(userName: String? = nil, palette: PaletteType = .system) -> ProfileViewConfiguration {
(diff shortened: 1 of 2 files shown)


etoledom (approved):
Looks good! 

I don't think the `Template` is needed though.

There are examples with `UIButton` and `UIContentUnavailableView` where their configurations offer similar pre-filled configurations for common use cases:

```swift
UIButton.Configuration.bordered()
UIButton.Configuration.filled()
UIButton.Configuration.plain()

UIContentUnavailableConfiguration.empty()
UIContentUnavailableConfiguration.loading()
UIContentUnavailableConfiguration.search()
```

So I'd say it's a common pattern. They can also be used as a starting point, to be modified with specific customisation.

etoledom (approved):

PR #248, opened 2024-05-08
[wip] Open API (POC)

etoledom:
# WIP

### Description

This is a proof of concept of an implementation of Open API.

The goals of this implementation are:
- To minimise external dependencies.
- To avoid disrupting the user of this SDK with build-time code generation.
- To balance automation and control.

Apple provides their own [open-api-generator](https://github.com/apple/swift-openapi-generator). They advice on using the plugin to automatically generate both client and models on build-time, making it not needed to commit the generated code in source control.

I believe that while this could work good for an executable/application project, is not the ideal for a library.

As an alternative, I've decided to use the generator as a stand-alone executable via `makefile`. This avoids adding the plugin as a dependency, and we can control when we want to run the code-generator.

`make generate` will:
- Checkout and build `https://github.com/apple/swift-openapi-generator`. 
- Generate the model and client as defined in `openapi-generator-config.yaml` with the model from `openapi.yaml`.
- Create/update `Sources/Gravatar/OpenApi/Generated/Client.swift` and `Sources/Gravatar/OpenApi/Generated/Types.swift`.

### Dependencies:

Sadly, `.package(url: "https://github.com/apple/swift-openapi-runtime", from: "1.4.0")` seems to be necessary.

My ideal for the generator was for it to create a simple `Profile` model without any extra dependency, and we continue using our `ProfileService`. But the `Types` generated file comes with `import OpenAPIRuntime`, so we need to add this dependency.

Funny enough, types from `OpenAPIRuntime` are not used in the models generated, but in the `internal enum Operations` created to be used with the client generated code. This is true even if we ask to generate only the Types (and not the client).

I thought we can just delete this part of the generated code, it would be fast and easy, but against the whole idea of using OpenApi.

Currently I don't see any great benefit on also generating the client code. It actually makes it troublesome to hook it up to our services structure. But again, OpenApi's idea is to create both model and client ready to use automatically, and it could be beneficial in the future if we add new endpoints to the API (let's say galleries or who knows).

I've also avoided adding `.package(url: "https://github.com/apple/swift-openapi-urlsession", from: "1.0.0"),` dependency, and using our own `HTTPClient` as OpenApi's `ClientTransport`. Ideally we manage to keep the ability for users to use their own client, but we will need some extra work to offer this in a proper way.

## TLDR

- This works and we can use it.
- I'd love to only generate the model and forget about the client, but if we do that, we won't get all the goodies from OpenApi.
- Generating the client could make things easier in the future, but the implementation of this client with our current services structure can complicate things.
- We can update the code by:
  - Updating `openapi.yaml` with the latest from the server.
  - Running `make generate`.
  - Done.
- The generated code follows casing convention from the yaml file, which means in our case that the model will be camel-cased. We necessarily need to wrap it in our own public UserProfile struct.


Sources/Gravatar/Network/Services/ProfileService.swift
```diff
@@ -37,69 +36,46 @@ public struct ProfileService: ProfileFetching {
...
+                let profile = try response.body.json
+                return UserProfile(profile: profile)
+            case .notFound(_):
+                throw ProfileServiceError.responseError(reason: .invalidHTTPStatusCode(code: 404))
+            case .tooManyRequests(_):
+                throw ProfileServiceError.responseError(reason: .invalidHTTPStatusCode(code: 429))
+            case .internalServerError(_):
+                throw ProfileServiceError.responseError(reason: .invalidHTTPStatusCode(code: 500))
+            case .undocumented(statusCode: let statusCode, _):
+                throw ProfileServiceError.responseError(reason: .invalidHTTPStatusCode(code: statusCode))
```
pinarol: I am under the impression that these cases will never run because we inject our own URLSessionHTTPClient as the "transport" object which throws an error(`HTTPClientError.invalidHTTPStatusCodeError(...)`) when these status codes are encountered. (I couldn't test it though.)

So this code will never run. For us, if this code runs then it means something unexpected happened. So we can just handle them in a "default" case and throw "unexpected" here. Sth like: 

> throw ProfileServiceError.responseError(reason: .unexpected(UnexpectedInvalidHTTPStatusCodeError))
> // I made up "UnexpectedInvalidHTTPStatusCodeError" but we can define a new kind of Error similar to that i think.

This would help us reduce the boilerplate code as well.

Sources/Gravatar/Network/Services/HTTPClient.swift
```diff
@@ -17,3 +19,60 @@ public protocol HTTPClient: Sendable {
...
+            guard 
+                let rawName = headerName as? String,
+                let name = HTTPField.Name(rawName),
+                let value = headerValue as? String
+            else { continue }
+            headerFields[name] = value
+        }
+        self.init(status: .init(code: httpResponse.statusCode), headerFields: headerFields)
+    }
+}
```
pinarol: Can we give a try to https://github.com/apple/swift-openapi-urlsession and see if it will help us get rid of this boilerplate here?

etoledom:
Replaced by https://github.com/Automattic/Gravatar-SDK-iOS/pull/249

PR #249, opened 2024-05-15
using-openapi-generator-swift5

etoledom:
This is an alternative to https://github.com/Automattic/Gravatar-SDK-iOS/pull/248

On this PR we use https://openapi-generator.tech/docs/generators/swift5/ instead of https://github.com/apple/swift-openapi-generator .

Benefits over Apple's generator:
- We can choose to generate and use a model without external dependencies.
- We can use our own Service infrastructure to make the request, without the need of the generated client.
- All model seem to be generated properly, and with camelCased fields.
  - The Apple's generator seems to have issues with a couple of types like [here](https://github.com/Automattic/Gravatar-SDK-iOS/blob/d0f7a521545d51f6683c14f42f3603add670156a/Sources/Gravatar/OpenApi/Generated/Types.swift#L230). 

Drawbacks over Apple's generator:
- This generator is a Java cli.
- To avoid all Java dependencies, I opted for a Docker based option.
  - Docker needs to be installed and running to generate the code.
- The generated structures used by profile (like `ProfileContactInfo`) are not nested to the `Profile` struct. This is not terrible but it would be a nice-to-have.
- Generating the code with `make install-and-generate` (when `openapi-generator` have not been yet cloned) is really slow.

On this example, I opted for having `Profile` public, I'm not very happy with this, but the extra boilerplate of wrapping it in a different struct seems unnecessary now, keeping in mind that we can have control (or at least a vote) on yaml spec changes (up to debate).


Sources/Gravatar/Network/Services/ProfileService.swift
```diff
@@ -36,6 +37,20 @@ public struct ProfileService: ProfileFetching {
         }
     }
 
+    public func v3fetchProfile(with profileID: ProfileIdentifier) async throws -> Profile {
```
etoledom: These `v3` prefixes are temporary to avoid changing too many things on this preview of PR.
This will replace `fetchProfile(:)` before merging.

Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
```diff
@@ -100,8 +100,10 @@ class DemoFetchProfileViewController: UIViewController {
     func fetchProfile(with profileID: ProfileIdentifier) async {
         let service = ProfileService()
         do {
-            let profile = try await service.fetch(with: profileID)
-            setProfile(with: profile)
+//            let profile = try await service.fetch(with: profileID)
+            let profileV3 = try await service.v3fetchProfile(with: profileID)
+            print("V3: \(profileV3)")
+//            setProfile(with: profile)
```
etoledom: This could fail as for now, since the backend returns an object instead of an array for `verifiedAccount`, and pronouns can return null, but the model is not optional.

Sources/Gravatar/OpenApi/Generated/URLSessionImplementations.swift
```diff
@@ -0,0 +1,648 @@
...
+#endif
+#if canImport(UniformTypeIdentifiers)
+import UniformTypeIdentifiers
+#endif
+
+//public protocol URLSessionProtocol {
+//    func dataTask(with request: URLRequest, completionHandler: @escaping @Sendable (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask
+//}
+//
+//extension URLSession: URLSessionProtocol {}
```
etoledom: This gets re-defined, so I had to comment it out (or remove it) to make this compile.
Eventually I think we can modify the template to avoid having this issue in the future.

Sources/Gravatar/Network/Services/ProfileService.swift
```diff
@@ -36,70 +36,43 @@ public struct ProfileService: ProfileFetching, Sendable {
...
         } catch let error as HTTPClientError {
             return .failure(.responseError(reason: error.map()))
-        } catch _ as ProfileService.CannotCreateURLFromGivenPath {
-            return .failure(.requestError(reason: .urlInitializationFailed))
         } catch let error as ProfileServiceError {
             return .failure(error)
-        } catch _ as DecodingError {
+        } catch let error as DecodingError {
+            print(error)
             return .failure(.noProfileInResponse)
```
pinarol: Not directly related with this PR but i think we should update the enum case `noProfileInResponse` as `decodingError` since `noProfileInResponse` can be misleading here.

etoledom: Sounds like it could be a case of `ResponseErrorReason`

etoledom:
@pinarol - I believe this is ready for review 👍

pinarol (approved):
Looks good! 🎉 Nice work! 👍

Issue #130, opened 2024-03-22
Improve swift concurrency throughout the SDK


etoledom:
We can use `-strict-concurrency=complete` flag to check the overall implementation:

```swift
.target(
    name: "Gravatar",
    swiftSettings: [
        .unsafeFlags(["-strict-concurrency=complete"])
    ]
),
```

or also

```swift
.target(
    name: "Gravatar",
    swiftSettings: [
        .enableUpcomingFeature("StrictConcurrency")
    ]
),
```

pinarol:
I will add some subtasks here:

- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/156
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/157
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/166
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/158
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/159
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/160
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/161
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/162
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/163
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/164
- [x] https://github.com/Automattic/Gravatar-SDK-iOS/issues/165

PR #251, opened 2024-05-16
[Concurrency] - Make GravatarProfileFetchResult sendable

pinarol:
Closes part of https://github.com/Automattic/Gravatar-SDK-iOS/issues/130 

### Description

Fixes this warning that appears when strict concurrency is enabled.

Screenshot 2024-05-16 at 15 53 55

### Testing Steps

CI green.


etoledom (approved):
Thank you for this! 🙏 

Every time I see this `fetchProfile` with completion handler I feel like I want to remove it 😅 

For now this will do it 👍

PR #252, opened 2024-05-16
[Concurrency] - Make ImageCaching Sendable

pinarol:
Closes part of https://github.com/Automattic/Gravatar-SDK-iOS/issues/130 

### Description

Resolves:

Screenshot 2024-05-16 at 16 02 32

Okay, I thought making ImageCaching functions "async" will force us to make the implementations async too. But it doesn't. One can still implement ImageCaching with no async functions. So this is much flexible for different use cases imo(Including implementing a mock ImageCaching for example). On the other hand, NSCache is already thread safe so I am marking it as "@unchecked Sendable" and keeping our default implementation of ImageCaching as is.

### Testing Steps

CI green.


etoledom (approved):
Nice!

Issue #164, opened 2024-04-03
[Concurrency] Provide async & await setImage in `UIImageView+Gravatar.swift` 


pinarol:
Let's provide async & await setImage methods  in `UIImageView+Gravatar.swift`. Because why not.....

There's one thing to consider before removing the older, compilation handler methods though. If we only provide async-await ones we can't retain the task and then later use it to cancel the ongoing download. This should be done by the caller party. 

The UIImageView extension from the WordPress repo calls our `cancelImageDownload()` method. This can be an indicator that canceling is a useful functionality and we should keep it. So if we end up keeping it we should also keep the compilation handler methods. (I mean those methods will still call the async-await one inside. For example:  [DemoUIImageViewExtensionViewController.swift diff](https://github.com/Automattic/Gravatar-SDK-iOS/pull/136/files#diff-41c5500adf48f5cd8c1bb69786e15e482306d54058fa6d3b58b18f06bbd7ae62))

Whole diffs are here: https://github.com/Automattic/Gravatar-SDK-iOS/pull/136/files

PR #254, opened 2024-05-27
Allow avatar customization

### LINKED ISSUE #253: Allow injecting custom avatars
Allow injecting custom avatars as it's a common thing for 3rd parties to have their own avatar.

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/253

### Description

Introduces `AvatarType` to allow 3rd parties inject their own avatar. There are 3 levels to it:

1. Avatar is a `UIImageView` or a subclass of it. We apply its style normally unless they set `skipStyling = true`.
2. Avatar is a `UIView` that has a `UIImageView` in it. In this case we still take care of downloading and setting the image but we don't apply any style because it's unclear where exactly to apply the style, to the parent or to the `UIImageView`?. It's best to let this be determined by the 3rd party component.
3. Fully custom. Avatar is provided via "AvatarProviding" protocol.

### Testing Steps

CI green.


Sources/GravatarUI/ProfileView/Avatar/AvatarType.swift
```diff
@@ -0,0 +1,87 @@
...
+            provider
+        }
+    }
+}
+
+/// Defines a UIView that has a UIImageView subview.
+@MainActor
+public protocol ImageViewWrapper {
+    // The `UIImageView` to display the avatar.
+    var imageView: UIImageView { get }
```
etoledom:
```suggestion
    /// The `UIImageView` to display the avatar.
    var imageView: UIImageView { get }
```

Sources/GravatarUI/ProfileView/Avatar/AvatarType.swift
```diff
@@ -0,0 +1,87 @@
...
+/// Provides a UIView to be used as the avatar in a `BaseProfileView`. This is for taking full control over the avatar..
+@MainActor
+public protocol AvatarProviding {
+    /// `UIView` to insert into the avatar's slot in a `BaseProfileView`.
+    var avatarView: UIView { get }
+
+    /// `BaseProfileView` calls this method to load the avatar from a URL.
+    /// You may not need to implement this method if the image is downloaded else where.
+    func setImage(with source: URL?, placeholder: UIImage?, options: [ImageSettingOption]?, completion: ((Bool) -> Void)?)
+
```
etoledom, line 79: I thought for a second about making this async, but then I noticed that we don't have an async alternative for the UIImageVIew.gravatar extension 🤔 

Maybe we can do this at some point in the future

etoledom (approved):
Looking great!

Sources/GravatarUI/ProfileView/Avatar/AvatarType.swift
pinarol: Yep, nowadays I'll work on that.

Sources/GravatarUI/ProfileView/Avatar/AvatarType.swift
```diff
@@ -0,0 +1,87 @@
...
+/// Defines the different types of avatars that can be passed to the `BaseProfileView` subclasses.
+public enum AvatarType {
+    /// Avatar is a `UIImageView` or a subclass of it.
+    /// If `skipStyling` is `true` then no size & shape & border changes will be applied to the `UIImageView` so the appearance needs to be managed externally.
+    case imageView(UIImageView, skipStyling: Bool = false)
+
+    /// Avatar is a `UIView` that has a `UIImageView` and its appearance needs to be managed externally.
+    case imageViewWrapper(ImageViewWrapper)
+
+    /// Avatar is a custom `UIView` whose appearance is fully managed externally.
```
pinarol:
```suggestion
    /// Avatar is a custom `UIView` whose appearance and contents are fully managed externally.
```

Sources/GravatarUI/ProfileView/Avatar/AvatarType.swift
```diff
@@ -0,0 +1,87 @@
...
+@MainActor
+public protocol ImageViewWrapper {
+    /// The `UIImageView` to display the avatar.
+    var imageView: UIImageView { get }
+
+    /// The parent view for the imageView. It doesn't have to be the direct parent.
+    var baseView: UIView { get }
+}
+
+/// Provides a UIView to be used as the avatar in a `BaseProfileView`. This is for taking full control over the avatar..
```
pinarol:
```suggestion
/// Provides a UIView to be used as the avatar in a `BaseProfileView`. This is for taking full control over the avatar.
```

Sources/GravatarUI/ProfileView/Avatar/AvatarType.swift
etoledom: Oh right, there's an old ticket for this already https://github.com/Automattic/Gravatar-SDK-iOS/issues/164
Nice! 🙏

PR #255, opened 2024-05-27
Enable StrictConcurrency for targets: Gravatar, GravatarTests

pinarol:
Closes part of #136 

### Description

Enables `StrictConcurrency` for targets: Gravatar, GravatarTests.
Resolves all the remaining warnings in GravatarTests.

### Testing Steps

CI green


Tests/GravatarTests/URLSessionMock.swift
```diff
@@ -12,36 +12,56 @@ final class URLSessionMock: URLSessionProtocol {
     let returnData: Data
     let response: HTTPURLResponse
     let error: NSError?
-    var callsCount = 0
+    private var _callsCount = 0
+    var callsCount: Int {
+        get async {
+            _callsCount
+        }
+    }
```
etoledom: Is this necessary?

I believe that having `private(set) var callsCount` to only be able to mutate it from within the actor should be enough to comply with the concurrency model.

And the same for all other private fields 🤔

pinarol: Unit tests read "callsCount" and assert on its value. So it can't be fully private.

pinarol: Sorry I got your question wrong. Let me check this one more time.

pinarol: You are right we don't need that, private(set) is sufficient. I pushed a new commit for this. Thanks!

etoledom (approved):
This is pretty cool! 🎉 

Thank you 🙏

PR #257, opened 2024-05-28
Adding api key header to v3 profile request

### LINKED ISSUE #256: Add Api-key to v3 profile request

etoledom:
Closes #256 

### Description

Adds a Configuration class to hold the API key globally in memory to be used in requests.
Using this api key in v3 Profile request.

For the demo app, I've added a `Secrets.swift` file to the project, which is being ignored by source control.
This file will be automatically generated when running `make dev` or `make dev-demo`.
It can also be generated specifically with `make secrets`.

I tried to make this happen in the build phase of the project, but there are several problems, like the new file is not immediately added to the current build process, which makes the build fail. And fastlane don't use xcode to build the demo apps.

After this file is generated, it won't be overridden, and we can keep our personal api-key in this generated file without worrying about committing it to source control.

### Testing Steps

- run `make dev-demo`
- Check that `Secrets.swift` has been generated.
- Add your api key to the new `Secrets.swift` file.
- Test the "Fetch Profile" demo screen.
- 🎉


etoledom:
This PR will be a draft until the issues in the backend are solved.

Sources/Gravatar/Network/Services/ProfileService.swift
```diff
@@ -75,3 +74,17 @@ extension ProfileService {
         }
     }
 }
+
+extension URLRequest {
+    private enum HeaderField: String {
+        case authorization = "Authorization"
+    }
+
+    fileprivate func authorized(with key: String?) -> URLRequest {
```
pinarol: Optional: We can maybe set "Configuration.shared.apiKey" as a default value to this param since the value will be same for all requests.

Sources/Gravatar/Network/Services/ProfileService.swift
```diff
@@ -68,10 +67,26 @@ extension ProfileService {
...
     }
 }
+
+extension URLRequest {
+    private enum HeaderField: String {
+        case authorization = "Authorization"
+    }
+
+    @MainActor
+    fileprivate func authorized(with key: String? = Configuration.shared.apiKey) -> URLRequest {
```
etoledom: Interesting...
This compiles just fine when `.unsafeFlags(["-strict-concurrency=complete"])` is set, but it's an error without this flag 🤔 

Should we commit the flag or just not do this change just yet?

The `@MainActor` is there to serialize access to the global variable. And Main Actor since this is most probably to be set in App Delegate or a view controller.

pinarol: Hmm 🤔 I think we can instead use a normal actor here instead of MainActor considering that this runs for every request on a thread other than the main thread. But when we put `@MainActor` here we make it do this header setting thing on the main thread which is a sort of unnecessary switch.

```
public actor Configuration {
    private var _apiKey: String?
    public var apiKey: String? {
        get async {
            _apiKey
        }
    }
    
    public static let shared = Configuration()

    private init() { }
    
    public func configure(with apiKey: String) {
        self._apiKey = apiKey
    }
}
```

I think there's no harm in using `Task { }` inside AppDelegate since now that the inner state is protected by an actor, consequent attempts to reach to the apiKey will wait for this configure(..) call to end. 

```
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        // Override point for customization after application launch.
        if let apiKey = apiKey {
            Task {
                await Configuration.shared.configure(with: apiKey)
            }
        }
        return true
    }
```

in URLRequest extension:

``` 

    fileprivate func authorized(with key: String? = nil) async -> URLRequest {
        let apiKey: String?
        if let key {
            apiKey = key
        }
        else {
            apiKey = await Configuration.shared.apiKey
        }
        guard let apiKey else { return self }
        let bearerKey = "Bearer \(apiKey)"
        var copy = self
        copy.setValue(bearerKey, forHTTPHeaderField: HeaderField.authorization.rawValue)
        return copy
    }
```

etoledom:
```swift
Task {
   await Configuration.shared.configure(with: apiKey)
}
```
This is something I wanted to avoid actually. But it might become a common thing 🤔 
Well let's go with it 👍

pinarol: At first it appears unusual but when you think about it, call to `configure(...)` starts before all the attempts to read the apiKey. And the actor guarantees that all other attempts to read the apiKey will wait for `configure(...)`to finish first. So it should be safe I think.

etoledom:
>  This PR will be a draft until the issues in the backend are solved.

Backend is ready 🎉

pinarol (approved):
👍

PR #260, opened 2024-05-29
ImageDownloader: Fix cancellation and some other problems

pinarol:
Closes #

### Description
 
AFAIU image download cancelling mechanism was broken by https://github.com/Automattic/Gravatar-SDK-iOS/pull/232.

Problem 1:

Cancelling an outer Task doesn't automatically cancel the Tasks created internally. We need to hold those tasks and cancel. Luckily we already hold those tasks in the cache now.

Problem 2:

If a download task fails for some reason it is stuck in the cache as "inProgress", and consequent calls with the same url never succeeds. We should remove the cache entry if the network call is not a success.

Problem 3.

The "imageProcessorFailed" error was not thrown as is but wrapped into .unexpected.

### Testing Steps

Demo app > UIImageView extension
Cancelling an ongoing task should work.
Retrying the cancelled download should also work("Ignore Cache" option needs to be unchecked to test this otherwise there's no bug).

### DIFF
--- a/Sources/Gravatar/Network/Services/ImageDownloadService.swift
+++ b/Sources/Gravatar/Network/Services/ImageDownloadService.swift
@@ -4,7 +4,7 @@ import UIKit
 ///
 /// This is the default type which implements ``ImageDownloader``..
 /// Unless specified otherwise, `ImageDownloadService` will use a `URLSession` based `HTTPClient`, and a in-memory image cache.
-public struct ImageDownloadService: ImageDownloader {
+public struct ImageDownloadService: ImageDownloader, Sendable {
     private let client: HTTPClient
     let imageCache: ImageCaching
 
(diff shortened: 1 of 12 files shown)


Sources/Gravatar/Network/Services/ImageDownloadService.swift
```diff
@@ -20,59 +20,61 @@ public struct ImageDownloadService: ImageDownloader {
...
+        let image: UIImage
+        do {
+            image = try await task.value
+        } catch {
+            await imageCache.setEntry(nil, for: url.absoluteString)
+            throw error
+        }
         await imageCache.setEntry(.ready(image), for: url.absoluteString)
         return ImageDownloadResult(image: image, sourceURL: url)
     }
```
etoledom: This function became a bit heavy, maybe we can break it apart to make it easier to glance over it?

Something like:
```swift
    public func fetchImage(with url: URL, forceRefresh: Bool = false, processingMethod: ImageProcessingMethod = .common()) async throws -> ImageDownloadResult {

// Cache stuff
        if !forceRefresh, let result = try await findCachedImage(with: url) {
            return result
        }

// Task creation
        let request = URLRequest.imageRequest(url: url, forceRefresh: forceRefresh)
        let task = Task<UIImage, Error> {
            try await fetchAndProcessImage(request: request, processor: processingMethod.processor)
        }

// Await for the image
        return try await getImage(from: task, url: url)
    }
```

etoledom (approved):
Looks pretty good! Thanks for fixing this 🙏 

Just one small idea on code review, but not blocking 🚢

Sources/Gravatar/Network/Services/ImageDownloadService.swift
pinarol: Done!

PR #262, opened 2024-05-31
Add "TestHelpers" target to share code between test targets

### LINKED ISSUE #181: Add TestHelpers target to share code between 2 unit test targets
Now that we don't have test specs in podspecs we can add TestHelpers target to share code between GravatarTests and GravatarUITests. We only need to update "Package.swift".

pinarol:
Closes #181 

### Description

Add "TestHelpers" target to share code between test targets.

### Testing Steps

CI green.


etoledom (approved):
Looking great! Thank you 🙏

Issue #166, opened 2024-04-03
[Concurrency] Fix SimpleCounter issue


pinarol:
The SimpleCounter increases the counter, it should be a sort of Actor. Considering that this an internal utility which is only used by the UI components to detect if the task is outdated we can safely mark it with `@MainActor`. We wouldn't want to await on the "next()" function and suspend the thread while this is already being called from the MainActor context. 

Full diffs: https://github.com/Automattic/Gravatar-SDK-iOS/pull/136/files

Issue #165, opened 2024-04-03
[Concurrency] Remove the internal ImageDownloader.fetchImage() method with the compilation handler


pinarol:
This should be done after https://github.com/Automattic/Gravatar-SDK-iOS/issues/164

Issue #162, opened 2024-04-03
[Concurrency] Make ActivityIndicatorType Sendable


pinarol:
There might be different ways to achieve this but I think there's no harm in tagging `ActivityIndicatorProvider` with `@MainActor`. We are sure that it will always run on the main thread. Same with `DefaultActivityIndicatorProvider`. We shouldn't need to await on the activity indicator show/hide methods since they are called from the main actor.

Check https://github.com/Automattic/Gravatar-SDK-iOS/pull/136/files for the full diff.

PR #263, opened 2024-05-31
[Concurrency] UIImageView extension now has async await methods

### LINKED ISSUE #164: [Concurrency] Provide async & await setImage in `UIImageView+Gravatar.swift`

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/164 https://github.com/Automattic/Gravatar-SDK-iOS/issues/165 #166 https://github.com/Automattic/Gravatar-SDK-iOS/issues/162

### Description

- Refactors the UIImageView extension to provide async throws methods. Removes the compilation handler methods.
- Removes `GravatarUI/ImageDownloader+Extension.swift` which contains the compilation handler method for downloading an image. Removes `GravatarUITests/ImageDownloadServiceTests.swift` which was testing the removed method. (We also have `GravatarTests/ImageDownloadServiceTests.swift`, that's a separate test and we should keep it)
- Removes `CancellableDataTask.swift` since it's not used from anywhere anymore.

### Testing Steps

There's no specific testing steps but `Demo app > UIImageView Extension` can be checked.


Sources/GravatarUI/GravatarCompatibleUI/UIImageView+Gravatar.swift
```diff
@@ -175,32 +155,28 @@ extension GravatarWrapper where Component: UIImageView {
...
-    @discardableResult
+    /// - Returns: The ``ImageDownloadResult``.
     public func setImage(
         with source: URL?,
         placeholder: UIImage? = nil,
-        options: [ImageSettingOption]? = nil,
-        completionHandler: ImageSetCompletion? = nil
-    ) -> CancellableDataTask? {
+        options: [ImageSettingOption]? = nil
+    ) async throws -> ImageDownloadResult {
```
etoledom, line 171: This function looks much better now overall, though it's still huge.
It would be good to be able to split it in a few different function calls, but it doesn't have to be on this PR.

etoledom (approved):
Looking great and works as expected 🎉

Sources/GravatarUI/GravatarCompatibleUI/UIImageView+Gravatar.swift
pinarol: I did some splitting now. 👍 This is a tough one though, i felt like splitting further can also make it more difficult to read and understand.

PR #271, opened 2024-06-05
Update ProfileServiceError & Cleanup ProfileService

### LINKED ISSUE #266: Update "noProfileInResponse" case in ProfileServiceError
We currently throw ProfileServiceError.noProfileInResponse when decoding the response fails and it's misleading. Let's update this.

pinarol:
Closes #266 

### Description

Updates these things in ProfileService:

- Updated ProfileServiceError to APIError, since it applies to all potential REST calls not just to the ProfileService. 
- Replaced "noProfileInResponse" with "decodingError".
- Removed some unnecessary and buggy result mapping code from ProfileService.
- Removed the compilation handler method in the ProfileService. Just for sake of less code.
- Added some unit tests for error cases.

### Testing Steps

Demo app > Fetch Profile


etoledom (approved):
Neat!

PR #275, opened 2024-06-06
Update README.md

etoledom:
### Description

- Adding an extra feature copied from [Gravatar docs portal](https://docs.gravatar.com/sdk/ios-sdk)
- Adding a section about creating an API Key, copied from [Gravatar docs portal](https://docs.gravatar.com/sdk/ios-sdk)


pinarol (approved):
Nice 🎉

PR #276, opened 2024-06-06
Remove the task from the cache when cancelled

pinarol:
Closes #

### Description

I bumped into a bug while testing the WP side. There, the existing download is [cancelled](https://github.com/wordpress-mobile/WordPress-iOS/blob/fb271448cf9231aaae2cbc8862f92e3e8ddde3c8/WordPress/Classes/ViewRelated/Gravatar/UIImageView%2BGravatar.swift#L61) before each download attempt. Even though it's not necessary to do that it should work fine(especially after the CacheEntry.inProgress option no one needs to worry about multiple requests for the same URL but that's a separate topic). 

Anyway, I come across with a screen where the avatar remained empty and all download attempts resulted in "cancelled". I think it can be because we don't remove the task from the cache when cancelling it, so next time we try to download the same image it tries to await a cancelled task, which then results in a cancelled error to be thrown. But it is not easy to reproduce and I was able to repro the issue only in post revisions page. (somehow that page attempts to download the image multiple times even if there's only 1 cell, I guess the tableview is reloaded multiple times).

### Testing Steps

Let's first try to reproduce the issue:

Checkout this PR https://github.com/wordpress-mobile/WordPress-iOS/pull/23320
Login to Jetpack app
Go to Posts > Open one of your posts
Use the ... on the top to open the menu
Tap on the Revisions item
(If you don't see Revisions there it means you didn't edit the file yet. In that case edit and publish the post. Then check again.)

 

The avatar remains empty. (I am not sure how consistent you can repro the issue though. I suspect that it's a race condition)

Now point Gravatar SDK to the latest commit from this PR and test it again. The avatar should load fine.

### DIFF
--- a/Sources/Gravatar/Network/Services/ImageDownloadService.swift
+++ b/Sources/Gravatar/Network/Services/ImageDownloadService.swift
@@ -79,6 +79,7 @@ public struct ImageDownloadService: ImageDownloader, Sendable {
             case .inProgress(let task):
                 if !task.isCancelled {
                     task.cancel()
+                    await imageCache.setEntry(nil, for: url.absoluteString)
                 }
             default:
                 break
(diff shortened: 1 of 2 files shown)


etoledom (approved):
I was able to reproduce the issue on WPiOS.

I have a post with many revisions, and when the image cache was cleared, some of the profile images in the revision history didn't load.

After using this fix, all profile image in the same history loaded every time.
It's safe to say that this fix solves the issue 🎉

PR #277, opened 2024-06-07
Add `@discardableResult` to `setImage(with source: URL?, ...)`

pinarol:
Closes #

### Description

Let's add `@discardableResult` here too since the result is in fact discardable. The other public method ( `setImage(
        avatarID:...)`) already has it.

### Testing Steps

CI green.


etoledom (approved):
Cool!

PR #280, opened 2024-06-07
Update .spi.yml

etoledom:
Closes #

### Description

Updating the spi.yml file to fix a syntax error

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -5,9 +5,7 @@
 
 version: 1
 builder:
-  configs:
-     - documentation_targets:
-      - GravatarUI
-      - Gravatar
+    configs:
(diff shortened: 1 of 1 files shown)


.spi.yml
```diff
@@ -5,9 +5,7 @@
 
 version: 1
 builder:
-  configs:
-     - documentation_targets:
-      - GravatarUI
-      - Gravatar
+    configs:
+    - documentation_targets: [Gravatar, GravatarUI]
```
etoledom, line 9: The list with dashes is also a valid syntax, but it makes it more difficult to see the correct indentation spaces.
In this way now is more clear.

pinarol (approved):

PR #284, opened 2024-06-12
Small fixes on 2.0.0

pinarol:
Closes part of #236 

### Description

Make `Palette.init(...)` public.
Give a default tint color to the avatar's activity indicator
Fix the logic about avatar placeholder so that custom color palette can be applied properly, add some snapshot tests to catch this bug in the future.

### Testing Steps

CI green


etoledom (approved):
Looking good! 🎉

PR #289, opened 2024-06-18
Etoledom/claim profile image multicolor

### LINKED ISSUE #288: Make Claim Profile placeholder image paintable with different colors
Ideally we are able to paint the image, the background and the border with different colors.

etoledom:
Closes #288 

### Description

This PR allows the Claim Profile image placeholder to be painted with different colors:

- Foreground (the face shape)
- Background
- Border

To achieve this, the colors on `Palette.AvatarColors` are used.

For the image view background color when no image is set (in placeholder state), the `Palette.PlaceholderColors` color is used.

In this way, arbitrary color convinations can be achieved:

image

### Testing Steps

- Test the profile views on the demo app, fetching for inexistent profiles.
- Check that the profile image placeholder colors look good.


Sources/GravatarUI/ProfileView/ProfileViewConfiguration.swift
```diff
@@ -166,23 +168,28 @@ extension ProfileViewConfiguration {
...
         case .summary:
-            ProfileViewConfiguration.summary(model: ClaimProfileModel(userName: userName), palette: palette).configureAsClaim()
+            ProfileViewConfiguration.summary(model: ClaimProfileModel(userName: userName), palette: palette).configureAsClaim(borderWidth: 1)
         }
     }
 }
 
 extension ProfileViewConfiguration {
-    private func configureAsClaim() -> ProfileViewConfiguration {
+    private func configureAsClaim(borderWidth: CGFloat = 2) -> ProfileViewConfiguration {
```
etoledom, line 183: Previously, the border came together with the image.
Now, UIKit is drawing the border for it to be able to be painted with a different color.

I noticed that on the big avatar images, the border of 1pt was way much thinner than the original, and 2pt looked too thick on the small ones. So I decided to have different border width in every case.

etoledom:
Replaced by https://github.com/Automattic/Gravatar-SDK-iOS/pull/292

PR #292, opened 2024-06-26
Use multicolor placeholder for claim placeholder

### LINKED ISSUE #288: Make Claim Profile placeholder image paintable with different colors

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/288

### Description

Easiest way I could find to apply multicolor placeholder without changing some internal logic. This solution is done considering,

- The background and tint color of avatar depend on both the placeholder and the palette type. We should consider both when customizing those colors.
- The avatar colors we use for the claim-profile state are not easily applicable to the other placeholders. 

So I come up with a way to customize background and tint color of avatar from outside. This way, avatar colors can be changed according to the placeholder when necessary.

### Testing Steps

Demo app related screens can be checked, palette change should be tested too.


Sources/GravatarUI/ProfileView/Placeholder/PlaceholderDisplayers/BackgroundColorPlaceholderDisplayer.swift
```diff
@@ -48,15 +48,21 @@ class BackgroundColorPlaceholderDisplayer<T: UIView>: PlaceholderDisplaying {
 @MainActor
 class AvatarPlaceholderDisplayer<T: UIView>: BackgroundColorPlaceholderDisplayer<T> {
     private var originalBorderWidth: CGFloat = 0
+    private var isShowing: Bool = false
 
     override func showPlaceholder() {
         super.showPlaceholder()
+        // Don't set originalBorderWidth if isShowing is true, otherwise it overrides the original value.
+        guard !isShowing else { return }
+        isShowing = true
```
pinarol, line 57: This fix is not directly related but it was necessary to preserve the borderWidth after the placeholders are hidden.

Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -232,13 +244,13 @@ open class BaseProfileView: UIView, UIContentView {
...
     ) {
-        self.paletteType = paletteType ?? .system
+        self._paletteType = paletteType ?? .system
         self.avatarLength = avatarLength ?? Self.avatarLength
         let placeholderDisplayer = ProfileViewPlaceholderDisplayer()
         self.placeholderDisplayer = placeholderDisplayer
         self.activityIndicator = ProfilePlaceholderActivityIndicator(placeholderDisplayer: placeholderDisplayer)
         self.avatarType = (avatarType ?? AvatarType.imageView(UIImageView()))
-        self.avatarProvider = self.avatarType.avatarProvider(avatarLength: self.avatarLength, paletteType: self.paletteType)
+        self.avatarProvider = self.avatarType.avatarProvider(avatarLength: self.avatarLength, paletteType: self._paletteType)
```
pinarol, line 253: Small note: This is inside the init() so I couldn't use `self.paletteType` here.

Tests/GravatarUITests/__Snapshots__/LargeProfileSummaryViewTests/testLargeProfileSummaryViewLoadingStateClearsWhenDataIsPresent.light.png
etoledom, line 1: The new snapshot here has a different border color, not sure if this is expected 🤔

pinarol: yes it's because I fixed something here https://github.com/Automattic/Gravatar-SDK-iOS/pull/292/files#r1655280360

The borders are expected to appear in this snapshot because we have borderWidth == 1. It didn't appear before because there was a bug that caused losing the "original" value of the `borderWidth` when `showPlaceholder()` is called more than once consequently.

etoledom (approved):
Looking great 🎉

PR #297, opened 2024-06-28
Solve the concurrency warning that appears in XCode 16 beta 2

pinarol:
Closes #

### Description

It is early for this but it seems like a harmless change even before switching to XCode 16.

Solves this warning that appears on XCode 16 beta 2

Screenshot 2024-06-28 at 15 36 54

### Testing Steps

The ImageView extension page from the demo app can be tested.


etoledom (approved):
Looks good 🎉

PR #302, opened 2024-07-01
Fix docc deploy to GH pages

etoledom:
### Description

Updating macos and xcode to use swift v5.10

**Note**: `"etoledom/gh-docs"` needs to be removed before merge

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -6,6 +6,6 @@
 version: 1
 builder:
     configs:
-    - documentation_targets: [Gravatar, GravatarUI]
+    - documentation_targets: [Gravatar]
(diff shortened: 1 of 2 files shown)


.spi.yml
```diff
@@ -6,6 +6,6 @@
 version: 1
 builder:
     configs:
-    - documentation_targets: [Gravatar, GravatarUI]
+    - documentation_targets: [Gravatar]
```
etoledom, line 9: This is a test to see if in this way we get the docs build on the Swift Package Index.

pinarol (approved):
LGTM 👍

Issue #303, opened 2024-07-01
Do not use `Profile.avatarUrl` in profile views


pinarol:
`/v3/profiles/...`  response,`Profile.avatarUrl` always gives the primary email's avatar even if you make the query with a secondary email. From the backend perspective secondary emails don’t have profiles, they just redirect to primary email’s profile. This causes us to show wrong avatar in the profile view.

So let's calculate the avatar identifier based on the email rather than taking from the profile response.

pinarol:
Closed with https://github.com/Automattic/Gravatar-SDK-iOS/pull/304

PR #304, opened 2024-07-01
Decouple avatarID from Profile model

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/303

### Description

Adding a new `avatarIdentifier` var into `ProfileViewConfiguration` to let the other parties determine the avatar id independent from the profile response. `Profile.avatarUrl` always gives the primary email's avatar even if you make the query with a secondary email.

### Testing Steps

Prerequisites: 
Have a secondary email in your Gravatar account and assign a different avatar for that email.

**Test 1 - Test old versions of demo apps**

This is to test backward compatibility.

Checkout the commit 9262beaa8b9f3bf4f2bfa553215f6253a1dc8996 from this PR and check these demo app screens:

* Profile Card
* Profile View Controller
* Profile Card Configuration

Fetch the profile for the secondary email.

Expected: The primary email's avatar loads and you don't notice a bug.

**Test 2 - Test new versions of demo apps**

Check the latest commit in this PR.

Check the same screens again with the secondary email.

Expected: The secondary email's avatar loads.

### DIFF
--- a/Sources/Gravatar/Identifiers/ProfileIdentifier.swift
+++ b/Sources/Gravatar/Identifiers/ProfileIdentifier.swift
@@ -37,4 +37,14 @@ extension ProfileIdentifier: Identifiable {
             hashID.id
         }
     }
+
+    // Returns the `AvatarIdentifier` popuplated from `self`.
+    public var avatarIdentifier: AvatarIdentifier {
+        switch self {
+        case .email(let email):
+            AvatarIdentifier.email(email.string)
+        case .hashID(let hashID):
+            AvatarIdentifier.hashID(hashID.string)
+        }
+    }
 }
--- a/.buildkite/commands/publish-pod.sh
+++ b/.buildkite/commands/publish-pod.sh
@@ -12,7 +12,7 @@ echo "--- :rubygems: Setting up Gems"
 install_gems
 
 echo "--- :cocoapods: Publishing $PODSPEC_PATH to CocoaPods CDN"
-publish_pod --synchronous "$PODSPEC_PATH"
+publish_pod --allow-warnings --synchronous "$PODSPEC_PATH"
 
 echo "--- :slack: Notifying Slack"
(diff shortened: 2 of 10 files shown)


Sources/Gravatar/Identifiers/ProfileIdentifier.swift
```diff
@@ -37,4 +37,14 @@ extension ProfileIdentifier: Identifiable {
...
     }
+
+    // Returns the `AvatarIdentifier` popuplated from `self`.
+    public var avatarIdentifier: AvatarIdentifier {
+        switch self {
+        case .email(let email):
+            AvatarIdentifier.email(email.string)
+        case .hashID(let hashID):
+            AvatarIdentifier.hashID(hashID.string)
+        }
```
etoledom, line 48:
```suggestion
        AvatarIdentifier.hashID(id)
```

I believe this could be simplified like this 🤔 
The switch is already performed by the `id` var

etoledom (approved):
Looks good! Thank you for the fix 🙏

Sources/Gravatar/Identifiers/ProfileIdentifier.swift
pinarol: It would be shorter for sure but I thought it would be good to keep the original character of the id, like if there's an email involved we still keep it in the `.email` enum. 🤔

etoledom: Oh I see. As we shouldn't need the literal email it shouldn't really matter, right? 
But not big feelings about this 👍 
Feel free to keep it as it is

pinarol: Nope we don't need the email directly and things should work with just hash. I just had a hunch like, as a 3rd party dev if i created the profile identifier with `.email(...)` then I would expect that this conversion respects my choice.

PR #285, opened 2024-06-12
Updating to xcode 15.4

etoledom:
Closes #273

### Description

Updating Xcode to 15.4, plus Swift to 5.10, since it's the swift version used by xcode 15.4.

### Testing Steps

CI 🍏

PR #307, opened 2024-07-02
Fix SwiftFormat Warnings

### LINKED ISSUE #306: Fix warnings in SwiftFormat
There are some warnings when running swiftformat:

```
warning: 'gravatar-sdk-ios': found 2 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    /Users/andrew/Documents/Git/GitHub/Gravatar/Gravatar-SDK-iOS/Sources/GravatarUI/Resources/AccountIcons.xcassets
    /Users/andrew/Documents/Git/GitHub/Gravatar/Gravatar-SDK-iOS/Sources/GravatarUI/Resources/Media.xcassets
```

```
warning: No Swift version was specified, so some formatting features were disabled. Specify the version of Swift you are using with the --swiftversion option, or by adding a .swift-version file to your project.
```

```
warning: Failed to write cache file at /Users/andrew/Library/Caches/com.charcoaldesign.swiftformat/swiftformat.cache.
```

andrewdmontgomery:
Closes #306

### Description

This resolves a few warnings that are appearing in the output of running `swiftformat`  (see issue #306)

#### Where should we record the version?
SwiftFormat wants us to specify the version of Swift to use ([SwiftFormat docs](https://github.com/nicklockwood/SwiftFormat?tab=readme-ov-file#swift-version)).  It prefers to find that version in a `.swift-version` file.  But we removed it in #285 .  In this PR, I've moved that into the `.swiftformat` file.  But I worry about missing this configuration setting the next time we update the version of Swift.

Thoughts?  Perhaps this is a good reason to create documentation around the process of updating Xcode.  If we did that, keeping this setting in the `.swiftformat` file would be a safe choice.

### Testing Steps
- [ ] Run `make swiftformat` and observe that there are no warnings (see issue #306 for examples)
- [ ] Run `make lint` and observe that there are no warnings there either

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,3 +1,5 @@
+--swiftversion 5.10
+
 ## File options
 
 --exclude .build,**/*.docc/**
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,3 +1,8 @@
+## Swift version
+
+# Some rules are applicable only to newer versions of Swift.
+# Any rules applicable to Swift versions that are newer than the version
+# specified here will be disabled.
 --swiftversion 5.10
 
 ## File options
--- a/Gravatar.podspec
+++ b/Gravatar.podspec
@@ -14,7 +14,7 @@ Pod::Spec.new do |s|
   }
   s.documentation_url = 'https://automattic.github.io/Gravatar-SDK-iOS/'
     
-  s.swift_version     = Gravatar::SWIFT_VERSION
+  s.swift_versions    = Gravatar::SWIFT_VERSIONS
 
   ios_deployment_target = '15.0'
 
--- a/GravatarUI.podspec
+++ b/GravatarUI.podspec
@@ -14,7 +14,7 @@ Pod::Spec.new do |s|
     }
     s.documentation_url = 'https://automattic.github.io/Gravatar-SDK-iOS/'
       
-    s.swift_version     = Gravatar::SWIFT_VERSION
+    s.swift_versions    = Gravatar::SWIFT_VERSIONS
     
     ios_deployment_target = '15.0'
     s.ios.deployment_target = ios_deployment_target
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@
 #   % make help
 
 # Cache
+# No spaces allowed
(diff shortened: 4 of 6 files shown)


etoledom (approved):
Looking great! No warnings on linting now 🎉

andrewdmontgomery:
After a [Slack discussion](https://a8c.slack.com/archives/C06FRGK6Y5R/p1719945473230659), I think moving the `--swiftversion` into the `.swiftformat` file (and dropping the `.swift-version` file) makes sense.  From the SwiftFormat [README.md](https://github.com/nicklockwood/SwiftFormat?tab=readme-ov-file#swift-version):

> Most SwiftFormat rules are version-agnostic, but some are applicable only to newer Swift versions. These rules will be disabled automatically if the Swift version is not specified, so to make sure that the full functionality is available you should specify the version of Swift that is used by your project.

In practice, we will only need to change this when we update SwiftFormat (which may contain new rules for newer versions of Swift).  So it really is a configuration setting for SwiftFormat, and there is no useful reason to keep it in sync with CocoaPods.

## Swift Versions in CocoaPods
Why did CocoaPods move away from using the `.swift-version` file?  They dropped it when they introduced [Multiple Swift Versions Support](https://github.com/CocoaPods/CocoaPods/issues/8191):

> Note: While the pod author was initially allowed to publish their podspec using .swift-version file or --swift-version parameter, neither of them are preserved when a consumer is integrating their pod, therefore it is imperative to motivate pod authors to update their podspec to use the swift_versions DSL going forward.

So we really do want to specify it in the podspec.

I've also updated the podspecs to use the `swift_versions = ['5.10']` syntax, in order to make it explicit that we can support multiple versions of Swift via CocoaPods.

So for example, if in the future, we retain support for Swift 5.10 (e.g. we leave the Package.swift configured with `// swift-tools-version: 5.10`), then we could expand CocoaPods to support additional versions of Swift.

andrewdmontgomery:
@pinarol does this all seem good to you?  I know we originally discussed putting the `.swift-version` file back.  But I believe this makes more sense.

pinarol (approved):
Yes, LGTM 🎉

PR #308, opened 2024-07-03
[Draft] Avatar picker

etoledom:
Closes #

To do:

- [ ] Implement v3 version of image upload (not ready yet)
- [ ] Implement new oauth UI with magic code (not sure if it's ready)
- [ ] UI for unauthorized (the user cancels the promp to oauth) (or maybe we just dismiss the editor 🤔 )
- [ ] Error handlings. Find a proper way to show possible request errors (maybe with the toast)
- [ ] Notify back to the 3rd party of the new image selected
- [ ] Refresh local image cash with the new uploaded image
- [ ] Review the code already in place
  - It would be good to make PRs agains this branch for discussions
- [ ] Injectable image picker replacing the native one (by 3rd parties) 
- [ ] Add open-api specs of the new endpoints (when ready)


Demo/Demo/Gravatar-Demo/MainTableViewController.swift
```diff
@@ -7,6 +7,7 @@
 
 import Foundation
 import UIKit
+import GravatarUI
```
etoledom, line 10: Not needed

Demo/Demo/Gravatar-Demo/Secrets.tpl
```diff
@@ -0,0 +1,6 @@
+struct Secrets {
+    static let apiKey: String? = nil
+    static let clientID: String = ""
+    static let clientSecret: String = ""
+    static let redirectURI: String = ""
+}
```
etoledom, line 6: I created a template for the secrets file to make it easier to create it with the script.
Nothing else changed, but you might need to remove a current `Secrets.swift` to regenerate a new one if you have the old version.

Sources/Gravatar/Configuration.swift
```diff
@@ -1,9 +1,29 @@
 import Foundation
+import SensitiveContentAnalysis
+
+@globalActor
+public actor ConfigActor: GlobalActor {
+    public static let shared = ConfigActor()
+}
```
etoledom, line 7: This is a test.

Setting up the configuration properties to other configuration related structs was complex adding many awaits. But making them managed by a global Configuration Actor, there's no need for this any longer.

So far I like how it works, though I'm open for other names (or solutions).

Sources/Gravatar/Network/Services/HTTPClient.swift
```diff
@@ -16,4 +16,6 @@ public protocol HTTPClient: Sendable {
     ///   - data: The data to be uploaded.
     /// - Returns: An asynchronously-delivered instance of the returned HTTPURLResponse.
     func uploadData(with request: URLRequest, data: Data) async throws -> HTTPURLResponse
+
+    func performDataTask(with request: URLRequest) async throws -> (Data, HTTPURLResponse)
```
etoledom, line 20: Out current `fetchData` can be used for any request method, not just `GET`.
Now that we have a `POST` request, it feels wrong to run it using `fetchData`, so probably a more generic name is better, while we can deprecate `fetchData`.

Sources/GravatarUI/ProfileEditor/AvatarPicker/AvatarCollectionViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+                        with: avatarModel?.url,
+                        placeholder: UIImage(systemName: "person.circle.fill")
+                    )
+                }
+            case .local(let image):
+                cell.imageView.image = image
+                cell.isLoading = true
+                return
+            case .none: break
+            }
```
etoledom: We can probably use the UIContentView pattern here to simplify this.

Sources/GravatarUI/ProfileEditor/AvatarPicker/AvatarImageModel.swift
```diff
@@ -0,0 +1,37 @@
+import UIKit
+
+struct AvatarImageModel: Hashable, Identifiable {
+    enum State: Hashable {
+        case remote(url: String, isLoading: Bool)
+        case local(image: UIImage)
+    }
```
etoledom: This should probably better be called `Source`.

Sources/GravatarUI/ProfileEditor/AvatarPicker/AvatarPickerViewController.swift
```diff
@@ -0,0 +1,232 @@
...
+
+    func uploadImage(_ image: UIImage) async {
+        let imageService = AvatarService()
+        do {
+            try await imageService.upload(image, email: email, accessToken: authToken)
+            presentAvatarUpdatedToast()
+        } catch {
+            print("Error: \(error)")
+        }
+    }
```
etoledom, line 191: This is still not great. After the update finishes, we don't have information about the new remote image.
The v1 image upload endpoint returns something like:
```json
{
  "entry": {
    "file": {
      "name": "2a7090185f86c545f2933659b4e0b894.jpg",
      "size": 2804160,
      "type": "image/png",
      "typeid": 1,
      "full_path": "profile.png"
    }
  }
}
```
Which is not enough information to construct the image_url.

I have asked for a new version (on v3) of this endpoint which will return the same json object than the avatar list endpoint. With this, we can update the internal metadata of the newly updated image

Sources/GravatarUI/ProfileEditor/ProfileEditorViewController.swift
```diff
@@ -0,0 +1,94 @@
+import Gravatar
+import UIKit
+
+/// Base controller to edit a profile, given a profile ID.
+public class ProfileEditorViewController: UITabBarController {
```
etoledom, line 5: The idea of this is to be able to init the Gravatar editor with any controller.
Now we do it with the Avatar picker, since it's the only one. But in other instances we could init with a Profile editor, links editor, etc... Or also a prompt to log in, if the user cancelled the initial prompt.

Sources/GravatarUI/ProfileEditor/ToastViewController.swift
```diff
@@ -0,0 +1,198 @@
+import UIKit
+
+private typealias Attribute = AttributeScopes.UIKitAttributes
+
+class ToastViewController: UIViewController {
```
etoledom, line 5: The toast is a normal view controller presented modally.

The "toast feel" is a custom modal presentation.

pinarol:
@etoledom is this draft PR still necessary?

PR #309, opened 2024-07-04
Fix errors on xcode_16 with swift 6

etoledom:
Closes #

### Description

This PR fixes some errors on Xcode 16, when it's set to build Swift 6.

### Testing Steps

- Using Xcode 16 beta 2
- Open the demo project
- Set `// swift-tools-version: 6.0` on `Package.swift`
- Set `Swift Language Version: Swift 6` On the demo project's Build Settings.
- Build
- Check that there are no warnings or errors.

### DIFF
--- a/Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
+++ b/Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
@@ -64,7 +64,9 @@ class DemoFetchProfileViewController: UIViewController {
         title = "Fetch Profile"
         view.backgroundColor = .white
 
-        [segmentedControl, emailField, fetchProfileButton, activityIndicator, profileTextView].forEach(rootStackView.addArrangedSubview)
+        for view in [segmentedControl, emailField, fetchProfileButton, activityIndicator, profileTextView] {
+            rootStackView.addArrangedSubview(view)
+        }
         view.addSubview(rootStackView)
 
         NSLayoutConstraint.activate([
--- a/Sources/GravatarUI/ProfileView/BaseProfileView.swift
+++ b/Sources/GravatarUI/ProfileView/BaseProfileView.swift
@@ -487,6 +487,7 @@ open class BaseProfileView: UIView, UIContentView {
 }
 
 /// Methods for managing actions on the profile view.
+@MainActor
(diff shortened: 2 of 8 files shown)


Demo/Demo/Gravatar-Demo/DemoFetchProfileViewController.swift
```diff
@@ -64,7 +64,9 @@ class DemoFetchProfileViewController: UIViewController {
         title = "Fetch Profile"
         view.backgroundColor = .white
 
-        [segmentedControl, emailField, fetchProfileButton, activityIndicator, profileTextView].forEach(rootStackView.addArrangedSubview)
+        for view in [segmentedControl, emailField, fetchProfileButton, activityIndicator, profileTextView] {
+            rootStackView.addArrangedSubview(view)
+        }
```
etoledom, line 69: This fixes the error:
```
.../DemoFetchProfileViewController.swift:66:9 
Call can throw, but it is not marked with 'try' and the error is not handled
```

Sources/GravatarUI/ProfileView/BaseProfileView.swift
```diff
@@ -487,6 +487,7 @@ open class BaseProfileView: UIView, UIContentView {
 }
 
 /// Methods for managing actions on the profile view.
+@MainActor
 public protocol ProfileViewDelegate: NSObjectProtocol {
```
etoledom, line 494: This fixes the error:
```
.../DemoProfileConfigurationViewController.swift:91:10 Main actor-isolated instance method 'profileView(_:didTapOnAvatarWithID:)' cannot be used to satisfy nonisolated protocol requirement
```

Probably the most common delegate is going to be a UIViewController or at least will execute UI code, so this makes those implementations easier.

The alternative is on the implementation side, to mark the delegate functions as `nonisolated` and then execute any UI-related code inside a main actor task. Something like:

```swift
nonisolated func profileView(_ view: BaseProfileView, didTapOnAccountButtonWithModel accountModel: AccountModel) {
    guard let accountURL = accountModel.accountURL else { return }
    Task { @MainActor in
        let safari = SFSafariViewController(url: accountURL)
        present(safari, animated: true)
    }
}
```

pinarol (approved):
Looks good 🎉

PR #310, opened 2024-07-08
Add SwiftUI CachedAsyncImage and AvatarView

pinarol:
Closes #

### Description

Adds 2 SwiftUI components

**CachedAsyncImage**, mimics **AsyncImage** API but also adds caching capabilities by using our ImageDownloadService under the hood.

**AvatarView** uses **CachedAsyncImage** and renders the given AvatarURL. 

I am keeping these internal for now since they might be prone to change.

### Testing Steps

Gravatar-SwiftUI-Demo app > Avatar View


Sources/GravatarUI/SwiftUI/AvatarView.swift
```diff
@@ -0,0 +1,149 @@
...
+        self.cache = cache
+        self.loadingView = loadingView
+        self.urlSession = urlSession
+        self._forceRefresh = forceRefresh
+        self.imageContent = imageContent
+        self.transaction = transaction
+    }
+
+    /// Use this initializer to create the content `Image` with a given clipShape, borderColor and borderWidth.
+    init<ClipShape>(
```
etoledom: I would expect to set these decorations via modifier functions 🤔 
And for common use cases, some helper.

Something like:
```swift
AvatarView()
    .cornerRadius(20)
    .borderColor(.blue)
    .borderWidth(2)
```

Is there a reason why is it better to set these properties via initializer?

pinarol: Good idea indeed. I removed that init in favor of an `avatarShape` function, but unfortunately shape was necessary to put a border so those operations are coupled. I called it `avatarShape` to prevent possible name conflicts 🤔 .

Sources/GravatarUI/SwiftUI/AvatarView.swift
```diff
@@ -0,0 +1,105 @@
...
+                scaledImage(placeholder)
+            }
+        }
+    }
+
+    private func scaledImage(_ image: Image) -> some View {
+        image
+            .resizable()
+            .scaledToFit()
+    }
```
etoledom, line 75: Small thing:

Could this be maybe a private extension over Image?
In that case we can do something like:
```swift 
placeholder?.scaledImage()
```
Instead of the `if let` block.

Sources/GravatarUI/SwiftUI/AvatarView.swift
```diff
@@ -0,0 +1,105 @@
...
+
+    func avatarShape(_ shape: some Shape, borderColor: Color = .clear, borderWidth: CGFloat = 0) -> some View {
+        self
+            .clipShape(shape)
+            .overlay(
+                shape
+                    .stroke(borderColor, lineWidth: borderWidth)
+            )
+    }
```
etoledom, line 84: I spent some time trying to figure out if there's a better way to do something like this.

`.avatarShape(RoundedRectangle(cornerRadius: 20), borderColor: Color.accentColor, borderWidth: 2)`

still feels to me like too verbose (maybe because of `RoundedRectangle(cornerRadius: 20),` and because all parameters have to go together).

I don't think this is wrong right now, so let's move forward. Might as well be the SwiftUI way.

But I'd look more in depth into style modifiers. I believe we should be able to eventually implement something like this:

https://peterfriese.github.io/Building-SwiftUI-Components-Tutorial/tutorials/buildingreusablecomponentstutorial/02-creating-a-view-modifier-for-the-avatar-shape/

etoledom: In this way we can apply avatar styles over a grouping view like a Stack or a List.

Sources/GravatarUI/SwiftUI/AvatarView.swift
pinarol: Makes sense. I am removing that `.scaledToFit()` line in the next PR. so at this point I am not even sure if that's worthy of an extension because there will be only .resizable() left.

Sources/GravatarUI/SwiftUI/AvatarView.swift
pinarol: I am converting this into a `extension View ` in the next PR. But the difference with Apple's example is that we also need to add a border, and for border we need to know the shape. In a way, applying a border requires to have a shape too. We can maybe separate applying a shape from applying a border, but then the order of things will become important. You will need to know that you need to apply the shape first. Or, you will need to make the internal implementation a bit more complicated so whatever order you set shape and border it will just work. But I am not sure if it's worth that complexity.

I think letting the user apply a `Shape` is better than giving them an enum of options, because as soon as it is a `Shape` we don't need to care about the actual implementation. This is better regarding the [open/closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). This way we don't need add a new case every time one wants a different shape. And it doesn't get any simpler to use

```
.shape(.circle)
```
is not much different from:

```
.shape(Circle())
```

```
.shape(.rectangle)
.cornerRadius(20)
```
is not much different from:

```
.shape(RoundedRectangle(cornerRadius: 20))
```

Sources/GravatarUI/SwiftUI/AvatarView.swift
etoledom: > because there will be only .resizable() left

If the `if let` block is not needed anymore then its all good 👍

etoledom (approved):
Looking great!
Let's go with this one 👍

PR #316, opened 2024-07-10
Deprecating public CodingKeys and init methods

etoledom:
### Description

On this PR we modify the OpenApi template used to generate the swift struct from the OpenApi specs.

The idea is to deprecate the public `CodingKeys` and the  parametrised `init` method from the generated structs, since they are prone to generate breaking changes on updates to the structures.

We can remove these changes from the template once the deprecation takes effect.

### Testing Steps

- CI should be 🍏


pinarol (approved):
Looks good 🎉

PR #317, opened 2024-07-11
Adding interests to Profile

### LINKED ISSUE #314: Add Interest to Profile
Updating openapi spec

etoledom:
Closes #314 

### Description

This PR updates the openapi spec to bring profile interests.

### Testing Steps

- Have a gravatar account with interests added
- Add you API key to the demo app's secrets file
- Run the Demo app
- Go to "Fetch Profile" screen 
- Fetch for your gravatar account with interests
- Check that the interests are shown on the response text view.


Sources/Gravatar/OpenApi/Generated/GalleryImage.swift
```diff
@@ -5,25 +5,37 @@ import Foundation
...
+    // Avoid deleting this init until the deprecation of is applied.
+    init(url: String, altText: String? = nil) {
+        self.url = url
+        self.altText = altText
+    }
+
     @available(*, deprecated, message: "CodingKeys will become internal on the next release.")
     public enum CodingKeys: String, CodingKey, CaseIterable {
         case url
+        case altText = "alt_text"
```
pinarol: I am afraid this is _technically_ a breaking change so we shouldn't add the new case to the existing enum :( seems like something else we'll need to do manually 🤔, i mean remove manually.

Sources/Gravatar/OpenApi/Generated/Profile.swift
```diff
@@ -99,6 +147,7 @@ public struct Profile: Codable, Hashable, Sendable {
         case pronunciation
         case pronouns
         case links
+        case interests
```
pinarol: Same

Sources/Gravatar/OpenApi/Generated/VerifiedAccount.swift
```diff
@@ -15,16 +17,28 @@ public struct VerifiedAccount: Codable, Hashable, Sendable {
...
+    init(serviceType: String, serviceLabel: String, serviceIcon: String, url: String) {
+        self.serviceType = serviceType
+        self.serviceLabel = serviceLabel
+        self.serviceIcon = serviceIcon
+        self.url = url
     }
 
     @available(*, deprecated, message: "CodingKeys will become internal on the next release.")
     public enum CodingKeys: String, CodingKey, CaseIterable {
+        case serviceType = "service_type"
```
pinarol: Same

etoledom:
From https://github.com/Automattic/Gravatar-SDK-iOS/pull/317#discussion_r1676051699

>I am afraid this is technically a breaking change so we shouldn't add the new case to the existing enum :( seems like something else we'll need to do manually 🤔, i mean remove manually.

@pinarol - Since we keep `CodingKeys` and `InternalCodingKeys` together, we need to write the decoding init "manually" for it to use `InternalCodingKeys`. I have modified the templates to write this init for us for now using `InternalCodingKeys`.

When the compiler writes this init, it will prefer to use `CodingKeys`.

These inits are public, but they still exist even if they are not written literally, so I believe that removing them later won't be a breaking change.

pinarol (approved):
LGTM 👍

PR #318, opened 2024-07-12
Add basic SwiftUI Avatar picker

pinarol:
Closes #

### Description

Adds a very basic `AvatarPickerView`. It takes in email and token then,

- Fetches the avatars and displays them in a grid layout that can adjust itself according to the device width
- Shows the selected avatar with a border.
- Shows a loading indicator.
- Allows retrying the fetch if an error occurs.
- If the result is empty, shows an appropriate message.

(We don't have a design yet but when we do, we'll just need to update the UI hopefully. For now all I care about is to handle those states.)

- Also, it uses system fonts and colors so it adapts font size changes and light/dark modes.

**These features don't exists yet:**

- Picking a different avatar and displaying a success toast
- Uploading a new avatar

### Testing Steps

Demo app > Avatar Picker View

You can also play with the `#Preview` items under `AvatarPickerView.swift`

You will need to capture an accessToken, I captured mine via changes in this branch `etoledom/avatar-picker`


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+import SwiftUI
+
+@MainActor
+class AvatarPickerViewModel: ObservableObject {
+    private let profileService: ProfileService = .init()
+    private var email: Email?
+    private var authToken: String?
+    @Published private(set) var avatarImageModels: [AvatarImageModel]?
+    @Published private(set) var selectedImageID: String?
+    @Published private(set) var avatarFetchingError: Error?
```
andrewdmontgomery: Looking at the logic in a few places, it looks like we either have models or an error, but not both.  That's often a good time to use an enum.  So I took a crack at it.  See #320

pinarol: I like it. I'll try it out. I might use Swift's predefined `enum Result<Success, Failure>` but other than that i think i'll just merge it.

pinarol: I made this change. You can have another look.

andrewdmontgomery: Oh yes, the `Result<Success, Failure>` approach looks great.

andrewdmontgomery: Ah, and you used that approach on the `currentAvatar` properties, too.  👍

pinarol: Yeap.

etoledom (approved):
Looking good!

I think it would be good for the picker in the demo app to have the same kind of presentation/navigation that we want users to implement.

We can build this on the way so let's continue moving forward 👍

PR #323, opened 2024-07-19
Use Rsync instead of CP for copying generated source files

andrewdmontgomery:
Closes #

### Description

This does a few things:
- ~Reorganize the Docc documentation to make a home for "Models" documentation~
- ~Update the `make generate` command so that it copies the documentation to the proper place~
- Update the `make generate` command to use `rsync` instead of `cp`

#### Rsync instead of CP
If we ever remove (or rename) a model, the `cp` command will copy the new source file.  But it will leave the old source file in place.  This `rsync` command will `--delete` any files in the destination that are not present in the source.

### Testing Steps

- [ ] CI should be 🟢 
- Go to the `openapi-generator/generated/OpenAPIClient/Classes` folder and remove one file:
  - Manually run the `rsync` command from the Makefile (don't run `make generate` since that will replace the file you just deleted)
  - [ ] Verify that the file you deleted above is now also missing from `Sources/Gravatar/OpenAPI/Generated` folder
- [ ] ~Documentation should include a "Models" section~
- [ ] ~The "Articles" section should still be shown at the top of the documentation~


etoledom (commented):
> Reorganize the Docc documentation to make a home for "Models" documentation

This looks interesting 🤔 
Technically, since we commit the model files, we do generate documentation for those types. See [Profile](https://automattic.github.io/Gravatar-SDK-iOS/gravatar/documentation/gravatar/profile) as an example.

About the files generated by openapi, I like the table, but the links seem to be broken. Not sure if they will work once they are published.
The link looks like this `../README.md#documentation-for-models`
We don't have a `README.md` at this level, and docc forces us to make the main `.md` file of the same name than the target 🤔 

My guess is that these generated docs are useful when the generated files are not committed in source control.

> Update the make generate command to use rsync instead of cp

Totally let's do this 🎉

andrewdmontgomery:
> Technically, since we commit the model files, we do generate documentation for those types. See [Profile](https://automattic.github.io/Gravatar-SDK-iOS/gravatar/documentation/gravatar/profile) as an example.

Goodness.  I must have been tired at the end of the week when I went looking for those models.  You are totally right.

> My guess is that these generated docs are useful when the generated files are not committed in source control.

Absolutely.  I'll remove all that extra stuff.

>> Update the make generate command to use rsync instead of cp
>Totally let's do this 🎉

And yep, I'll make this PR just this.

andrewdmontgomery:
@etoledom I changed the name of this PR to reflect the more limited scope.

etoledom (approved):
Amazing! Thank you 🙏

etoledom:
I've just updated from trunk to make this PR mergable

PR #324, opened 2024-07-23
Add `.avatarPickerSheet(...)` modifier

pinarol:
Closes #

### Description

Adds `.avatarPickerSheet(...)` modifier to encapsulate the AvatarPickerView and display it as a modal.

### Testing Steps

Demo App > Avatar Picker View


etoledom (approved):
Looking pretty good!

In the future we can also extend this with other kind of presentations if we choose to 👍

PR #326, opened 2024-07-24
Prep work for adding support to Enterprise distribution via App Center

mokagio:
### Description

Restructures the project settings and adds some Fastlane tooling to enable Enterprise distribution via App Center.

Unfortunately, I had to update the source code to make certain type `public` in order to build the two apps for distribution. See [b6610db](https://github.com/Automattic/Gravatar-SDK-iOS/pull/326/commits/b6610dbf6915780e51fc0a066c1456a0a7447d92) @andrewdmontgomery @pinarol @etoledom I don't pretend to know what's best from an API design point of view in terms of access control etc, so I'll leave it up to you to adapt the code to fit your plans and keep distribution possible. I would strongly encourage against abusing `@testable` to access non `public` types outside of test targets, though.

### Testing Steps

Notice the green CI, which verifies the infra and coding changes have not broken the demo apps in Debug mode.

For Enterprise distribution, try running `make build-demo-for-distribution`.


andrewdmontgomery:
@pinarol @etoledom 

I wonder if we want to rename `Gravatar-Demo`:

- `Swift-Demo`
- `SwiftUI-Demo`

This would involve renaming/updating:

- `Gravatar-Demo` scheme
- `Gravatar-Demo` target
- Update Makefile with new scheme name
- Updating the `Gravatar-Demo` folder name (for these, we could keep the `Gravatar-` prefix... or not.

The reason I dropped `Gravatar` above is that this would also update the `PRODUCT_NAME`.  And as it stands, when the app gets installed, the name is getting truncated: `Gravatar-Sw...`

Here's a draft PR up with these changes, targeting this branch, so you can see what it would look like: #327

etoledom:
I'm OK editing the names.

> * `Swift-Demo`
> * `SwiftUI-Demo`

Since the difference is the framework maybe it could be:
- UIKit-Demo
- SwiftUI-Demo

andrewdmontgomery:
> Since the difference is the framework maybe it could be:
> 
> * UIKit-Demo
> * SwiftUI-Demo

I like that.  I'll update the PR.

andrewdmontgomery:
@mokagio asked something in the PR I made with the rename:

https://github.com/Automattic/Gravatar-SDK-iOS/pull/327#pullrequestreview-2197856842
> Have you considered renaming the folders as well? I think if the rename is dove via Xcode, all the reference in the project file will update automatically.

I did.  In fact, I did rename the `Gravatar-Demo` group/folder to `Gravatar-UIKit-Demo`.  It does work they way you describe.  The only thing that needs to be manually fixed is the `info.plist File` Build Setting for each target (since the folder name is in the path).

As I mentioned above:

> (for these, we could keep the Gravatar- prefix... or not)

I'm good either way.  It's entirely up to us, @etoledom and @pinarol .

andrewdmontgomery:
That PR got merged before we could test it.  So we should test building the demo apps in this PR (with Xcode and `make`).  I tested, obviously.  But someone else should, too.

Also, @pinarol , this got merged before you had a chance to see it. So chime in.  Any concerns, we can revert.

One thing I noticed: I forgot to update the `Makefile` with the new scheme names.  I was about to update my branch when I noticed you merged it.  So I can just push a commit here with that, if that's alright.

mokagio:
> So I can just push a commit here with that, if that's alright.

@andrewdmontgomery all good, and thanks for following up!

pinarol:
> The reason I dropped Gravatar above is that this would also update the PRODUCT_NAME. And as it stands, when the app gets installed, the name is getting truncated: Gravatar-Sw...

> Also, @pinarol , this got merged before you had a chance to see it. So chime in. Any concerns, we can revert.

@andrewdmontgomery 

I am confused because from now on when the app installed there will be no "Gravatar" in the name. Wouldn't it be more confusing for non-developers who will install the app via AppCenter? I would have difficult time finding the app in my real device among many other icons and my first intuition wouldn't be searching via the "Swift" keyword. The name should be something easily findable via spotlight search with "Gravatar" keyword. So I think we should keep "Gravatar" in the name even though it gets a bit truncated from the end or we can come up with sth short but still keep it.

Just updating the "Display Name" field from XCode should be sufficient I think. How does `Gravatar UIKit`, `Gravatar SwiftUI` sound? We don't get truncation this way but even if it gets truncated from the end a bit, this is much more clear for a non-dev:

Screenshot 2024-07-25 at 10 05 10

Also, we don't need to emphasize "Swift" in the name or in the icon. Almost every app is in Swift now, it is not a distinctive thing. We need to emphasize "Gravatar" in the name and the distinctive part is whether it is UIKit or SwiftUI. 

We can maybe emphasize that this is a demo app with a suitable icon like combining the Gravatar icon with a construction item. 🤔

mokagio:
Unfortunately, I had to update the source code to make certain type public in order to build the two apps for distribution. See [b6610db](https://github.com/Automattic/Gravatar-SDK-iOS/pull/326/commits/b6610dbf6915780e51fc0a066c1456a0a7447d92) @andrewdmontgomery @pinarol @etoledom I don't pretend to know what's best from an API design point of view in terms of access control etc, so I'll leave it up to you to adapt the code to fit your plans and keep distribution possible. I would strongly encourage against abusing @testable to access non public types outside of test targets, though.

---

Also, this PR is almost done and I'm keen on your feedback. It's still in draft because I need to verify something about the `MATCH_PASSWORD` env var behavior.

fastlane/lib/code_signing_helpers.rb
```diff
@@ -0,0 +1,32 @@
...
+  MATCH_S3_SECRET_ACCESS_KEY
+  MATCH_PASSWORD
+].freeze
+
+# Required for app_store_connect_api_key to generate the key information to pass down the call chain.
+ASC_API_KEY_ENV_VARS = %w[
+  APP_STORE_CONNECT_API_KEY_KEY_ID
+  APP_STORE_CONNECT_API_KEY_ISSUER_ID
+  APP_STORE_CONNECT_API_KEY_KEY
+].freeze
```
mokagio:
```suggestion
```

fastlane/lib/code_signing_helpers.rb
```diff
@@ -0,0 +1,25 @@
...
+  storage_mode: 's3',
+  s3_bucket: 'a8c-fastlane-match',
+  s3_region: 'us-east-2'
+}.freeze
+
+# Required for sync_code_signing to authenticate with S3.
+CODE_SIGNING_ENV_VARS = %w[
+  MATCH_S3_ACCESS_KEY
+  MATCH_S3_SECRET_ACCESS_KEY
+  MATCH_PASSWORD
```
mokagio: I need to check the behavior without this env var. I expect it to look in the keychain / prompt for it, which I think we Apps Infra discussed elsewhere to be a valid behavior, but I haven't had the time to check.

etoledom:
Here is a fast PR updating the icons of the demo apps:

https://github.com/Automattic/Gravatar-SDK-iOS/pull/328

pinarol:
> Unfortunately, I had to update the source code to make certain type public in order to build the two apps for distribution.

Thanks a lot. I did a rearrangement on the access modifiers since we don't want to make some of them public. https://github.com/Automattic/Gravatar-SDK-iOS/pull/326/commits/2dd26e8de650d221d8767b8db5167d7b33057696

andrewdmontgomery:
@pinarol you make some great points above.  I was thinking of this from within our own context: getting builds to our designer.  But you are right.

I like the `Gravatar-SwfitUI` and `Gravatar-UIKit` names.  I concidered them, too.  The problem I was trying to avoid is that we will likely have a `Gravatar` mobile app.  And this will not be related.

Still, I think either way, there will be some ambiguity, and the names you've suggested deal much better with the existing problem: recognizable demo app names.

> Just updating the "Display Name" field from XCode should be sufficient I think. How does Gravatar UIKit, Gravatar SwiftUI sound? 

I suggest that we update the target and scheme names, too.  Consistency is beneficial here:
- Target name == Scheme name == Product name

I'll put up a quick PR for that

andrewdmontgomery:
> I suggest that we update the target and scheme names, too. Consistency is beneficial here:
>
> Target name == Scheme name == Product name
> I'll put up a quick PR for that

#329 

Take a look, comment, etc.

Note that I split the difference a little.  Having the target name match the product name I think is an intuitive connection: 

- Building the target produces a product of the same name.

The scheme and folder names (and the `.xcconfig` files) can be more verbose, and I think that verbosity is pretty nice: `Gravatar-SwiftUI-Demo` reads nicely.

So we have good consistency, with concise and intuitive target/product names.

At least that's my pitch.  Chime in on that PR with thoughts.

mokagio:
> I did a rearrangement on the access modifiers since we don't want to make some of them public. https://github.com/Automattic/Gravatar-SDK-iOS/commit/2dd26e8de650d221d8767b8db5167d7b33057696

@pinarol thank you for that. I hope my tweaks were decent enough to build on top of without too much hassle.

Demo/Gravatar-Demo.xcodeproj/project.pbxproj
```diff
@@ -267,9 +281,10 @@
 			isa = PBXResourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				49EFFB562C51AA280086589A /* UIKit-Demo.Base.xcconfig in Resources */,
```
AliSoftware: `.xcconfig` files should not be included in the final bundle / built product (same for the other ones I see below, if I'm reading the ~~matrix~~ `.pbxproj` file right)

Demo/Demo/Gravatar-UIKit-Demo/DemoRemoteSVGViewController.swift
```diff
@@ -1,3 +1,5 @@
+#if DEBUG
+
 import UIKit
 @testable import GravatarUI
```
AliSoftware, line 4: Wut?! `@testable import` in a non-test source file? (I'm actually surprised Xcode doesn't even complain about it and would still allow this to compile on a non-test target … 🤔)

fastlane/lib/code_signing_helpers.rb
AliSoftware: > I expect it to look in the keychain / prompt for it

Correct

> which I think we Apps Infra discussed elsewhere to be a valid behavior

Agreed

Demo/Demo/Base.xcconfig
```diff
@@ -0,0 +1,6 @@
+DEVELOPMENT_TEAM = PZYM8XX95Q
+
+// By default use the code sigining identity for development builds that Automatticians have access to.
+// This is better tha using the generic "Apple Development" identity because it prevents Xcode selecting one of the other matching identities in the users' keychain.
```
AliSoftware:
```suggestion
// This is better than using the generic "Apple Development" identity because it prevents Xcode selecting one of the other matching identities in the users' keychain.
```

fastlane/Fastfile
```diff
@@ -20,21 +40,91 @@ platform :ios do
...
+def bundle_id_prototype_build_swiftui!
+  read_from_xcconfig!(key: 'PRODUCT_BUNDLE_IDENTIFIER', xcconfig: XCCONFIG_PROTOTYPE_BUILD_SWIFTUI)
+end
+
+def bundle_id_prototype_build_uikit!
+  read_from_xcconfig!(key: 'PRODUCT_BUNDLE_IDENTIFIER', xcconfig: XCCONFIG_PROTOTYPE_BUILD_UIKIT)
+end
+
+def read_from_xcconfig!(key:, xcconfig:)
+  configs = Xcodeproj::Config.new(xcconfig).attributes
```
AliSoftware, line 126: Side note that this doesn't _resolve_ the values in the `xcconfig` files being read, but just parses and read the values in the `xcconfig` file raw.

To actually _resolve_ the value of a build config key in the most bulletproof way, I think it would require `xcodebuild -showBuildSettings` so it knows the full context (and how each target is set up to use which build config to deduce the right inheritance rules of the various levels at which build configs are applied and resolve them appropriately). Which is probably why _fastlane_ itself calls `xcodebuild -showBuildSetting` at some point instead of using `Xcodeproj::Config` as its implementation details.

If the only problem we're trying to solve/workaround is what you documented in `Base.xcconfig`—i.e. that you have to redeclare `PROVISIONING_PROFILE_SPECIFIER` in each xcconfig file explicitly for this to work—I'm sure there are better workaround than having to shell out and call `xcodebuild -showBuildSettings` though (especially given such command takes some time, as you discovered when you had to adjust the timeout for it).

The workaround you adopted currently (redefining explicitly in each xcconfig) is actually totally acceptable for that. But, while I haven't checked the docs of `Xcodeproj` to check, maybe there is an in-between solution with that gem to at least make it take the `#include` into account when passing the file + recursively resolve interpolations in the values? 🤔

AliSoftware (commented):
_Couldn't test this as I don't have my Mac around, but couldn't help myself give that diff a quick look anyway when I saw the GH notif_ 😅

fastlane/Fastfile
AliSoftware: I wonder if https://rubydoc.info/gems/xcodeproj/Xcodeproj%2FProject%2FObject%2FAbstractTarget:resolved_build_setting can help there 🤔

mokagio: :nerd_sniped:

I had a quick go using `irb` and my preliminary finding is that resolved travels the hierarchy but does not "resolve" the variable usages:

image

_The `true` in the second example is for `resolve_against_xcconfig` which accounts for config files._

But then I looked at the issues and found the closed https://github.com/CocoaPods/Xcodeproj/issues/505 . Looking at the comments, I had a go at changing the syntax in the `xcconfig`:

```diff
- PROVISIONING_PROFILE_SPECIFIER = match InHouse $PRODUCT_BUNDLE_IDENTIFIER
+ PROVISIONING_PROFILE_SPECIFIER = match InHouse $(PRODUCT_BUNDLE_IDENTIFIER)
```

And... it worked 🎉 

image

Thanks @AliSoftware. I'll follow up with a dedicated PR to streamline the `xcconfig`s  based on these findings.

pinarol:
> @pinarol thank you for that. I hope my tweaks were decent enough to build on top of without too much hassle.

No worries at all. I am glad you made the necessary changes instead of getting blocked. 👍

pinarol (approved):
LGTM. 👍

PR #328, opened 2024-07-25
Changing icons of demo app

etoledom:
### Description

Proposing different icons for the demo app:

CleanShot 2024-07-25 at 16 28 47@2x

The idea is that they represent the skeleton of a Gravatar card, using the color schemes of UIKit and SwiftUI.


pinarol (approved):
Looks awesome to me! 🎉 
@andrewdmontgomery WDYT?

andrewdmontgomery:
I love that for SwiftUI, since SwiftUI uses the Blue/Black color palette.  Can we use the Orange/White palette (of Swift) for UIKit?

The main reason for me is here: 
image

I know Gravatar's color is blue.  But the `G` icon within the App Icon conveys "this is Gravatar" very well.  We can use color to differentiate the important difference.

etoledom:
@andrewdmontgomery - @pinarol - Updated!

CleanShot 2024-07-26 at 09 36 25@2x

pinarol (approved):
Even more awesomeness 🎉

PR #329, opened 2024-07-25
Update target/scheme names

andrewdmontgomery:
Closes #

### Description

This updates the **target names** and the **scheme names**:
- Update **target names** so that they reflect the **product names** they will be used to generate
- Update the **scheme names** to match
- Update `.xcconfig` file names to match the scheme/folder names

Since the name for schemes and folders can be more verbose, I've chosen to keep the `Gravtar-` prefix.  This has the added benefit of matching the existing folder names for each demo app.

image

Screenshot 2024-07-25 at 11 06 37 AM

image

image

### Testing Steps

- [ ] Build (and run) demos in Xcode
- [ ] Build demos from command line `make build-demo` (this will build both)


fastlane/Fastfile
```diff
@@ -39,8 +39,8 @@ platform :ios do
     )
   end
 
-  desc 'Builds the Gravatar-Demo project with a specific scheme'
-  lane :build_demo do |scheme: 'Gravatar-Demo'|
+  desc 'Builds the Gravatar-UIKit-Demo project with a specific scheme'
```
mokagio: The project remained called `Gravatar-Demo`, though

```suggestion
  desc 'Builds the Gravatar-Demo project with a specific scheme'
```

mokagio (approved):
Does what it says in the title 👍 

Looks good to me technically apart for the minor fix in the `Fastfile` comment (too much find and replace? 😄 )

I'll leave it up to the dev team members to give another approval and merge

pinarol (approved):
Thanks a lot!

one tiny nit: I think the display name looks better without the `-`

Screenshot 2024-07-26 at 09 45 50

compared to:

Screenshot 2024-07-26 at 09 46 18

We can remove the `-` by:

Screenshot 2024-07-26 at 09 48 37

But this is just a `nit` and a personal opinion and I am fine both ways. 

-----

@etoledom Can you also check?

etoledom:
> 
> @etoledom Can you also check?

I agree. `Display Name` only affects the name under the icon, right? It would be nice to not have a dash there 🙏

pinarol:
> Display Name only affects the name under the icon, right?

Yep.

andrewdmontgomery:
Yeah, I like it with the space, too.

And there's no reason a target can't have a space.

And having the target name be slightly different from the product name adds unnecessary complication.

I'll update the target names to replace the `-` with a space.

andrewdmontgomery:
I resolved the project file conflicts from the target branch, and then made the change.  Looks excellent.

| iPhone SE (3rd Generation) | iPhone 15 Pro Max |
| --- | --- |
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-07-26 at 10 57 35 | Simulator Screenshot - iPhone 15 Pro Max - 2024-07-26 at 10 57 38 |

PR #330, opened 2024-07-26
Build and upload prototype build from CI

mokagio:
### Description

Adds the automation necessary to build a prototype builds in CI and upload it to App Center

### Testing Steps

Check out the latest Buildkite build and the bot comments in this PR

image

image

Notice that the icon is the old one because this branched from #326 before #329 landed.


.buildkite/pipeline.yml
```diff
@@ -60,5 +60,31 @@ steps:
...
+
+      - label: ":arrow_up: Upload SwiftUI demo to App Center"
+        depends_on: build_swiftui
+        plugins: [$CI_TOOLKIT]
+        command: .buildkite/commands/upload-to-appcenter.sh SwiftUI
+
+      - label: ":arrow_up: Upload UIKit demo to App Center"
+        depends_on: build_uikit
+        plugins: [$CI_TOOLKIT]
+        command: .buildkite/commands/upload-to-appcenter.sh UIKit
```
mokagio: This could be DRYed using a Buildkite matrix. Leaving it for a followup.

.buildkite/pipeline.yml
```diff
@@ -60,5 +60,31 @@ steps:
...
+        depends_on: test
+        command: BUILD_NUMBER=$BUILDKITE_BUILD_NUMBER make build-demo-for-distribution-uikit
+        artifacts:
+          - ".build/artifacts/*.ipa"
+          - ".build/artifacts/*.dSYM.zip"
+
+      - label: ":arrow_up: Upload SwiftUI demo to App Center"
+        depends_on: build_swiftui
+        plugins: [$CI_TOOLKIT]
+        command: .buildkite/commands/upload-to-appcenter.sh SwiftUI
```
mokagio: The upload step could run in a simpler machine instead of the current macOS one.

This has been discussed elsewhere and at some point I / someone will create an agent with the required capability.

For the time being, we can spare the extra 60s x 2 on the Mac queue...

fastlane/Fastfile
```diff
@@ -110,6 +143,63 @@ def configure_code_signing(bundle_ids:, readonly:)
   )
 end
 
+def upload_to_appcenter(appcenter_slug:, app_name:)
+  EnvManager.require_env_vars!('APPCENTER_API_TOKEN')
```
spencertransier, line 147: I'm familiar with the idea of `EnvManager` from your RFC, but where does `EnvManager` get defined in this project? I looked for it but feel like I missed somewhere obvious 😄

fastlane/Fastfile
```diff
@@ -50,22 +64,41 @@ platform :ios do
...
+  lane :upload_demo_to_appcenter do |name:, build_number:|
+    UI.user_error!("Unsupported name #{name}") unless %w[SwiftUI UIKit].include? name
+
+    slug = "Gravatar-SDK-Demo-#{name}"
+    upload_to_appcenter(appcenter_slug: slug, app_name: "#{name}-Demo")
+
+    next unless is_ci
+
+    annotate_pr_with_appcenter_link(app_suffix: name)
+    annotate_buildkite_with_appcenter_link(app_suffix: name, appcenter_slug: slug, build_number: build_number)
```
spencertransier, line 99: Nit, but I was also wondering what your thoughts are about breaking these out onto multiple lines versus having them on a single line. I like breaking them out for readability when there is more than one argument being passed. What's your opinion on it? 

```suggestion
    annotate_buildkite_with_appcenter_link(
      app_suffix: name,
      appcenter_slug: slug,
      build_number: build_number
    )
```

fastlane/Fastfile
```diff
@@ -50,22 +64,41 @@ platform :ios do
...
+      }
     )
   end
 
+  desc 'Upload the demo app with the given name to App Center'
+  lane :upload_demo_to_appcenter do |name:, build_number:|
+    UI.user_error!("Unsupported name #{name}") unless %w[SwiftUI UIKit].include? name
+
+    slug = "Gravatar-SDK-Demo-#{name}"
+    upload_to_appcenter(appcenter_slug: slug, app_name: "#{name}-Demo")
```
spencertransier: Same question as https://github.com/Automattic/Gravatar-SDK-iOS/pull/330/files#r1693772678 (sorry, out of order)

```suggestion
    upload_to_appcenter(
      appcenter_slug: slug,
      app_name: "#{name}-Demo"
    )
```

fastlane/Fastfile
mokagio: It was introduced in the PR this one was based off:

- Source: https://github.com/Automattic/Gravatar-SDK-iOS/pull/326/files#diff-50ef2286ed07d63e7d519e2d961dec131f1387d69389067a06cc7b728d15217f
- Import in `Fastfile`: https://github.com/Automattic/Gravatar-SDK-iOS/pull/326/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R14

I have also have a WIP PR to move it in `release-toolkit`: https://github.com/wordpress-mobile/release-toolkit/pull/578/files

.buildkite/pipeline.yml
AliSoftware: Actually I'm not sure it can, because the `depends_on` attribute can't use `{{matrix}}` iinm. Which is what blocked us from using a Buildkite matrix in a similar use case I discussed with the Beeper-Desktop team (ref: p1720535625287679/1720533115.283839-slack-CC7L49W13) 🙃

Makefile
```diff
@@ -50,11 +50,21 @@ build-demo-swiftui: bundle-install
...
+
+build-demo-for-distribution-uikit: fetch-code-signing check-build-number
+	bundle exec fastlane build_demo_for_distribution \
+		scheme:$(SCHEME_DEMO_UIKIT) \
+		build_number:$(BUILD_NUMBER)
+
+check-build-number:
+ifndef BUILD_NUMBER
+	@echo "BUILD_NUMBER not set in the environment. Will default to 0."
+	override BUILD_NUMBER = 0
```
AliSoftware, line 66: TIL about this `override` syntax/directive in `Makefile` 👍

fastlane/Fastfile
```diff
@@ -110,6 +143,63 @@ def configure_code_signing(bundle_ids:, readonly:)
...
+    build_number: build_number
+  }
+  appcenter_install_url = "https://install.appcenter.ms/orgs/#{APPCENTER_OWNER_NAME}/apps/#{appcenter_slug}/releases/#{appcenter_id}"
+  list = metadata.map { |k, v| " - **#{k}**: #{v}" }.join("\n")
+
+  buildkite_annotate(
+    context: "appcenter-info-#{appcenter_slug}",
+    style: 'info',
+    message: "Gravatar #{app_suffix} Demo [App Center Build](#{appcenter_install_url}) Info:\n\n#{list}"
+  )
```
AliSoftware, line 200: 💡  I think it'd be quite useful to also call `buildkite_metadata(set: metadata)` as well, in addition to `buildkite_annotate`. In fact the whole reason why other projects define `metadata = { … }` as a Hash (only to `metadata.map { … }.join("\n")` that hash when constructing the bullet-point `list` to include in the annotation—as opposed to define `list` directly as a heredoc text) is because we use that `metadata` to set the various Build metadata as well. Which can then be useful to do fancy stuff like filtering the builds on Buildkite by metadata, or using that metadata in API calls for various use cases, etc.

```suggestion
  metadata = {
    build_type: 'Prototype',
    'appcenter:id': appcenter_id,
    build_number: build_number
  }
  buildkite_metadata(set: metadata)
  
  appcenter_install_url = "https://install.appcenter.ms/orgs/#{APPCENTER_OWNER_NAME}/apps/#{appcenter_slug}/releases/#{appcenter_id}"
  list = metadata.map { |k, v| " - **#{k}**: #{v}" }.join("\n")
  buildkite_annotate(
    context: "appcenter-info-#{appcenter_slug}",
    style: 'info',
    message: "Gravatar #{app_suffix} Demo [App Center Build](#{appcenter_install_url}) Info:\n\n#{list}"
  )
```

AliSoftware (approved):

Makefile
andrewdmontgomery: whoa, same

.buildkite/pipeline.yml
```diff
@@ -60,5 +60,31 @@ steps:
...
     if: build.tag != null
-    agents:
-      queue: "mac"
+
+  - group: ":appcenter: Prototype Builds"
+    steps:
+      - label: ":package: Build SwiftUI demo"
+        key: build_swiftui
+        depends_on: test
+        command: BUILD_NUMBER=$BUILDKITE_BUILD_NUMBER make build-demo-for-distribution-swiftui
```
andrewdmontgomery, line 69: Using the `BUILDKITE_BUILD_NUMBER` for these builds seems elegant.  Nothing to track, always increments, never reused.

andrewdmontgomery (approved):
This approach looks great from our side.  The automation is easy to understand and the output (in the PR) is just what we wanted.

It looks like you have some issues to work out, still.  But unless something changes, I can approve this approach and leave the implementation details to you folks.

andrewdmontgomery:
❓ Question

Is it possible to have these prototype builds only happen when the PR has a specific label?

It occurred to me this morning that, since that this repository is public, and these prototype builds are only available to Automatticians, their presence in PR's could be mildly confusing.  There are a few ways we could handle that confusion (and I'm a big fan of a layered approach to this, delivering both of these if we can):

- Explain it in the copy of the automated PR comment
- Only build prototypes when we need them

That second point is the one I'm wondering about.  If we can limit prototype builds to just the PR we label `prototype build` (for example), then:
- 95% of our PR builds won't produce a prototype build (no confusion)
- And the rest could have copy that briefly describes something like "Automattic internal use only"

If, on the other hand, those PR labels aren't available to BuildKite, then I think we keep them on every build and just focus on documenting their limited availability.  I believe the ease-of-use of having these builds available to non-developers (without any extra steps) outweighs any worry about public confusion over them.

andrewdmontgomery:
@mokagio I should clarify.  Don't let this question (above) block this PR.  We are happy to iterate, and we're eager to get these prototype builds into the hands of our designer.

AliSoftware:
> Is it possible to have these prototype builds only happen when the PR has a specific label?

Using GitHub labels as conditionals for build jobs is a tricky thing and not ideal (at least in Buildkite), because:
 - If the CI build has already started by the time you decide to add the label to your GitHub PR, that won't suddenly be detected and trigger the job after the fact. So it'd only work effectively if you happen to add your label while you write you PR description but before you hit the button to create the PR (and thus trigger the build); or if you push a new commit (thus triggering a new CI build for it) after you added the label to the PR.
 - Indeed, Buildkite doesn't listen to label-added/label-removed webhook events by default, so it can't react to the event of adding/removing a label to a PR to suddenly decide to retry/re-trigger a job
 - In practice there _is_ an option in Buildkite to trigger a new build on the label-added/label-removed webhook events; but that have the effect of triggering a new _build_ from scratch upon reception of said event, similar to when you push a new commit. Which means that:
    - It will rebuild _all the jobs_ of the build (including lint, build-and-test, etc) not just the job doing the Prototype Build
    - and it will do that regardless of the label that was added/removed (so if you add a label like `tooling` or `bug` to your PR, it will trigger a new build only to realize that's a label we don't care about in the context of doing a Prototype Build or not

So TL;DR: using PR labels for conditionals on CI jobs has a lot of side effects and edge cases to consider and is thus usually not the best solution for this kind of need.

---

What we could do instead if we wanted to only trigger Prototype Builds on certain PRs but not by default, is to make the condition be on the PR title, or on the branch name. For example, [at Beeper-Desktop, they decided to only generate testable builds if the branch name starts with `build.*`](https://github.com/beeper/beeper-desktop-new/blob/main/.buildkite/pipeline.yml#L96).

Relying on the PR title would still have the same quirk as with labels wrt to live-detecting a change though: if you open your PR without the magic word in its title in the first place, editing the PR's title after the fact won't suddenly make Buildkite detect it nor trigger a Prototype build automatically: you'd still have to push a new commit after changing the PR title (or use the Rebuild button on the last build on Buildkite' UI) for it to take the change into account. But arguably changing the title of a PR is less common than adding a label. Still a tradeoff though.

---

In the end I think building the Prototype Build unconditionally on all PRs is OK; besides, we've configured the PR comment to be folded by default (i.e. the QR code is hidden within a `<details>…</details>` block) so hopefully that doesn't create _too much_ noise?

AliSoftware (commented):
@mokagio I think this can be addressed in a follow-up PR to not block this one and allow to move things forward… but I noticed that we now build the Demo projects twice:

 - Once via the "Build Demo Project" job (which ends up calling `fastlane build_demo scheme:Gravatar-UIKit-Demo` and `fastlane build_demo scheme:Gravatar-SwiftUI-Demo`)
 - Once to generate the Prototype Builds (separate "Build UIKit demo" and "Build SwiftUI demo" jobs)

image

WDYT about completely removing the "Build Demo Projects" job from the `pipeline.yml` (which means we could probably also delete the `build_demo` lane to clean things up, only keeping the `build_demo_for_distribution` lane used for the Prototype Builds) at some point?

Or am I missing something and the "Build Demo Projects" does something that the "Build * Demo" jobs don't (I thought about one running the unit tests and the others not… but in practice and as you mention in the comment you left in the `build_demo` lane, there is no Unit Tests on the demo projects, so that's not it)

AliSoftware:
Hi @pinarol @andrewdmontgomery !
I took the liberty to update the PR to fix the lastest quirks I saw in CI (typo on `Gavatar-SwiftUI-Demo.Release.xcconfig` file name, removed `.xcconfig` file from Build Phase, fix incorrect artifact basename used by `appcenter_upload` when referencing `.ipa` and `.dSYM.zip`) to make the Prototype Builds job finally succeed.

Can you do a last confidence check by validating that the QR code on the two comments above ([here](https://github.com/Automattic/Gravatar-SDK-iOS/pull/330#issuecomment-2253600847) and [here](https://github.com/Automattic/Gravatar-SDK-iOS/pull/330#issuecomment-2253600949)) work, that you're able to install the Prototype Builds on your device, and that they work as expected?

If so I think we can finally merge this PR (we can always address potential adjustments in follow-up PRs if needs be anyway)

pinarol:
I have installed both builds via AppCenter successfully 🎉 :shipit:

pinarol:
Thank you very much! I'll merge.

AliSoftware:
> WDYT about completely removing the "Build Demo Projects" job from the `pipeline.yml` […] at some point?

Followed-up on this in https://github.com/Automattic/Gravatar-SDK-iOS/pull/348

> Is it possible to have these prototype builds only happen when the PR has a specific label?

PS: @andrewdmontgomery note that with https://github.com/Automattic/Gravatar-SDK-iOS/pull/348 removing the "Build Demo Project" job—in favor of the compilation of those Demo projects being validated by CI by the fact that the Prototype Builds for those Demos are being build and would surface any compilation issue on them there anyway—it now makes even more sense to build those Prototype Builds _unconditionally_ (and avoid skipping them based on labels or whatnot, otherwise we'd also not validate correct compilation of the demos) 🙃

mokagio:
Thanks for bringing this home @AliSoftware 👏 

While it's a team workflow decision, my two cents on the Prototype Builds is that it's worth building them unconditionally, also. CI Compute and storage are cheap compared to the value of the feedback we get from those builds. And notice that the feedback occurs not only when someone installs the build but also when the build builds.

PR #331, opened 2024-07-26
Put a profile card into the avatar picker

pinarol:
Closes #

### Description

Adding a profile card into the avatar picker. For now I added to the top because it is easier to test this way. When we have the designs we can replace it easily.

The profile card has ghosting view too. `ProfileViewPlaceholderColorManager` manages it so the same effect can be applied to another place easily.

### Testing Steps

Play around with the SwifttUI demo app.


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -163,15 +188,50 @@ struct AvatarPickerView: View {
...
 }
+
+private struct PreviewModel: ProfileSummaryModel {
+    var avatarIdentifier: Gravatar.AvatarIdentifier? {
+        .email("xxx@gmail.com")
+    }
+
+    var displayName: String {
+        "Shelly Kimbrough"
+    }
```
etoledom: Maybe it would be good to move this struct to inside the `#Preview` block, since this code is removed from production build.

pinarol: Right, done. 👍

pinarol:
Ready for another look @etoledom

etoledom (approved):
Looking good!
🎉

PR #334, opened 2024-07-26
Improve documentation organization

andrewdmontgomery:
Closes #

### Description

Adds some organization to the `Gravatar` module's documentation.

### DIFF
--- a/Sources/GravatarUI/Exports.swift
+++ b/Sources/GravatarUI/Exports.swift
@@ -1,1 +1,0 @@
-@_exported import Gravatar
--- a/.github/workflows/docc.yml
+++ b/.github/workflows/docc.yml
@@ -40,7 +40,7 @@ jobs:
             transform-for-static-hosting /tmp/docbuild/Build/Products/Debug-iphoneos/Gravatar.doccarchive \
             --hosting-base-path Gravatar-SDK-iOS/$GRAVATAR_URL \
             --output-path gravatar-docs;
-          echo "<script>window.location.href += \"$GRAVATAR_URL/documentation/gravatar\"</script>" > gravatar-docs/index.html;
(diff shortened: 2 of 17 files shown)


andrewdmontgomery:
I saw that we were exporting `Gravatar` from `GravatarUI`, and this was allowing us to link symbol documentation in `GravatarUI` to various symbols in `Gravatar`.

But the downside of that approach is that the `GravatarUI` documentation then contains all of the `Gravatar` symbols, making it much harder to find and understand the symbols from `GravatarUI`.

So I've removed that export, and made adjustments to a few documentation entries in `GravatarUI` that were linking to `Gravatar` symbols (which won't work without exporting them in `GravatarUI`).

It's a trade-off:
- Links work, but `GravatarUI` and `Gravatar` documentation are both included – and completely unorganized – in the `GravatarUI` documentation
- Keep the documentation of the modules separate, but lose linking to `Gravatar` symbols from `GravatarUI` documentation.

There's another challenge with having all documentation in the `GravatarUI` module.  The more formatting, images, etc. that we add to `Gravatar` documentation, the further it will drift from whatever gets automatically generated in the `GravatarUI` module.  Put another way: if it is worth adding formating to the `Gravatar` module documentation, it is also worth adding that same formatting _again_ to the `Gravatar` module documentation that is included with the `GravatarUI` module.  That's twice the work.  We could potentially automate some of it.

Sources/GravatarUI/Exports.swift
```diff
@@ -1 +0,0 @@
-@_exported import Gravatar
```
etoledom, line 1: The reason we did this was to be able to `import GravatarUI` in a project, and have all the types from Gravatar already imported also, without the need to `import Gravatar` all the time.

How this influence the docs was unintended.

etoledom: Now for example, the demo apps won't build without adding also `import Gravatar`.

PR #335, opened 2024-07-29
Adjust `TestHelper` access, drop use of `@testable import`

andrewdmontgomery:
Closes #

### Description

Since the `TestHelper` target is not listed as a Product in the `Package.swift`, it is not available for use by third parties who integrate our SDK.

Instead of using `@testable`, let's set the access levels in the `TestHelper` module such that they reflect our usage of the module from the `GravatarTests` and `GravatarUITests` modules.

- Stops using`@testable import` imports when importning `TestHelper` in the `GravatarTests` and `GravatarUITests` unit tests.
- Makes the bare minimum access changes within the `TestHelper` module to accommodate this change
- Uses `package` access

### Testing Steps

- [ ] All `GravatarTests` should pass locally
- [ ] All `GravatarUITests` should pass locally
- [ ] `make validate-pod` should pass locally (note: there is one warning about deprecation, which CI ignores)
- [ ] You should _not_ be able to `import TestHelper` from the Demo App
- [ ] CI should be 🟢

### DIFF
--- a/Sources/TestHelpers/HTTPClientMock.swift
+++ b/Sources/TestHelpers/HTTPClientMock.swift
@@ -1,19 +1,19 @@
 import Foundation
 import Gravatar
 
-struct HTTPClientMock: HTTPClient {
+package struct HTTPClientMock: HTTPClient {
     private let session: URLSessionMock
 
-    init(session: URLSessionMock) {
+    package init(session: URLSessionMock) {
         self.session = session
     }
 
-    func fetchData(with request: URLRequest) async throws -> (Data, HTTPURLResponse) {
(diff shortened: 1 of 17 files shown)


andrewdmontgomery:
I wasn't here when this was created.  So I don't know the history.  I'll bet there was a reason to `@testable import` for this originally.

This seems to work well now, though.  Is there a reason to continue to use the `@testable import` still?

pinarol (approved):
Nice!

pinarol:
> Is there a reason to continue to use the @testable import still?

Nope I don't think there is.

PR #336, opened 2024-08-01
Implementing v3 version of image upload endpoint

etoledom:
### Description

This PR implements `POST api.gravatar.com/v3/me/avatars`.

One big difference between v1 and v3 is that v1 was able to square an image if it wasn't square already. V3 in contrast won't square the image but expect it to be square or return an error otherwise.

The Native image picker on iOS won't always crop a perfect square:
- Sometimes will be off by 1 or 2 pixels
- It won't square an image in landscape if its completely zoomed out, which makes it not ideal.

On this PR, we solve the problem in the demo app side, by rendering a square image and embedding the original photo in it, which is what the v1 endpoint was doing on the backend.

For example:

On the image picker, we can have this logic internally, or build our own cropper without these issues.

### Testing Steps

- Upload an image on the UIKit demo app.

### DIFF
--- a/Sources/Gravatar/Network/Services/ImageUploadService.swift
+++ b/Sources/Gravatar/Network/Services/ImageUploadService.swift
@@ -13,7 +13,7 @@ struct ImageUploadService: ImageUploader {
     }
 
     @discardableResult
-    func uploadImage(_ image: UIImage, accessToken: String, additionalHTTPHeaders: [HTTPHeaderField]?) async throws -> URLResponse {
+    func uploadImage(_ image: UIImage, accessToken: String, additionalHTTPHeaders: [HTTPHeaderField]?) async throws -> (data: Data, response: HTTPURLResponse) {
         guard let data = image.jpegData(compressionQuality: 0.9) else {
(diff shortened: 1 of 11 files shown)


Sources/Gravatar/Network/Services/ImageUploadService.swift
```diff
@@ -38,33 +38,26 @@ struct ImageUploadService: ImageUploader {
...
+    body.append("Content-Disposition: form-data; name=\(UploadParameters.imageKey); filename=\(UploadParameters.filename)\r\n")
     body.append("Content-Type: \(UploadParameters.contentType);\r\n\r\n")
     body.append(imageData)
     body.append("\r\n")
-
-    // Account Payload
-    body.append("--\(boundary)\r\n")
-    body.append("Content-Disposition: form-data; name=\"\(UploadParameters.accountKey)\"\r\n\r\n")
-    body.append("\(account)\r\n")
-
```
etoledom, line 63: Account information is not needed, so there's no need to require an email parameter.

pinarol (approved):
LGTM 👍

PR #337, opened 2024-08-06
SwiftUI Picker - System photos picker

etoledom:
### Description

This PR implements the system photos picker for the Gravatar picker in SwiftUI.

**Notes:** 
- As a demo, I have implemented both add image buttons: in-grid and bottom CTA.
- We are only using the UIKit photos picker, and not the new SwiftUI PhotosPicker, since the later only supports Photo Library, doesn't support cropping, and the selection seems bugged (made for multi-select only, though it can return only one).
- The interface of the `ImagePickerView` is similar to the one of SwiftUI's PhotosPicker, to facilitate a replacement in the future.
- After the selected image is uploaded to the backend, we add the local image to the cache, to avoid an unnecessary download.

**Not included on this PR**:
- Selector between Camera and Library.
- Possibility to inject custom images picker / cropper. (do we want this in the first release?)

### Testing Steps

- SwiftUI demo app.


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -182,6 +183,69 @@ struct AvatarPickerView: View {
...
+            .background(
+                RoundedRectangle(cornerRadius: Constants.avatarCornerRadius)
+                    .stroke(Color(UIColor.systemGray2), style: StrokeStyle(lineWidth: 2, dash: [7]))
+            )
+            .tint(Color(UIColor.systemGray2))
+    }
+}
+
+extension Color {
+    static let gravatarBlue = Color(uiColor: UIColor(red: 0.11, green: 0.31, blue: 0.77, alpha: 1.00))
```
pinarol: This color seems like an important color. Let's either put this into the existing UIColor+DesignSystem extension or define a new one for swiftui like Color+DesignSystem put there.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -93,6 +95,36 @@ class AvatarPickerViewModel: ObservableObject {
...
+
+        let service = AvatarService()
+        do {
+            let avatar = try await service.upload(image, accessToken: authToken)
+            await ImageCache.shared.setEntry(.ready(squareImage), for: avatar.url)
+
+            let newModel = AvatarImageModel(id: avatar.id, source: .remote(url: avatar.url))
+            add(newModel, replacing: localID)
+        } catch {
+            avatarsResult = .failure(error)
```
pinarol: Let's not use avatarsResult for upload result. We would want to be able to distinguish the failure of avatar fetching from failure of avatar upload.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -3,45 +3,29 @@ import SwiftUI
...
         _model = StateObject(wrappedValue: model)
     }
 
     public var body: some View {
-        ScrollView {
-            header()
-            errorMessages()
-
-            profileView()
+        VStack {
```
pinarol, line 34: Just curious. Why was the extra layer of VStack necessary?

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
pinarol: And I realized we don't have any error state in the designs for when the avatar upload fails. I think it is missing atm. But still let's have a different result object for upload.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -93,6 +95,36 @@ class AvatarPickerViewModel: ObservableObject {
...
+        let squareImage = image.squared()
+        let localID = UUID().uuidString
+
+        let localImageModel = AvatarImageModel(id: localID, source: .local(image: image), isLoading: true)
+        add(localImageModel)
+
+        let service = AvatarService()
+        do {
+            let avatar = try await service.upload(image, accessToken: authToken)
+            await ImageCache.shared.setEntry(.ready(squareImage), for: avatar.url)
```
pinarol, line 114: Do we ever fetch this avatar with the same `avatar.url` again? I feel like next time we try to fetch the avatar, the url will be different because of other query params like size and stuff. So this entry won't be used, no?

pinarol (commented):
Looks nice overall but I failed to make a successful upload, getting an error from server. Trying to figure out why.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
etoledom: To keep the CTA button pinned at the bottom, outside the scroll view.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
etoledom: Since we use the same `Avatar` model, the `avatar.url` is the same than the remote avatars used in the grid. It already includes the base url and size param. This is why I decided to use the same model (plus the image structure returned by the upload endpoint is the same than the one returned by the avatar list endpoint)

We will see how this will be handled in the open-api spec, but I'd expect it to be the same structure reused.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
etoledom: True. Since we don't have yet a proper error handling (at the time I was writing this PR), I didn't put much time on this. 

I'd work on this proper error handling on a separate PR once we have the design.

etoledom: We will probably have a per-avatar error state to show in the grid, but let's wait for the proper design before implementing something like this.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -182,6 +183,69 @@ struct AvatarPickerView: View {
...
+struct UploadImageCTAButtonView: View {
+    public var body: some View {
+        Text("Upload image")
+            .font(.callout).fontWeight(.bold)
+            .frame(maxWidth: .infinity)
+            .foregroundColor(.white)
+            .padding()
+            .background(RoundedRectangle(cornerRadius: 4).fill(Color.gravatarBlue))
+    }
+}
```
etoledom: I'm not sure if there's a benefit from having these views as separated structs, or having them as view builder functions.

I had these `UploadImageCTAButtonView`, `ActivityIndicatorView`, `PlusButtonView`, as functions before, but then I decided to make separated structs to make the main `AvatarPickerView` smaller.

Maybe one benefit could be that we could have these structs in a separate file with their own previews, and/or snapshot tests.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -182,6 +183,69 @@ struct AvatarPickerView: View {
...
+            .tint(Color(UIColor.systemGray2))
+    }
+}
+
+extension Color {
+    static let gravatarBlue = Color(uiColor: UIColor(red: 0.11, green: 0.31, blue: 0.77, alpha: 1.00))
+}
+
+private enum Constants {
+    static let horizontalPadding: CGFloat = .DS.Padding.double
```
etoledom: I don't particularly like having these constants declared globally, even though is private (I prefer them to be constrained inside the struct) but it was necessary to share with the other new view structs.

Any special thought?

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
etoledom: Yup sounds good!

I can do this now or we do this when we implement dark mode (if it needs special handling with custom colors).

pinarol: Let's do this now. Let's try to handle dark mode as we go. We can use this color for both light/dark modes I think. It didn't look weird when I tried.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
pinarol: The only benefit would be to easily reuse them from other views, currently we don't have such other views so the benefit isn't here yet. But I don't mind having them as separate views. Also, why don't we put them in separate files also?

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
pinarol: Ideally, if we are going to have separate reusable views, they shouldn't use the constants that belong to AvatarPickerView. We should pass these numbers as input parameters to them and put Constants back into AvatarPickerView imho because those Constants only mean something in that specific context. WDYT?

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
pinarol: okay then let's at least remove that line add a TODO since it is almost certain that we'll have a quite different error handling for avatar upload.

etoledom:
@pinarol - Ready for another look 👍 

- Moved the color to the design system declaration
  - The designs has the same blue for both light and dark modes.
- Moved the new struct views to their own files with previews.
- The `CTAButtonView` has variable title.
- Constants are back to be bounded to AvatarPickerView.
- Removed the wrong error handling from avatar upload.
- Few other small things.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -112,13 +126,17 @@ struct AvatarPickerView: View {
...
+
             ForEach(avatarImageModels) { avatar in
                 AvatarView(
                     url: avatar.url,
-                    placeholder: nil,
+                    placeholder: avatar.localImage,
                     loadingView: {
-                        ProgressView()
-                            .progressViewStyle(CircularProgressViewStyle())
+                        OverlayActivityIndicatorView()
```
pinarol: We don't need the overlay part for the downloading state we just need the progress view part. The overlay is messing up with the background color. Can we keep the plain old ProgressView here?

pinarol (approved):
Looking good 🎉 Just one more tiny comment but I don't need to check again I think.

PR #338, opened 2024-08-06
Avatar picker empty views

pinarol:
Closes #

### Description

Adds the subviews inside the borders to handle empty avatar state and the error state.

Screenshot 2024-08-07 at 15 27 44

Screenshot 2024-08-06 at 11 33 08

Screenshot 2024-08-07 at 15 58 42

### Testing Steps

- Use the previews "Empty elements" and "Load from network"


pinarol:
I'll wait for https://github.com/Automattic/Gravatar-SDK-iOS/pull/337 to be merged first. We'll have some conflicts otherwise.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -31,90 +25,114 @@ struct AvatarPickerView: View {
...
-                } else if model.isAvatarsLoading {
-                    avatarsLoadingView()
+            if case .success(let avatarImageModels) = model.avatarsResult,
+               !avatarImageModels.isEmpty
+            {
+                header()
+                avatarGrid(with: avatarImageModels)
+                imagePicker {
+                    CTAButtonView("Upload image")
                 }
```
etoledom: The button now is scrolling out of view together with the grid.

I guess we don't have a good definition on the design of how the elements should scroll, so probably no need to be too strict about this right now.
I'd say for example that the avatar card shouldn't scroll out either.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -31,90 +25,114 @@ struct AvatarPickerView: View {
...
-
-                Spacer()
-                Text("Try Again")
-                    .font(.subheadline)
-            }
-        })
+    private func contentLoadingErrorView(
+        title: String,
+        subtext: String,
+        image: Image?,
```
etoledom: This image could default to `nil` to avoid sending nil from the init call.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -31,90 +25,114 @@ struct AvatarPickerView: View {
...
-                errorText("You don't have any avatars yet. Why not start uploading some now?")
+                contentLoadingErrorView(
+                    title: "Let's setup your avatar",
+                    subtext: "Choose or upload your favorite avatar images and connect them to your email address.",
+                    image: Image("setup-avatar-emoji", bundle: .module), actionButton: {
+                        imagePicker {
+                            CTAButtonView("Upload image")
+                        }
+                    }
+                )
```
etoledom: `actionButton: ` should probably go on a new line, right? 🤔 

Interesting that the swift formatter didn't catch this.

etoledom (commented):
Looking great!

I have two questions:

I get this error message even when the token is not expired, but just wrong.
I'm not sure if we can differentiate these two scenarios, can we?

If we don't I guess is fine since this is the scenario we want to cover more carefully.




----

I've been trying to figure out from where the different subtitle text color come from:



```swift
Text(title)
    .font(.title2)
    .fontWeight(.semibold)
    .foregroundColor(Color(UIColor.label))
    .padding(.init(top: 0, leading: 0, bottom: .DS.Padding.half, trailing: 0))
Text(subtext)
    .font(.subheadline)
```

`Text(subtext)` doesn't specify the color, so I'd spect it to be black/white as is shown on the `ContentLoadingErrorView` preview. But it's grey on the `AvatarPickerView` preview and runtime 🤔 

I'm very curious 😅

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
pinarol: I'll revert this change. Not directly related with this PR anyway.

etoledom (approved):
The code change proposals won't need a new review (feel free to change or keep what you think is best)

And the questions are not blocking, so I'd say let's merge when ready 👍

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
pinarol: Yep.

pinarol:
> I get this error message even when the token is not expired, but just wrong.
> I'm not sure if we can differentiate these two scenarios, can we?

It depends on the backend but usually we get 401 on both cases. We should still test these in real life once Gravatar OAUTH is ready.

etoledom:
> It depends on the backend but usually we get 401 on both cases. We should still test these in real life once Gravatar OAUTH is ready.

Okay, let's keep it this way for now 👍

pinarol:
> Text(subtext) doesn't specify the color, so I'd spect it to be black/white as is shown on the ContentLoadingErrorView preview. But it's grey on the AvatarPickerView preview and runtime 🤔

Interesting inconsistency. Should be solved when I set the color explicitly. I just forgot to do that.

PR #340, opened 2024-08-07
Fix the issue about `@Environment` var being accessed from outside a View

pinarol:
Closes #

### Description

Fixes the issue:

> Accessing Environment<ColorScheme>'s value outside of being installed on a View. This will always read the default value and will not update.

So we shouldn't use `@Environment` var inside a non-View struct. We need to find a way to listen to color scheme changes and then re-render `AvatarPickerProfileView` so it updates its colors. I am not sure if this is the best solution so let's discuss other ideas if you have any.

### Testing Steps

Check `AvatarPickerProfileView` previews "Empty", "Empty & Loading" in both light & dark appearance. Observe: Colors are adapting.

Run the SwiftUI demo app and open the avatar picker. Observe: The above warning has disappeared. ☝️


etoledom (approved):
It seems to me that there should be a different way to manage these kind of scenarios.

I don't know which way can it be, and it will probably take time to investigate.

Since this change seems to work well and solves the issue, I suggest moving forward with this fix and go back to it later on.

What do you think?

pinarol:
Yeah I have the same feeling. 👍

PR #342, opened 2024-08-08
Update profile background according to designs

pinarol:
Closes #

Figma: https://www.figma.com/design/xgXj4jig9m0DYfKjFq6swG/QE-Native-Experience?node-id=44-1094&t=p8Nm4GVoG19D7XeP-0

And some discussions: `gravatarp2` ->  `/2024/08/01/native-avatar-picker-i1/#comment-2328`

### Description

Just updating the profile view background according to designs and latest discussions. Let me know if I missed sth.

### Testing Steps

Check out the Avatar picker in the SwiftUI demo app or the previews.


etoledom (approved):
Looking great!

I like the result in dark mode 🎉

PR #348, opened 2024-08-09
[Tooling] Remove the 'Build Demos' step from CI pipeline

AliSoftware:
# Why

As discussed in https://github.com/Automattic/Gravatar-SDK-iOS/pull/330#pullrequestreview-2229978625, it seems to me that the "Build Demo Project" job is redundant with us already building those Demo projects as part of Prototype Builds (as those would already fail and signal us if there was a compilation issue on the demo projects needing to be fixed).

<table>
<tr><th width="80px">Before</th><td>image</td></tr>
<tr><th>After</th><td>
image
</td></tr>
</table>

# How

I've thus removed the "Build Demo Project" job and the `.buildkite/commands/build-demos.sh` script from the CI pipeline.

I've kept the `build_demo` lane in the `Fastfile` for the time being, because it's also used in the `Makefile` (`build-demo-uikit` and `build-demo-swiftui` targets), and thus I figured maybe developers might still want to use that with `make` from time to time to locally validate and build those projects from the CLI (as opposed to via Xcode), even if we don't use those `make` targets and `build_demo` lane on CI anymore.

_(But let me know if you think developers won't ever need that `make build-demo-uikit`/`make build-demo-swiftui` and you prefer for us to remove it to clean things up)_

# Testing

As long as CI is green I think we should be good to go.

_(If we want to be extra sure, one could consider pushing a dummy commit introducing an intentional build failure on the Demo projects, check that CI detects it and fail appropriately, then revert the commit once it has proved its point.)_

# Next Steps: Update Required Checks

We'll need to update the GitHub branch protection rules to
 - [x] Remove "Build Demo Project" as a required CI check, now that it won't exist anymore
 - [x] Add the "Build UIKit Demo" and "Build SwiftUI Demo" jobs as required CI checks instead
 - [x] Update the `build-and-test` required check to its new name `build-and-test-swift-package`
 - [x] Inform developers that they'll need to rebase any of their WIP branches on top of develop after that change lands (so that they are not blocked by CI waiting for those 2 jobs as they'd not exist yet on their branch if they created their branch before this PR has landed)


andrewdmontgomery:
This great.  Good catch.

The original step of building the demo projects was there to make sure that the demo targets continue to be buildable.  A lot of our work happens by opening the Package in Xcode (not the Demo project).

With these new prototype builds, we definitely don't need to do that twice.

If for some reason we were to stop building prototypes, it would make sense to re-introduce these.

As for the Fastfile/Makefiles, I'm in favor of keeping them.  But I don't have a strong opinion.  They were originally there so that developers could perform CI steps to troubleshoot.  These prototype builds change that somewhere, since developers won't have the ability to run those locally.

andrewdmontgomery:
Ah, this will raise the question of whether prototype builds should block a PR.

If we drop the `build-demo-project` step, and allow the prototype build steps to perform this test for us, then the prototype build steps should become "required", since they implicitly test whether the Demo Project is buildable, and we want to block a PR if either of those targets is not buildable.

AliSoftware:
> Ah, this will raise the question of whether prototype builds should block a PR.
> 
> If we drop the `build-demo-project` step, and allow the prototype build steps to perform this test for us, then the prototype build steps should become "required", since they implicitly test whether the Demo Project is buildable, and we want to block a PR if either of those targets is not buildable.

Yep, hence the "Next Steps" section in my PR description 😉

andrewdmontgomery:
From https://github.com/Automattic/Gravatar-SDK-iOS/pull/330#issuecomment-2277834751
> it now makes even more sense to build those Prototype Builds unconditionally (and avoid skipping them based on labels or whatnot, otherwise we'd also not validate correct compilation of the demos) 

Very good point.

So if, for some reason, we decide to make those conditional, we'll need to rethink this.

That's fine.  Let's do this.  If we need to revisit, this is easy to revert (or rethink).

andrewdmontgomery:
> Yep, hence the "Next Steps" section in my PR description 😉

I am multitasking, and it shows.  Sorry!   (and thanks) 😁

andrewdmontgomery:
> consider pushing a dummy commit introducing an intentional build failure on the Demo projects

👍 works great

https://github.com/Automattic/Gravatar-SDK-iOS/pull/349#issuecomment-2278160648

andrewdmontgomery (approved):
Looks great, nice improvement

PR #350, opened 2024-08-12
Add a reusable SwiftUI Toast

### LINKED ISSUE #344: Quick Editor: Toast message
We need to show toast messages when the avatar changes successfully:
 
It should disappear on its own after a while, or, immediately with the "got it" button. 

Screenshot 2024-08-08 at 17 18 02

pinarol:
Closes #344 

From [figma](https://www.figma.com/design/xgXj4jig9m0DYfKjFq6swG/QE-Native-Experience?node-id=44-1094&t=yLKB0mpLbWKJq07R-0):

Screenshot 2024-08-12 at 15 43 15

### Description

- Adds an easily reusable toast view.
- Each toast is added to the toast queue by the business logic. The queue is observed by the ToastContainerView and reflected to UI.
- We can display multiple toast at a time.
- A Toast will dismiss itself after a while automatically.
- The toast duration is calculated according to a formula so longer messages will be displayed longer.
- Light & dark modes supported.

https://github.com/user-attachments/assets/3d89d702-7a1f-4d1f-a89f-27b761a2af9b

### Testing Steps

* SwiftUI demo app > Avatar Picker > Change avatar

You can also play around with ToastContainerView preview.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
@@ -45,13 +45,12 @@ class AvatarPickerViewModel: ObservableObject {
     @Published private(set) var isAvatarsLoading: Bool = false
     @Published var avatarIdentifier: AvatarIdentifier?
     @Published var profileModel: AvatarPickerProfileView.Model?
-    @ObservedObject var toastManager: ToastManager
+    @ObservedObject var toastManager: ToastManager = .init()
 
     init(email: Email, authToken: String) {
         self.email = email
         avatarIdentifier = .email(email)
         self.authToken = authToken
-        self.toastManager = ToastManager()
     }
 
     /// Internal init for previewing purposes. Do not make this public.
@@ -65,7 +64,6 @@ class AvatarPickerViewModel: ObservableObject {
         if let profileModel {
             self.profileResult = .success(profileModel)
         }
-        self.toastManager = ToastManager()
     }
 
     func selectAvatar(with id: String) {
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
@@ -45,11 +45,13 @@ class AvatarPickerViewModel: ObservableObject {
     @Published private(set) var isAvatarsLoading: Bool = false
     @Published var avatarIdentifier: AvatarIdentifier?
     @Published var profileModel: AvatarPickerProfileView.Model?
+    @ObservedObject var toastManager: ToastManager
 
     init(email: Email, authToken: String) {
         self.email = email
         avatarIdentifier = .email(email)
         self.authToken = authToken
+        self.toastManager = ToastManager()
     }
 
     /// Internal init for previewing purposes. Do not make this public.
@@ -63,6 +65,7 @@ class AvatarPickerViewModel: ObservableObject {
         if let profileModel {
             self.profileResult = .success(profileModel)
         }
+        self.toastManager = ToastManager()
     }
 
     func selectAvatar(with id: String) {
@@ -81,6 +84,7 @@ class AvatarPickerViewModel: ObservableObject {
             do {
                 setLoading(to: true, onAvatarWithID: id)
                 try await postAvatarSelection(with: id, identifier: .email(email))
+                toastManager.showToast("Avatar updated! May take a few minutes to appear every where.", type: .info)
             } catch APIError.responseError(let reason) where reason.cancelled {
                 // NoOp.
             } catch {
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
@@ -84,7 +84,7 @@ class AvatarPickerViewModel: ObservableObject {
             do {
                 setLoading(to: true, onAvatarWithID: id)
                 try await postAvatarSelection(with: id, identifier: .email(email))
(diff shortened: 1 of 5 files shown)


Sources/GravatarUI/SwiftUI/Toast/ToastManager.swift
```diff
@@ -0,0 +1,37 @@
...
+    private func calculateToastDuration(for message: String) -> TimeInterval {
+        let baseTime: TimeInterval = 2.0
+        let timePerCharacter: TimeInterval = 0.03
+        return baseTime + timePerCharacter * Double(message.count)
+    }
+}
+
+enum ToastType: Int {
+    case info
+    case error
```
pinarol, line 30: We still don't have the designs for an error toast so I didn't handle the UI side yet.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -45,11 +45,13 @@ class AvatarPickerViewModel: ObservableObject {
     @Published private(set) var isAvatarsLoading: Bool = false
     @Published var avatarIdentifier: AvatarIdentifier?
     @Published var profileModel: AvatarPickerProfileView.Model?
+    @ObservedObject var toastManager: ToastManager
```
etoledom: Can we maybe init the toast manager this directly here?
`@ObservedObject var toastManager = ToastManager()`

Then we don't need to instantiate it on every `AvatarPickerViewModel.init` method.

Sources/GravatarUI/SwiftUI/Toast/ToastContainerView.swift
```diff
@@ -0,0 +1,35 @@
...
+struct ToastContainerView: View {
+    @ObservedObject var toastManager: ToastManager
+
+    var body: some View {
+        VStack {
+            Spacer()
+            ForEach(toastManager.toasts) { toast in
+                Toast(toast: toast) { toast in
+                    toastManager.removeToast(toast.id)
+                }
```
etoledom: Just thinking that it's not easy to see what this closure is for. Maybe we can leave the label?
`Toast(toast: toast, dismissHandler: { toast in`

Anyway, it's internal component so not a big problem, but for future references might be good 🤔

Sources/GravatarUI/SwiftUI/Toast/ToastManager.swift
```diff
@@ -0,0 +1,37 @@
...
+    func removeToast(_ toastID: UUID) {
+        withAnimation {
+            toasts.removeAll { $0.id == toastID }
+        }
+    }
+
+    private func calculateToastDuration(for message: String) -> TimeInterval {
+        let baseTime: TimeInterval = 2.0
+        let timePerCharacter: TimeInterval = 0.03
+        return baseTime + timePerCharacter * Double(message.count)
```
etoledom, line 24: Nice! This dismiss time feels comfortable

etoledom (approved):
Looking pretty good!

A couple of observations which are **personal preferences**:

The opacity animation feels nice and soft, but I think that a transition from bottom animation cash more the attention, which I believe is something we want to achieve.

Maybe we can add a transition animation together with the opacity. This also plays good with the stacking animation.

`.transition(.opacity.combined(with: .move(edge: .bottom)))`

CleanShot 2024-08-13 at 14 09 12

It would be great to have a dismiss by the swipe down gesture. I believe many people would try this by default as first attempt of dismissing (more than tapping in a button).

The button (specially underlined) seems more like a web UX.

But if it's too complicated I don't mind to not have it for now 👍

Sources/GravatarUI/SwiftUI/Toast/ToastContainerView.swift
```diff
@@ -0,0 +1,35 @@
...
+        .frame(idealWidth: .infinity)
+        .animation(.spring(), value: toastManager.toasts)
+        .edgesIgnoringSafeArea(.bottom)
+        .padding(.bottom, .DS.Padding.half)
+    }
+}
+
+#Preview {
+    VStack {
+        var toastManager = ToastManager()
```
etoledom: This is giving a warning. Should probably be `let`.

etoledom (commented):
One more thing:

I see this warning on the console when the toast appears:
`Invalid frame dimension (negative or non-finite).`

Comming from:

CleanShot 2024-08-13 at 15 09 00@2x

pinarol:
> .transition(.opacity.combined(with: .move(edge: .bottom)))

This looks nice. I am adding.

> It would be great to have a dismiss by the swipe down gesture. I believe many people would try this by default as first attempt of dismissing (more than tapping in a button).

This is not a very simple thing to get right and not sure it should be in the 1st iteration. If we take care of the "must have" features timely we can tackle this I think. We can also consider "tap to any place to close" and keep things simple for now.

pinarol:
> Invalid frame dimension (negative or non-finite).

Seems like that line was unnecessary anyway so I removed it. The error is gone.

Sources/GravatarUI/SwiftUI/Toast/ToastContainerView.swift
pinarol: Yep. 👍

Sources/GravatarUI/SwiftUI/Toast/ToastContainerView.swift
pinarol: Agreed. Updated.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
pinarol: Right. Updated.

pinarol:
Ready for another look @etoledom

etoledom (approved):
Great! Thanks for the changes 🙏

PR #351, opened 2024-08-13
Extract Grid view from Picker view

etoledom:
Closes #

### Description

This PR tries to simplify the logic on AvatarPickerView and AvatarPickerViewModel.

I don't think this is a big simplification, but there's one big conceptual change which I like:

The avatar grid itself is now completely independent of the AvatarPickerView, with its own model which managed the status of the grid as a whole and of individual avatar cells, including avatar loading, selection, and (in the future) error state.

Avatar Grid model and view are independent of any networking (expect of image loading). All requests are managed by the  `AvatarPickerViewModel`, and updates the grid accordingly.

`AvatarPickerViewModel` holds a reference to the GridModel and updates it as needed, while `AvatarPickerView` will pass the grid model reference to the grid view.

In this way `AvatarPickerViewModel` is not responsible for holding the list of avatars, which avatar is selected, or individual avatar cell status, but it still holds the responses from networking.

### Testing Steps

- Using the SwiftUI example project.
- Upload avatars
- Select different avatars
  - Check that everything works as expected.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/GridModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/GridModel.swift
@@ -1,0 +1,55 @@
+import Foundation
+
+/// Describes and manages a grid of avatars.
+class GridModel: ObservableObject {
+    @Published var avatars: [AvatarImageModel]
+    @Published var selectedAvatar: AvatarImageModel?
+
+    var isEmpty: Bool {
+        avatars.isEmpty
+    }
+
+    init(avatars: [AvatarImageModel], selectedAvatar: AvatarImageModel? = nil) {
+        self.avatars = avatars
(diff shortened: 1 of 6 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/GridModel.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
+
+/// Describes and manages a grid of avatars.
+class GridModel: ObservableObject {
```
pinarol: `GridModel` is quite a generic name. Since this is dedicated to avatars, it can be renamed as `AvatarGridModel`.

pinarol (approved):
Looks good to me. 👍  Can be merged after resolving the new warnings related to AvatarGrid:

Screenshot 2024-08-14 at 12 22 34

PR #352, opened 2024-08-14
[Tooling] Prevent uploading Prototype Builds when there's no Pull Request

iangmaia:
This PR configures the main pipeline to prevent the Prototype Build upload jobs from running when there's no Pull Request yet.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -56,6 +56,7 @@ steps:
   # Prototype Builds of Demo Projects
   ###################
   - group: ":appcenter: Prototype Builds"
+    if: build.pull_request.id != null
     steps:
       - label: "🛠️ Build SwiftUI Demo"
         key: build_swiftui
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -56,7 +56,6 @@ steps:
   # Prototype Builds of Demo Projects
   ###################
   - group: ":appcenter: Prototype Builds"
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -56,6 +56,7 @@ steps:
   # Prototype Builds of Demo Projects
   ###################
   - group: ":appcenter: Prototype Builds"
+    if: build.pull_request.id != null
```
AliSoftware: I think we might still want to _build_ the demo projects on trunk (otherwise we have no other step in this pipeline that would validate on `trunk` that the Demo projects are compiling without errors), so I'd move this condition on the 2 "Upload * Demo to App Center" individual steps instead of on the whole `group`

iangmaia: Nice catch -- this briefly crossed my mind but I thought that building the package would be enough. But indeed this is in fact something else, and potentially easy to break when merging back to `trunk`. Updated on f88826db70833602a7871b39faf8e887fabb480f.

AliSoftware (approved):
New setup LGTM 👍 

Also, regarding https://github.com/Automattic/buildkite-ci/pull/492#pullrequestreview-2238123877 I see that the `pipeline.yml` already have `agent: { queue: "mac" }` at its top, so even if we want to also use the occasion of this to switch to the default steps, this is already ready on this end, and no additional change would be needed for that here. 👍 

:shipit:

PR #354, opened 2024-08-14
Handle avatar upload fail

### LINKED ISSUE #347: Quick Editor: Failure & retry states for image upload
We don’t have designs yet but we can maybe put a temp retry button and implement the logic until designs are ready. 
We should also show a toast message: "Ooops, there was an error uploading the image."

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/347

### Description

- Adds error style for the toast
- Replaces "Got it" with "X" icon
- Shows error toast when upload fails
- Adds a retry button on the failed image
- Shows error toast when avatar selection fails

https://github.com/user-attachments/assets/e583fd53-afc9-4b8a-8734-7c1693861c16

### Testing Steps

- SwiftUI Demo app > Avatar Picker View. You can disconnect from internet and try uploading image and changing your avatar.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
@@ -45,13 +45,12 @@ class AvatarPickerViewModel: ObservableObject {
     @Published private(set) var isAvatarsLoading: Bool = false
     @Published var avatarIdentifier: AvatarIdentifier?
     @Published var profileModel: AvatarPickerProfileView.Model?
-    @ObservedObject var toastManager: ToastManager
+    @ObservedObject var toastManager: ToastManager = .init()
 
     init(email: Email, authToken: String) {
         self.email = email
         avatarIdentifier = .email(email)
         self.authToken = authToken
-        self.toastManager = ToastManager()
     }
 
     /// Internal init for previewing purposes. Do not make this public.
@@ -65,7 +64,6 @@ class AvatarPickerViewModel: ObservableObject {
         if let profileModel {
             self.profileResult = .success(profileModel)
         }
-        self.toastManager = ToastManager()
     }
 
     func selectAvatar(with id: String) {
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
@@ -45,11 +45,13 @@ class AvatarPickerViewModel: ObservableObject {
     @Published private(set) var isAvatarsLoading: Bool = false
     @Published var avatarIdentifier: AvatarIdentifier?
     @Published var profileModel: AvatarPickerProfileView.Model?
+    @ObservedObject var toastManager: ToastManager
 
     init(email: Email, authToken: String) {
         self.email = email
         avatarIdentifier = .email(email)
         self.authToken = authToken
+        self.toastManager = ToastManager()
     }
 
     /// Internal init for previewing purposes. Do not make this public.
@@ -63,6 +65,7 @@ class AvatarPickerViewModel: ObservableObject {
         if let profileModel {
             self.profileResult = .success(profileModel)
         }
+        self.toastManager = ToastManager()
     }
 
     func selectAvatar(with id: String) {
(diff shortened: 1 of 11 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -157,16 +156,32 @@ class AvatarPickerViewModel: ObservableObject {
...
+              let model = grid.avatars.first(where: { $0.id == localID }),
+              let localImage = model.localUIImage
+        else {
+            return
+        }
+        grid.setLoading(to: true, onAvatarWithID: localID)
+        await doUpload(squareImage: localImage, localID: localID, accessToken: authToken)
+    }
+
+    private func doUpload(squareImage: UIImage, localID: String, accessToken: String) async {
```
etoledom, line 169: Now we have a `performUpload` and `doUpload` functions, which are pretty much equivalent in names 🤔 

What `performUpload` is doing now is basically creating and appending a new local image model to the grid.

Maybe we can remove it and merge it with `upload`

```swift
func upload(_ image: UIImage) async {
    guard let authToken else { return }

    let squareImage = image.squared()

    let localID = UUID().uuidString
    let localImageModel = AvatarImageModel(id: localID, source: .local(image: squareImage), isLoading: true)
    grid.append(localImageModel)

    await doUpload(squareImage: squareImage, localID: localID, accessToken: authToken)
}
```

pinarol: Makes total sense. Done.

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarGrid.swift
```diff
@@ -58,6 +59,11 @@ struct AvatarGrid: View {
                     if avatar.isLoading {
                         OverlayActivityIndicatorView()
                             .cornerRadius(Constants.avatarCornerRadius)
+                    } else if avatar.uploadHasFailed {
+                        DimmingRetryButton {
+                            onRetryUpload(avatar)
+                        }
+                        .cornerRadius(Constants.avatarCornerRadius)
                     }
```
etoledom: I like the name `DimmingRetryButton`.

Maybe we can rename `OverlayActivityIndicatorView` as `DimmingActivityIndicator`, since they are aesthetically equal 🤔

Sources/GravatarUI/SwiftUI/Toast/Toast.swift
```diff
@@ -14,33 +16,57 @@ struct Toast: View {
...
+                    .resizable()
+                    .aspectRatio(contentMode: .fill)
+                    .frame(width: 12, height: 12)
             }
         }
         .padding(.horizontal, .DS.Padding.double)
         .padding(.vertical, .DS.Padding.split)
         .background(backgroundColor)
+        .overlay(Rectangle().frame(width: toast.type == .error ? 4 : 0, height: nil, alignment: .leading).foregroundColor(Color.red), alignment: .leading)
         .cornerRadius(4)
```
etoledom: Not sure if this is a convention (or why the lint didn't or should pick this up), but I think it would improve readability if we add some new lines here.
```swift
        .background(backgroundColor)
        .overlay(
            Rectangle()
                .frame(width: toast.type == .error ? 4 : 0, height: nil, alignment: .leading)
                .foregroundColor(Color.red), alignment: .leading
        )
        .cornerRadius(4)
```

etoledom (approved):
Looking great!
Just a few non-blocking comments.

One more thing that could be important. I see this run-time warning:
`Invalid frame dimension (negative or non-finite).`

CleanShot 2024-08-16 at 10 56 25@2x


Tested both upload error and avatar selection error.

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarGrid.swift
pinarol: Done.

Sources/GravatarUI/SwiftUI/Toast/Toast.swift
pinarol: Exactly. 👍

pinarol:
> Invalid frame dimension (negative or non-finite).

Gosh my XCode doesn't show me those errors inline. Seems like that line is not necessary anyway so I fixed it by just removing it.

PR #355, opened 2024-08-15
Avatar Picker: Introduce horizontal grid layout

### LINKED ISSUE #345: Quick Editor: Horizontally scrolling avatars
https://www.figma.com/design/xgXj4jig9m0DYfKjFq6swG/QE-Native-Experience?node-id=44-1094&t=6N8vBEajkfr2yJEe-0

pinarol:
Closes #345 

From designs:

Screenshot 2024-08-15 at 15 34 50

### Description

Introduces the horizontally scrolling avatars. For now i didn't implement anything interactive like using number of avatars or height of the modal to decide the layout type. 

https://github.com/user-attachments/assets/b9476158-388e-4b33-a8d7-31709469e66e

- Extracts the avatar view to a separate View named `AvatarPickerAvatarView` to be able to use it both from horizontal & vertical grids.
- Updates the UI according to designs to put the content inside the borders whether it's vertical or horizontal. 
- Extracted the border code to an internal modifier function because it is used by error views too.
- Updated the demo app to add the layout options.

### Testing Steps

- SwiftUI demo app > Avatar picker view > Play around with the "content layout" parameter.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
@@ -2,6 +2,8 @@ import SwiftUI
 
 struct AvatarPickerAvatarView: View {
     let avatar: AvatarImageModel
+    let maxLength: CGFloat
+    let minLength: CGFloat
     let shouldSelect: () -> Bool
     let onAvatarTap: (AvatarImageModel) -> Void
     let onImageSelected: (UIImage) -> Void
@@ -18,10 +20,10 @@ struct AvatarPickerAvatarView: View {
         )
         .scaledToFill()
         .frame(
-            minWidth: AvatarGridConstants.minAvatarWidth,
-            maxWidth: AvatarGridConstants.maxAvatarWidth,
-            minHeight: AvatarGridConstants.minAvatarWidth,
-            maxHeight: AvatarGridConstants.maxAvatarWidth
+            minWidth: minLength,
+            maxWidth: maxLength,
+            minHeight: minLength,
+            maxHeight: maxLength
         )
         .background(Color(UIColor.secondarySystemBackground))
         .aspectRatio(1, contentMode: .fill)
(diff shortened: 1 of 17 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
```diff
@@ -0,0 +1,59 @@
+import SwiftUI
+
+struct AvatarPickerAvatarView: View {
+    let avatar: AvatarImageModel
+    let shouldSelect: () -> Bool
+    let onAvatarTap: (AvatarImageModel) -> Void
+    let onImageSelected: (UIImage) -> Void
```
etoledom: I don't think we need `onImageSelected` here.

This is the callback when a new image is selected from the + cell in the grid, which only exists on the Vertical layout.

Maybe the name is confusing? 🤔

etoledom: Also `HorizontalAvatarGrid` shouldn't need this.

etoledom (commented):
I see some flashing when selecting a different cell on Horizontal layout:

CleanShot 2024-08-16 at 11 29 36

I don't see this happening on Vertical layout.

@pinarol - Just wondering if you have seen this also

etoledom:
One more thing:

On the Horizontal layout, it doesn't seem obvious that I can scroll horizontally when I see 3 avatars aligned on the screen. (Testing on iPhone 15 sim)

The design shows a bit of a 4th avatar appearing from the right. This helps the user see that there's more content and discover the scrolling. 

Can we do this also and have this extra bit of 4th avatar appearing independent of the screen width?
I know that on UICollectionView we can have a layout based on percentages of the screen, maybe there's a similar thing on SwiftUI 🤔

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
pinarol: I see, yes the name is confusing then. I'll rename and remove from here.

pinarol:
I made some changes to avoid glitches. The horizontal version uses fixed size avatars now.

Also, I tried calculating the avatar size with the help of GeometryReader to display 3.25 avatars on every device but it is messing up the height of `HorizontalAvatarGrid`. I tried setting the height manually but no good way of doing it. So I left it. I made the avatar larger and the space between each avatar smaller. This way it is also more close to the designs. I think this will make it less likely to happen. 

Also, it is not great to always show 3.25 avatars, it is causing giant avatars in landscape mode and in ipad. This needs a more sophisticated formula. But the bigger problem is the height of `HorizontalAvatarGrid`.  
Can you try again? @etoledom

etoledom (approved):
Looking great!
Thanks for the changes 🙏

PR #356, opened 2024-08-16
Setup Legacy Localization Support

### LINKED ISSUE #339: Setup support for localization

andrewdmontgomery:
Closes #339 

### Description

This sets up localization of the SDK and Demo apps:
- Configures the SDK Package to support localization
- Configures the `GravatarUI` target with a `Localizable.strings` file
- Updates Fastlane and the Makefile to allow for automated generation of strings from source code
- Updates the source code implementation of localized strings:
   - Uses reverse-dns notation for string keys
   - Consolidates some localized strings into a `private enum Localized` as an extension of the SwiftUI view they are used in
   - Updates (and adds) some "comments" to localized strings

For this PR, I would like to agree on the key names.  Once we've agreed on those, I can upload our `.strings` to GlotPress and set up the automation for (eventually) downloading translations. 

Feedback on the "comments" is very welcome.  But we can make changes to those up until we submit strings for localization.

#### Note: this is just basic support for localization
This only introduces support for localization.  Notably:
- This does not integrate with GlotPress
- The existing Makefile/Fastlane functionality is there for demonstration purposes only.  Once we have decided on a localization flow, this part will change considerably.  But the basic functionality will be the same: the content of the `Localizable.strings` files will be maintained by automation, not manually.

### Testing Steps

Required for this PR:
- [ ] Check the changes to localized string keys that I have made.  Ok?  Too much?
- [ ] Check out the `private enum Localized {...}` implementations.  Ok?  Too much?

Nice to have:
- [ ] Check out the comments on strings to make sure they accurately reflect the context of the string

#### Update an SDK source file and generate strings

1. Find a source file with a localizable string
2. Modify it:
   - Add or modify a comment
   - Modify a key
   - Modify a value
 3. Update the `Localizable.strings` file using the command line: `make generate-strings`
 - [ ] OBSERVE: you should see your change appear in `Sources/GavatarUI/Resources/en.lproj/Localizable.strings`

#### Update a Demo source file and generate strings

Right now, only one string is localizable.  It's the `Text("Email:")` view in `DemoAvatarPickerView`.  For this test, you can either modify that one, or (if you're familiar) add localization support to one of other views.

1. Find a source file with a localizable string
2. Modify it:
  - Add or modify a comment
  - Modify a key
  - Modify a value
 3. Update the `Localizable.strings` file using the command line: `make generate-strings`
 - [ ] OBSERVE: you should see your change appear in `Sources/GavatarUI/Resources/en.lproj/Localizable.strings`


andrewdmontgomery:
I want to call out one change especially:

`Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift`

Right now, the `private enum Localized {...}` uses some nested `enum` to structure these string references.  I also tried this with a flat `enum Localized`, too, with entries like:

- `static let contentLoadingSuccessTitle = ...`
- `static let contentLoadingSuccessSubtext = ...`
- `static let contentLoadingSessionExpiredTitle = ...`
- `static let contentLoadingSessionExpiredSubtitle = ...`

And then calling:

- `title: Localized.contentLoadingSuccessTitle`
- `title: Localized.contentLoadingSuccessSubtext`
- `title: Localized.contentLoadingSessionExpiredTitle`
- `title: Localized.contentLoadingSessionExpiredSubtext`

But with nesting, I found it a bit easier to keep track of everything:

- Easier to reason within the `Localized` enum
- Easier to reason from the call site

I find that the calling site is easier to read this way:

- `title: Localized.ContentLoading.Success.title`
- `title: Localized.ContentLoading.Success.subtext`
- `title: Localized.ContentLoading.Failure.SessionExpired.subtext`
- `title: Localized.ContentLoading.Failure.SessionExpired.subtext`

And it's easier to scan the `enum`:

```swift
private enum Localized {
    enum Header {
        static let title = ...
        static let subtext = ...
    }

    enum ContentLoading {
        enum Success {
            static let title = ...
            static let subtext = ...
        }

        enum Failure {
            // etc.
        }
    }
}
```

Thoughts?

cc: @AliSoftware

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/CTAButtonView.swift
```diff
@@ -1,9 +1,9 @@
 import SwiftUI
 
 struct CTAButtonView: View {
-    let titleKey: LocalizedStringKey
+    let titleKey: String
```
AliSoftware: Now that we're supposed to pass the already-localized string to this iinm, you should probably consider also renaming the variable from `titleKey` to just `title` or to `localizedTitle` for consistency?

Sources/GravatarUI/SwiftUI/CachedAsyncImage.swift
```diff
@@ -85,21 +85,21 @@ struct CachedAsyncImage<Content: View>: View {
         with: .email("email@google.com"),
         options: .init(preferredSize: .points(100))
     ) else {
-        return Text("Invalid URL")
+        return Text(verbatim: "Invalid URL")
```
AliSoftware, line 88: Is this use of `verbatim:` intentional here? Is this not supposed to be ultimately localized?

Sources/GravatarUI/SwiftUI/Toast/ToastContainerView.swift
```diff
@@ -29,13 +29,13 @@ struct ToastContainerView: View {
         Button {
             toastManager.showToast("Hi! This is a toast! You can show multiple toasts at a time!")
         } label: {
-            Text("Show info toast!")
+            Text(verbatim: "Show info toast!")
```
AliSoftware, line 32: Ditto

fastlane/lib/lanes/localization.rb
```diff
@@ -0,0 +1,84 @@
...
+    lane :generate_strings_file_demo do |options|
+      Dir.mktmpdir do |tempdir|
+        demo_en_lproj = File.join('Demo', 'Demo', 'Localizations', 'en.lproj')
+        ios_generate_strings_file_from_code(
+          paths: [
+            File.join('Demo', 'Demo', 'Gravatar-UIKit-Demo'),
+            File.join('Demo', 'Demo', 'Gravatar-SwiftUI-Demo')
+          ],
+          output_dir: tempdir,
+          swiftui: true
```
AliSoftware: Should we really use the SwiftUI flag of genstrings here while by convention from what we discussed in the P2 we'd recommend against using `Text("English copy")` or similar `Text` initializers to do localization via SwiftUI and instead recommend using NSLocalizedString everywhere in code for consistency?

fastlane/lib/lanes/localization.rb
```diff
@@ -0,0 +1,84 @@
...
+        demo_en_lproj = File.join('Sources', 'GravatarUI', 'Resources', 'en.lproj')
+        ios_generate_strings_file_from_code(
+          paths: [
+            File.join('Sources', 'GravatarUI')
+          ],
+          output_dir: tempdir,
+          swiftui: true
+        )
+  
+        Dir.chdir("..") do
```
AliSoftware: Seems a bit confusing imho to have a `..` here imho, especially while you're still within the `Dir.mktempdir`

For a second I thought you previously also `chdir`'d into the `tempdir` after having `mktempdir` it (while you have not), which made me incorrectly think you would be escaping out of the `tempdir` you were supposed to work from.

After more careful reading I now realize that you never actually `chdir`'d into the `tempdir`, and this `Dir.chdir('..')`, despite being nested within a `Dir.mktempdir` block, referenced the root of the repo, since that `..` is relative to the `fastlane/` folder not to `tempdir`.

I think it'd be clearer to remove this `Dir.chdir` from this line and use `'..'` directly in the `File.join('..', demo_en_lproj, 'Localizable.strings')` below instead, wdyt?

fastlane/lib/lanes/localization.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+  
+    private_lane :utf16_to_utf8 do |options|
+      next unless options[:source]
+      next unless options[:destination]
+  
+      source = options[:source]
+      destination = options[:destination]
+  
+      if File.exist?(source)
```
AliSoftware: Using an early exit (`next unless File.exist?(source)`) would allow you to reduce the level of indentation in the rest of this lane's code.

PS: yes [to early-exit a Fastlane `lane` (or `private_lane`) you have to use `next` and not `return`](https://docs.fastlane.tools/advanced/lanes/#stop-executing-a-lane-early) (because technically lanes are Ruby blocks, not functions)

fastlane/lib/lanes/localization.rb
```diff
@@ -0,0 +1,84 @@
...
+  
+          utf16_to_utf8(
+            source: utf16_strings,
+            destination: utf8_strings
+          )
+        end
+      end
+    end
+  
+    private_lane :utf16_to_utf8 do |options|
```
AliSoftware, line 58: I'm curious what's the need for this.

 - Is it just a matter of personal preference to have the strings files in UTF8 rather than UTF16?
 - Or is it just because the demo projects were configured with the file references to those being UTF8 and you didn't realize you could change it in the Xcode project?
 - Or is it because it's causing trouble with SwiftPM expecting localized string files and bundled text resources referenced in a Package.swift to be in UTF8?

I know that, by design, `genstrings` expects source code files it parses o be UTF8, but will generate UTF16 strings files as output. Not sure why Apple chose that in particular, but that hasn't caused us any issues in other projects, as long as the .strings files are referenced with the right encoding in the pbxproj too.

If this is a genuine need for things to work well with SPM/Package.swift or similar to have UTF8 generated files in your case, I wonder if it wouldn't be better for us to update our Fastlane action's implementation in our `release-toolkit` to allow passing the preferred encoding used for generated files, and thus move your implementation of this extra conversion step form here to the action itself, to make the action more versatile and hide this implementation detail behind the action itself?

AliSoftware (commented):
Re: the nested enums, I'm all good with this approach.

In fact, when I was still active in developing SwiftGen, I made the "nested" template for strings resources be the default instead of the "flat" one for the same reason, because I found the call site to look nicer 👍

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/CTAButtonView.swift
pinarol: Right. Just `title` would be sufficient btw 👍

Sources/GravatarUI/SwiftUI/CachedAsyncImage.swift
```diff
@@ -85,21 +85,21 @@ struct CachedAsyncImage<Content: View>: View {
...
-            Text("empty")
+            Text(verbatim: "empty")
         case .success(let image):
             image
         case .failure:
-            Text("failure")
+            Text(verbatim: "failure")
         default:
-            Text("failure")
+            Text(verbatim: "failure")
```
pinarol, line 102: These are inside a `#Preview` so we don't need the "verbatim" keyword here I think. I mean, it doesn't hurt but it can be a little confusing.

pinarol:
Some keys have too much nesting for my taste at the first look but it all makes sense after reading the code a bit. So LGTM. 👍

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/CTAButtonView.swift
andrewdmontgomery: good catch

fastlane/lib/lanes/localization.rb
andrewdmontgomery: ooh, nope.  Good catch.  I added this while I was still coming to understand the limitations of `genstrings` and SwiftUI.

andrewdmontgomery:
> Some keys have too much nesting for my taste at the first look but it all makes sense after reading the code a bit. So LGTM. 👍

Yeah, that's how I felt, too.

Sources/GravatarUI/SwiftUI/CachedAsyncImage.swift
andrewdmontgomery: Good question.  It is intentional, in the sense that I want to be explicit about our intensions.  But I'm wide open to the decision of whether this should be localized.  @pinarol has a comment about this, too (below).  So I'll pick up this conversation there.

Sources/GravatarUI/SwiftUI/CachedAsyncImage.swift
andrewdmontgomery: There are two bits at play, here.

1. Without the `verbatim` parameter name, these strings are _technically_ localizable.  As you will see in the next point, I am going to make a change that will prevent these `Text("static string")` instances from being localized... for now.  But that is a bit of a hack (see below).  If we do not want to localize these (and that is an open question), then we should explicitly state that these strings should be used verbatim.  In the future, when we move to String Catalogs, without that `verbatim` (or anothe approach, see below), these will be added as key/value pairs to the String Catalogs.
2. Right now, when we call `genstrings`, we're passing `-SwiftUI`, which causes it to look for localizable strings in `Text()` objects.  As @AliSoftware mentioned in another comment, we probably don't want to do pass that option.  We actually want to avoid using those default initializers with static strings, and instead pass `NSLocalizedString()` objects.  I will remove that option.  Technically, that means we could stop using `verbatim`.  But that's a bit of a hack, which depends on the fact that `genstrings` doesn't localize SwiftUI types.

Basically, we should choose:
- If we don't want to localize these, we should make them not localizable.  There are two ways to do that:
   -  Use `verbatim` to make that explicit
   - Pass these strings as variables into `Text()`.  This will have the same effect since only static strings are localized.  (e.g. `Text(varName)`).  SwiftUI assumes that strings passed as variables should not (by default) be localized, since they could be coming from user input.  That would be weird, here.
- If we do want to localize these, we'll update them to use `NSLocalizedString()`.

fastlane/lib/lanes/localization.rb
andrewdmontgomery: Yeah, good catch.  This approach is a relic from a previous attempt.

fastlane/lib/lanes/localization.rb
andrewdmontgomery: > Or is it just because the demo projects were configured with the file references to those being UTF8 and you didn't realize you could change it in the Xcode project?

Nice, I didn't know that.

> I know that, by design, genstrings expects source code files it parses o be UTF8, but will generate UTF16 strings files as output. Not sure why Apple chose that in particular, but that hasn't caused us any issues in other projects, as long as the .strings files are referenced with the right encoding in the pbxproj too.

That was the challenge.  What actually tripped me is the fact that those UTF16 string files aren't readable by the GitHub Mac client:
image

After a little poking around (and only a little), it seems that UTF-16 support in github is complicated.  In the process, I discovered that Apple recommends (or at least _recommended in the past_) saving these as UTF-8:

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html
> Note: It is recommended that you save strings files using the UTF-8 encoding, which is the default encoding for standard strings files. Xcode automatically transcodes strings files from UTF-8 to UTF-16 when they’re copied into the product.

That might also explain why `genstrings` generates UTF-16 files.

So rather than bang my head against this, I just wrote a little code to convert.

> but that hasn't caused us any issues in other projects

That's good to know, and I noticed that.  I was originally envisioning this implementation as temporary: just a way for us to manually generate the strings file while we sort out the automation.  I expected us to retire this implementation when the time came to setup the CI parts of this.

AliSoftware: Yeah I remember now how `genstrings` generating UTF16 might be a leftover from old standards (probably related to ObjC and `NSString` being internally backed by UTF16 buffers), while those days Apple changed their preferred encoding for it and now recommends UTF8 (which is in line with `Swift.String`'s backing as well).

IMHO The mere fact that GitHub doesn't play well with UTF16 files and will thus making it more challenging to look at diff of those… is reason enough to save the generated files in UTF8.

I'll take not to add an option for it in `ios_generate_strings_files_from_code` at some point so we can move your implementation of this helper directly into the action 👍

fastlane/lib/lanes/localization.rb
andrewdmontgomery: ooh yes, thanks.

fastlane/lib/lanes/localization.rb
andrewdmontgomery: Nice.  When I wrote this, I definitely did NOT have a strong opinion on whether we should (in the long run) commit these as UTF-8 vs UTF-16.  At the very least, I was going to ping you and AppsInfra to chime in, since you've been living with this approach for far longer than I have.

But it's validating (to say the least) to hear your thoughts.  Thanks.

andrewdmontgomery:
@AliSoftware do you have any thoughts about how/when to generate the `en` source `Localizable.string`?  Does it make sense to have the PR's that make changes to localizable strings also commit the changes within the PR?  In this case, we'd want a check for uncommitted strings.  We could even (potentially) have a githook.

AliSoftware:
> @AliSoftware do you have any thoughts about how/when to generate the `en` source `Localizable.string`? Does it make sense to have the PR's that make changes to localizable strings also commit the changes within the PR? In this case, we'd want a check for uncommitted strings. We could even (potentially) have a githook.

Ah, good question… usually for mobile apps we do that at the moment of code freeze, which then allows those generated `en.lproj` strings to be frozen for the duration of the beta-testing phase as well as sending those generated English copies for translation (to GlotPress or whatever platform we use), and then download the translations at the end of the beta-testing phase (typically ≈5 days later) just before doing the final release.

But to my understanding, for Gravatar SDK, you don't plan to follow a "code-freeze + beta-testing phase" release process, but instead just create a tag whenever you feel ready to release a new version?

---

If so, the best options I could think of from the top of my head right now could be:

## Option 1: As a precommit hook (as you mention)

 - That would allow/force developers to include any necessary strings changes generated by the lane in the commits they push to their branch before their PR gets merged
 - We could then have a CI check that would re-running the same lane to generate the strings file, and only pass if the diff for the strings files is empty, fail if it's not (indicating the developers didn't run the lane and commit the necessary new strings locally on their end)

The main drawback of this approach is that git hooks are local. So developers and contributors would have to install the hook on their development machine for the process to work well as expected.

## Option 2: make CI push the updated strings file

Alternatively, we could decide to have CI do the generation of the file and pushing the changes itself.

But tbh I'm not a fan of having CI do git pushes. First because this involves security considerations (CI has to have a token that is allowed to push, and we should thus make sure this token is super secure).

Also, because that could easily create conflicts—especially if a developer do quick iterations on their PRs while the CI has already started a build on a preceding commit and that CI build is about to push the commit with updated strings at the end of that build… but the developer end up pushing a new commit in-between…

Finally, if we were to go with that approach, that would beg the question if CI should do such commit to regenerate strings files (if diff for it is non-empty):

### 1a: On each commit of PR branches
  - Would allow the generated strings file can be reviewed alongside the PR, and are properly included in the branch when the PR ends up being merged…
  - …but with the drawback of a PR possibly ending up with many CI-generated commits if one iterates a lot of a PR, and the added risk of conflicts as well

### 2b: Only on `trunk`, once the PR has landed

 - This would reduce the number of intermediate CI-generated commits and reduce the risk of conflicts on PRs worked on iteratively…
 - …but would mean the changes in the strings file won't be included in the PR being reviewed, nor be associated with the branch being merged, which could cause inconsistencies when looking at the git history, doing git-bisects, reverting PRs, …

### 2c: Only on tags, at the time the release is created

 - Would limit the number of commits done by CI (and thus risk of conflicts, etc), by only doing it once we're ready to release
 - But the major drawback is that this means the new strings generated at that time would, by design, not have any translation at all, since the tag would already be for building the final release. So I don't think that's an option that makes a ton of sense (but just listed here for completeness)

---

TBH Non of those options feel perfect to me.

The last option would be for you to consider introducing a release process for the SDK, i.e. not just "when I want to do a release I just create a GitHub Release and git tag and CI will publish it and that's done", but a process more like "code-freeze", where you'd at least:
 - "create the preparation phase of the release" at the start of the process—via a lane you'd run that would create a `release/` branch, push commits with the updated `strings` file regenerated to it, maybe use the occasion to update the release notes and/or version constants…
 - then wait a few days for the translations to be ready
 - and finally trigger another lane that would finalize the release (pull the latest translations, run CI on it e.g. to validate placeholders consistency, create the GitHub Release and tag, and push to CocoaPods or any other registry)

That would make the release process of the SDK a bit more convoluted… but would provide better flexibility (not just for localizations, but also release notes, potential beta-testing…)

andrewdmontgomery:
> But tbh I'm not a fan of having CI do git pushes.

Fully agree.  And for so many reasons.

> The main drawback of [using a githook] is that git hooks are local

Yeah.  And then we're left with the question of what to do when that doesn't happen.  Does CI fail and block a PR?  (don't love that.)  Does CI start auto-committing? (again, don't love that.)

> The last option would be for you to consider introducing a release process for the SDK

I think this is the answer.  The fact of the matter is, since the SDK will be localized, **_we already have a release process_**, in the sense that there is a process, with steps that block a release:

2. We need to submit strings for localizing
3. We need to wait for those strings to be localized
4. We need to integrate those localizations

That's a process, and it blocks a release.  We should understand the implications of that process and make decisions about how we want it to progress.  And then we use that to automate the actions that are needed to complete that process consistently.

We also try to sync up our releases (roughly) with the Android SDK.  Yet another reason to make this a formal step in the release process.

Ok, I'll work on what that might look like.  And thanks for your roughed out idea of how that might work.

andrewdmontgomery:
btw, we already have a release process:
https://lanternp2.wordpress.com/sdk-releases/ios-sdk-release-manual/

AliSoftware:
> btw, we already have a release process: Pfnfe0-d4-p2

Neat!

💡  Once the process is more settled (as I assume that checklist might soon change slightly to account for those questions around L10n and translations integration), we should probably consider migrating that checklist from a P2 page to our internal ReleaseV2 tool in MC at some point in the future. Even if the first iteration of that would still only be a checklist of manual steps—and we only automate those steps (via buttons on the release scenario of ReleasesV2 calling appropriate fastlane lanes in CI) later—that could already be beneficial e.g. to give ability to track the progress of such checklists within the tool, allow release managers to leave notes and breadcrumbs on each checklist item, etc…

Though I guess you'll probably want to run a couple of releases first—to test and consolidate the current process (and account for changes in it like for L10n)—and officialize it more in ReleaseV2 only when it's considered "stable enough"…?

andrewdmontgomery:
Whoa, that's nifty.  I was unaware of this tool.  This looks lovely.

> Though I guess you'll probably want to run a couple of releases first—to test and consolidate the current process (and account for changes in it like for L10n)—and officialize it more in ReleaseV2 only when it's considered "stable enough"…?

Yeah, the P2 page (and the spreadsheet that supports it) is here to be low-lift, and to give us the flexibility to discover our processes before we start stamping that process into a tool.

I like this goal.

pinarol (approved):
LGTM 🎉

PR #358, opened 2024-08-20
Add navigation bar to the avatar picker

### LINKED ISSUE #343: Quick Editor: Navigation Bar
https://www.figma.com/design/xgXj4jig9m0DYfKjFq6swG/QE-Native-Experience?node-id=44-1094&t=6N8vBEajkfr2yJEe-0

Screenshot 2024-08-09 at 13 17 36

Let's cover both light & dark modes.
Clarification needed: What happens when the Gravatar icon is tapped?

pinarol:
Closes #343 

### Description

Adds navigation bar. I used `NavigationView` for now since it is supported in iOS 15. `NavigationStack` is supported only for iOS 16+.

- Adds "Done" button to close the picker
- Gravatar icon opens Safari with profile url

Updates the "Discover Your Gravatar Card" to "View profile" based on designs. Ties the action to open the profile in Safari.

Refactors some of the paddings based on the designs.

### Testing Steps

- Play around with the SwiftUI demo app.

### DIFF
--- a/Sources/GravatarUI/Base/URL+Additions.swift
+++ b/Sources/GravatarUI/Base/URL+Additions.swift
@@ -1,0 +1,7 @@
+import Foundation
+
+extension URL: Identifiable {
+    public var id: String {
+        absoluteString
+    }
+}
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerProfileView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerProfileView.swift
@@ -19,7 +19,7 @@ struct AvatarPickerProfileView: View {
     @StateObject private var placeholderColorManager: ProfileViewPlaceholderColorManager = .init()
     @Environment(\.colorScheme) var colorScheme: ColorScheme
 
-    private(set) var viewProfileAction: ((URL?) -> Void)? = nil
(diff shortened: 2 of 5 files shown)


etoledom (approved):
Looks good to me!

:shipit: 

There's one detail which troubles me a bit, but I don't think there's much we can do without making things overly complicated:

SFSafariViewController has a custom modal presentation animation which emulates a navigation push. It's the standard animation in which the in-app browser is presented.


https://github.com/user-attachments/assets/2fe89dd6-433c-43f6-8e88-16883e9478c5

SwiftUI doesn't seem to be compatible with these custom presentation animations, and we lose the standard SFSafariViewController presentation.

I saw that most resources online mentions using `.sheet` presentation, which is the easiest. I tested using NavigationLink but it won't work either.
The closest is to use `.fullScreenCover()` instead of `.sheet()`, so it presents full screen.

The complex option I saw is to present SFSafariView in UIKit side from UIApplicationDelegate or UIWindow, and call it from SwiftUI using a global reference. But we also don't have the UIKit context in a pure swiftUI app. ¯\_(ツ)_/¯ 

Either way, lets go along with this one. Hopefully Apple will do something about this at some point.

Either way, I'm fine how it is now

pinarol:
I liked "fullScreenCover" better. 👍 For now it seems like the best we can get out of SwiftUI about this.

PR #359, opened 2024-08-21
Gravatar OAuth in SwiftUI

etoledom:
### Description
This PR implements the Gravatar OAuth flow.

CleanShot 2024-08-22 at 14 51 23

### Testing Steps

- Build and run the SwiftUI demo app
- Go to Profile Editor with oauth demo screen.
- insert an email and open the editor.
  - Check that OAuth flow works as expected.


Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -0,0 +1,190 @@
...
+}
+
+enum OAuthError: Error {
+    case notConfigured
+    case couldNotCreateOAuthURLWithGivenSecrets
+    case couldNotGetCodeFromURL(String)
+    case oauthResponseError(String)
+    case unknown(Error)
+    case couldNotStoreToken(Error)
+}
```
etoledom: This became a bit messy and I don't like it:

- `notConfigured`: Secrets are not set up. This maybe could be a `assertionFailure` instead of throwing? The idea is to warn the developer.
- `couldNotCreateOAuthURLWithGivenSecrets` this comes from using `Encodable` to add the query items to the URL. Probably if we add them manually we wouldn't need this error. And if this error happens, it's probably a developer mistake which should be catch on dev time (maybe assertion also?)
- `case couldNotGetCodeFromURL(String)` error on parsing the callback URL to get the code. This could be a runtime error but not sure how could this be handled 🤔 .
- `oauthResponseError(String)` Something went wrong at the backend. Needs to be handled in the UI.
- `unknown` ¯\_(ツ)_/¯  ... Probably on the UI also.
- `couldNotStoreToken(Error)` Keychain error. I got errors here because the keychan was wrongly set up. But it was a dev issue. Probably there are instances that this could be a user issue, but again, not sure how this should be handled. Technically we have the token and the user should be able to edit the profile, so I don't think that throwing is a good option. The problem is that they will need to be authenticated again next time.

Sources/GravatarUI/SwiftUI/OAuthSession/Keychain.swift
```diff
@@ -0,0 +1,73 @@
+import Foundation
+
+protocol SecureStorage: Sendable {
+    func setPassword(_ password: String, for key: String) throws
+    func deletePassword(with key: String) throws
+    func password(with key: String) throws -> String?
```
pinarol: `password` is not quite the right terminology I think, how about `token` or `secret`? `password` is more for "something I know" kind of secrets. 🤔

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -0,0 +1,190 @@
...
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
+        return decoder
+    }()
+
+    init(authenticationSession: AuthenticationSession = OldAuthenticationSession(), storage: SecureStorage = Keychain()) {
+        self.authenticationSession = authenticationSession
+        self.storage = storage
+    }
+
+    func isSession(with email: Email) -> Bool {
```
pinarol: how about `hasSession(with: ...)`?

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -0,0 +1,190 @@
...
+        request.httpBody = try params.queryItems.string?.data(using: .utf8)
+
+        return request
+    }
+}
+
+enum OAuthError: Error {
+    case notConfigured
+    case couldNotCreateOAuthURLWithGivenSecrets
+    case couldNotGetCodeFromURL(String)
```
pinarol: this was a bit ambiguous to me so I suggest:

> couldNotParseAccessCode

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
pinarol: I would add the assertionFailures for `notConfigured` and `couldNotCreateOAuthURLWithGivenSecrets` but not necessarily replace the "throw" statements. We still need the throws for prod. Throws also allow us return non-optional values.

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -0,0 +1,190 @@
...
+
+    func deleteSession(with email: Email) {
+        try? storage.deletePassword(with: email.rawValue)
+    }
+
+    func sessionToken(with email: Email) -> String? {
+        try? storage.password(with: email.rawValue)
+    }
+
+    func authenticate(with email: Email) async throws -> String {
```
pinarol: It is not easily understandable what this method does and returns just by looking at the signature. `authenticate` is just half of OAuth. The work done here is I think described better by something like:

> acquireAccessToken(for email: Email) async throws -> String

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
pinarol: > couldNotStoreToken(Error) Keychain error. I got errors here because the keychan was wrongly set up. But it was a dev issue. Probably there are instances that this could be a user issue, but again, not sure how this should be handled. Technically we have the token and the user should be able to edit the profile, so I don't think that throwing is a good option. The problem is that they will need to be authenticated again next time

As you said, since the user is able to continue I think we don't need to re-throw this one. This can be something we swallow inside a separate `do { ... } catch {...}` and handle just with an assertionFailure.

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -0,0 +1,190 @@
...
+            return error
+        case let error as NSError:
+            if error.domain == ASWebAuthenticationSessionErrorDomain {
+                return .oauthResponseError(error.localizedDescription)
+            }
+            return .unknown(error)
+        case let error as Keychain.KeychainError:
+            return .couldNotStoreToken(error)
+        default:
+            print("🛑 Error: \(type(of: error)) - \(error.localizedDescription)")
```
pinarol: let's remove the "print" call. We can probably use the Apple’s os.log framework for logging, doesn't have to be in this PR though. 🤔

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -0,0 +1,190 @@
...
+    }
+}
+
+enum OAuthError: Error {
+    case notConfigured
+    case couldNotCreateOAuthURLWithGivenSecrets
+    case couldNotGetCodeFromURL(String)
+    case oauthResponseError(String)
+    case unknown(Error)
+    case couldNotStoreToken(Error)
```
pinarol, line 112: I think `DecodingError` deserves a separate case here.

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -0,0 +1,190 @@
...
+}
+
+extension OAuthError {
+    static func from(error: Error) -> OAuthError {
+        switch error {
```
pinarol, line 118: We can catch "DecodingError"s and re-throw them after adding an assertionFailure maybe? This means our decoding keys or types are not matching the response, so the developer should fix it.

```
case let error as DecodingError:
   assertionFailure("Unable to decode the response. Error: \(error.description)")
   throw .decodingError(error)

```

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -42,7 +42,7 @@ struct AvatarPickerView: View {
                         avatarsLoadingView()
                     }
                 }
-                .task {
+                .onAppear() {
```
pinarol: just curious, why is this change necessary?

Sources/GravatarUI/SwiftUI/ProfileEditor/ProfileEditor.swift
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    var body: some View {
+        VStack {
+            if isSession, let token = oauthSession.sessionToken(with: email) {
+                switch entryPoint {
+                case .avatarPicker:
+                    AvatarPickerView(model: .init(email: email, authToken: token))
+                }
+                Button("Log out (for testing only)") {
```
pinarol: This is meant to be a temp Button I guess but I feel like it can be useful during development in the long run as well. So how about we wrap it with some kind of compiler flag and keep it?

```
#if DEMO

#endif
```

pinarol (commented):
Looks pretty good overall. Just dropping some non-critical comments and a question.

We need UI for this 2 case but it doesn't have to be in this PR.

- we don’t have the token yet but we’ll try (we need an activity indicating UI)
- we tried to get the token but we failed (we can reuse the ContentLoadingErrorView here as well)

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
etoledom: Oh, I had a problem where the picker wasn't loading the first time. I thought this could be the problem. It wasn't and I forgot to change it back.

Also, `model.refresh()` is not async, and it opens a task itself, so I think `.task` here is not really necessary 🤔 

Though both options will work just fine.

Sources/GravatarUI/SwiftUI/OAuthSession/Keychain.swift
etoledom: I used password since is what we set in the keychain, though we use it to store an auth token.

On the implementation we use `kSecClass: kSecClassGenericPassword` (on line 22 for example).

Now, since it's for internal usage, I guess we can change to `secret` which I think is actually a more generic word than `password` ¯\_(ツ)_/¯

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
etoledom: Sounds good 👍

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
etoledom: Sure, this was for debugging and see the types of the error thrown to be catch 👍 
`os.log` is a good idea. It might help developers debug issues on their users.

Sources/GravatarUI/SwiftUI/ProfileEditor/ProfileEditor.swift
etoledom: Sounds good!

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
pinarol: Maybe we should make model.refresh() async. 🤔 

What's good about `.task` is, it will automatically cancel the task when the view is destroyed. `.onAppear()` does not have this feature. I think `.task` is slightly a better practice but we'll need to make model.refresh() async so the cancellation is propagated to child tasks I think 🤔. No need to dive into this in this PR though.

pinarol: I created an issue to investigate this https://github.com/Automattic/Gravatar-SDK-iOS/issues/362

etoledom: > What's good about .task is, it will automatically cancel the task when the view is destroyed.

Makes sense 🤔 

> I created an issue to investigate this https://github.com/Automattic/Gravatar-SDK-iOS/issues/362

Nice! Thanks 🙏

Sources/GravatarUI/SwiftUI/ProfileEditor/ProfileEditor.swift
etoledom: @pinarol - I tried this and the Build Conditions from the demo app won't be passed to the package, so there's no access to it from within the SDK.

I decided to move this button to the demo screen, previous to present the picker.

Now we will have a `Log out ` button when we detect a session:

CleanShot 2024-08-22 at 14 28 18@2x

pinarol: I see. That works too. 👍

Sources/GravatarUI/SwiftUI/ProfileEditor/ProfileEditor.swift
```diff
@@ -0,0 +1,58 @@
...
+            performAuthentication()
+        }
+    }
+
+    @MainActor
+    func performAuthentication() {
+        Task {
+            isAuthenticating = true
+            if !oauthSession.hasSession(with: email) {
+                _ = try? await oauthSession.retrieveAccessToken(with: email)
```
etoledom, line 50: The possible errors are still not yet properly handled here, but we can do this alongside building the error screens to present these errors.

pinarol:
Something caught my attention, I see the diff for my secrets file:

Screenshot 2024-08-22 at 15 59 07

Seems like the path doesn't match the paths in the `.gitignore`

etoledom:
@pinarol - I believe this is ready for another look 👀

etoledom:
> Something caught my attention, I see the diff for my secrets file:
> 
> Screenshot 2024-08-22 at 15 59 07
> Seems like the path doesn't match the paths in the `.gitignore`

You can delete this file.

This was an error when we renamed the demo apps. We left this file which is not supposed to be in source control inside the old folder name. Now it shows because I removed its entry from .gitignore.

You get this since this file still exist in your machine, but it's safe to be deleted 👍

Demo/Demo/Gravatar-SwiftUI-Demo/DemoProfileEditorViewView.swift
```diff
@@ -0,0 +1,60 @@
+import SwiftUI
+import GravatarUI
```
pinarol: typo in the filename ? "DemoProfileEditorViewView.swift"

Sources/GravatarUI/SwiftUI/OAuthSession/Keychain.swift
```diff
@@ -0,0 +1,73 @@
...
+protocol SecureStorage: Sendable {
+    func setSecret(_ secret: String, for key: String) throws
+    func deleteSecret(with key: String) throws
+    func secret(with key: String) throws -> String?
+}
+
+struct Keychain: SecureStorage {
+    enum KeychainError: Error {
+        case unexpectedPasswordData
+        case cannotConvertPasswordIntoData
```
pinarol: For consistency, let's replace `password` with `secret` here as well.

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -25,3 +25,16 @@ extension View {
             )
     }
 }
+
+@MainActor
+extension View {
```
pinarol: no need to add a new `extension View` I think? we already have one in the same file.

pinarol (approved):
Looks good, just leaving some nits. I don't need to check again.

pinarol:
> You get this since this file still exist in your machine, but it's safe to be deleted 👍

Oh I see. Cool.

PR #363, opened 2024-08-22
Add option to pick image by using the camera

etoledom:
### Description

Adding the option to upload a new Avatar image by taking a photo using the camera.

*Note*: We are forcing the sources to be used by the Avatar Picker. I'm not sure if we should make it an option to the 3rd party dev, since this is Gravatar domain.

In case we want to give this option, we can do it in a next PR.

### Testing Steps

- Build the SwiftUI demo app on device (or use the App Center build).
- Add a new avatar using the device's camera.
- Check that the new image can be selected as the avatar.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/SystemImagePickerView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/SystemImagePickerView.swift
@@ -11,35 +11,84 @@ struct SystemImagePickerView<Label>: View where Label: View {
         // So far, the new SwiftUI PhotosPicker only supports Photos library, no camera, and no cropping, so we are only using legacy for now.
         // The interface (using a Label property as the element to open the picker) is the same as in the new SwiftUI picker,
         // which will make it easy to change it later on.
-        LegacyImagePicker(label: label, onImageSelected: onImageSelected)
+        ImagePicker(label: label, onImageSelected: onImageSelected)
     }
 }
 
-/// UIImagePickerController SwiftUI Representable wrapper.
-struct LegacyImagePicker<Label>: View where Label: View {
+private struct ImagePicker<Label>: View where Label: View {
+    enum SourceType: CaseIterable, Identifiable {
+        case photoLibrary
+        case camera
+
+        var id: Int {
+            self.hashValue
+        }
+    }
+
     @State var isPresented = false
+    @State private var sourceType: SourceType?
+
     @ViewBuilder var label: () -> Label
     let onImageSelected: (UIImage) -> Void
 
     var body: some View {
         VStack {
-            Button(action: {
-                isPresented.toggle()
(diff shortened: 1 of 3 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/SystemImagePickerView.swift
```diff
@@ -11,35 +11,84 @@ struct SystemImagePickerView<Label>: View where Label: View {
...
     var body: some View {
         VStack {
-            Button(action: {
-                isPresented.toggle()
-            }, label: {
+            Menu {
+                ForEach(SourceType.allCases) { source in
+                    Button {
+                        sourceType = source
+                        isPresented.toggle()
```
pinarol: `toggle()` is nice when we want to switch between "true <-> false" repeatedly on a button tap or something like that. But that's not the case here, the intention is just setting `isPresented = true`, right? 

In general I'd say let's not use `toggle()` in places where we just want to set the value to either "false/true". It can be prone to bugs.

pinarol (commented):
Looks good overall. One thing that caught my attention is that the button shouldn't be appearing that close to the bottom. 

<image src="https://github.com/user-attachments/assets/5f18d8f0-65c9-4200-968f-609bb95881f0" width=300/>

We use `ignoresSafeArea` modifier which looks like working well for system image picker, but not that well for the camera view. The button shouldn't be inside the bottom safe area.

So we either need to,

1. Add bottom padding with the amount of safe area bottom inset. In this case we need to use GeometryReader to get the safe area bottom inset i think 🤔 

OR

2. We do the trick here mentioned [here](https://medium.com/@jpmtech/adding-a-bottom-sheet-or-partial-modal-in-swiftui-b962bd9cb962). Which would look sth like this:

Maybe there are other options too...

```
            case .camera:
                ZStack {
                    Color.black.ignoresSafeArea(edges: .all)
                    LegacyImagePickerRepresentable(sourceType: source.map()) { image in
                        onImageSelected(image)
                    }
                }
```

This can be worth investigating a bit. 🤔

etoledom:
> 2. We do the trick here mentioned [here](https://medium.com/@jpmtech/adding-a-bottom-sheet-or-partial-modal-in-swiftui-b962bd9cb962). Which would look sth like this:

I implemented this and seems to work good, for as long as Apple don't change the camera background color 🙈 

> 1. Add bottom padding with the amount of safe area bottom inset. In this case we need to use GeometryReader to get the safe area bottom inset i think 🤔

This seems to give a similar result to not call `.ignoresSafeArea()`, right? And it leaves an empty space at the bottom.

I'd go with the color fill for now.

etoledom:
@pinarol - Ready for another look 👀

pinarol (approved):
Looks good 👍 🎉

PR #367, opened 2024-08-23
Etoledom/adding navigation bar to profile editor

etoledom:
### Description

This PR adds a navigation bar on the Profile Editor.

The idea is that there's only one `NavigationView` at the root of any flow, and each view defines their own navigation modifiers. In a similar way that controllers modify their own `NavigationBarItem`.

I've created a GravatarNavigationModifier to make this easier and share the structure of our nav bar.

### Testing Steps

- Run the SwiftUI demo app
- Go to Profile Editor example screen
  - Check that the Navigation bar shows as expected.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
@@ -49,7 +49,7 @@ struct AvatarPickerView: View {
         }
         .gravatarNavigation(
             title: Constants.title,
-            actionButtonDisabled: $model.isProfileLoading,
+            actionButtonDisabled: model.profileModel?.profileURL == nil,
             onActionButtonPressed: {
                 openProfileInSafari()
             },
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
@@ -25,51 +25,38 @@ struct AvatarPickerView: View {
     @State private var safariURL: URL?
 
     public var body: some View {
-        NavigationView {
-            ZStack {
-                VStack(spacing: 0) {
-                    email()
-                    profileView()
-                    ScrollView {
-                        errorView()
-                        if !model.grid.isEmpty {
-                            content()
-                        } else if model.isAvatarsLoading {
-                            avatarsLoadingView()
-                        }
-                        Spacer()
-                            .frame(height: Constants.vStackVerticalSpacing)
+        ZStack {
+            VStack(spacing: 0) {
+                email()
+                profileView()
+                ScrollView {
+                    errorView()
+                    if !model.grid.isEmpty {
+                        content()
+                    } else if model.isAvatarsLoading {
+                        avatarsLoadingView()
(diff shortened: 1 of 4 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -25,51 +25,38 @@ struct AvatarPickerView: View {
...
+                    model.refresh()
                 }
             }
+
+            ToastContainerView(toastManager: model.toastManager)
+                .padding(.horizontal, Constants.horizontalPadding * 2)
         }
+        .gravatarNavigation(
+            title: Constants.title,
+            actionButtonDisabled: $model.isProfileLoading,
```
pinarol: Why not keep `model.profileModel?.profileURL == nil` to determine the disabled state? If the profile service fails for some reason, we would want to keep the button disabled since there will be no `profileURL`.

Also we probably doesn't need to pass a `Binding`?

Sources/GravatarUI/SwiftUI/ProfileEditor/ProfileEditor.swift
```diff
@@ -20,24 +24,39 @@ struct ProfileEditor: View {
...
+                        performAuthentication()
                     }
-                } else {
-                    ProgressView()
                 }
+            } else {
+                ProgressView()
             }
-        }.task {
+        }.gravatarNavigation(title: Constants.title, actionButtonDisabled: .constant(true), onDoneButtonPressed: {
```
pinarol: We can put each param on a newline, swiftformat fails us again...

Sources/GravatarUI/SwiftUI/GravatarNavigationModifier.swift
```diff
@@ -0,0 +1,52 @@
+import SwiftUI
+
+struct GravatarNavigationModifier: ViewModifier {
+    @State var title: String
+    @Binding var actionButtonDisabled: Bool
```
pinarol: I think both can be just normal variables?

etoledom: `title` I think yes, but the `actionButtonDisabled` wasn't updating.

I'll give it a second pass to see if I was doing something wrong before 🤔

etoledom: Ok It all work. My mistake was to try with `@State` for `actionButtonDisabled` from the beginning and it wasn't updating. But it being normal variable works just fine 👍

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
etoledom: Yes, that's true. I wen't into a little rabbit hole explained [here](https://github.com/Automattic/Gravatar-SDK-iOS/pull/367#discussion_r1728858840)

But now it should be working more simply.

I wasn't really sure why Binding is needed since don't need to update its value. Turns out it's not needed. 
Thank you for this catch!

etoledom:
@pinarol - Ready for another look

pinarol (approved):
Looks good 🎉

PR #369, opened 2024-08-23
Handle token error

etoledom:
### Description

This PR implements a way to handle an invalid token error:

1. If the authorisation process was made by the SDK, we request the user to log-in again, and show again the OAuth flow.
2. If the token was provided externally, we give the option to the user to close the picker.

The description text for option 2 is not meant to be definitive.

CleanShot 2024-08-23 at 16 15 39

#### Test external token flow:
- Build and run the SwiftUI example app.
- Go to the Avatar picker view example screen.
- Insert a wrong token.
  - Check that a `Close` button is presented on the UI
 
#### Test internal token flow:

- Go to `OAuthSession.swift` line 39
- Generate a wrong token to trigger the expired token error
  -  i.e: ` let token = "2"`
- Build and run the SwiftUI example app
- Go to the Profile Editor example screen.
- **Log Out** if needed. 
- Open the editor
- Accept the authorisation prompt.
  - Check that a `Log in` option is presented to the user.
  - If you try to log in, the oauth flow should be presented
  - The error will appear again.
- Kill the app
-  Restore the original code in `OAuthSession.swift` line 39
- Build and run again.
- Open the Profile Editor **without** pressing on log out.
  - Check that the error is still present.
  - Tap on Log in.
-  Accept the authorisation prompt.
  - Check that now the error is gone, and the picker works as expected.
- THE END.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
@@ -103,8 +103,8 @@ struct AvatarPickerView: View {
                 )
             case .failure(APIError.responseError(reason: let reason)) where reason.httpStatusCode == HTTPStatus.unauthorized.rawValue:
                 let buttonTitle = tokenErrorHandler == nil ? "Close" : "Log in"
-                let subtext = tokenErrorHandler == nil ?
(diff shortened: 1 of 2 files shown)


etoledom:
@pinarol - Updated the text

pinarol (approved):
LGTM 👍

PR #375, opened 2024-08-28
Allow 3rd party apps to inject their existing image editors to the QuickEditor

### LINKED ISSUE #365: Quick Editor: Allow 3rd party apps to inject their own image cropper
WordPress/Jetpack app has its own image cropper for their avatars, since their avatars are circle, the cropper shows a circle frame to help with the cropping. And it is likely other apps can also have one. 

It would be good to allow other apps to use their own image croppers.

pinarol:
Closes #365 

### Description

Allows 3rd party apps to inject their existing image editors to the QuickEditor. This is basically for cropping the image but it doesn't have to be limited to it.

### Testing Steps

SwiftUI Demo app > Avatar Picker view > "Custom image cropper" switch opens a dummy image cropper.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
@@ -8,19 +8,10 @@ public enum AvatarPickerContentLayout: String, CaseIterable, Identifiable {
     case horizontal
 }
 
-private enum AvatarPicker {
-    enum Constants {
-        static let horizontalPadding: CGFloat = .DS.Padding.double
-        static let lightModeShadowColor = Color(uiColor: UIColor.rgba(25, 30, 35, alpha: 0.2))
-        static let title: String = "Gravatar" // defined here to avoid translations
-        static let vStackVerticalSpacing: CGFloat = .DS.Padding.medium
-        static let emailBottomSpacing: CGFloat = .DS.Padding.double
-    }
-}
-
 @MainActor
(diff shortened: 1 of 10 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -8,22 +8,26 @@ public enum AvatarPickerContentLayout: String, CaseIterable, Identifiable {
     case horizontal
 }
 
-@MainActor
-struct AvatarPickerView: View {
-    private enum Constants {
+private enum AvatarPicker {
+    enum Constants {
```
pinarol: I had to move these constants out of AvatarPickerView because the compiler doesn't allow us to define static vars inside generic types.

Sources/GravatarUI/SwiftUI/AvatarPicker/SystemImagePickerView.swift
```diff
@@ -48,20 +50,43 @@ private struct ImagePicker<Label>: View where Label: View {
...
+                            sourceType = nil
+                            onImageSelected(croppedImage)
+                        }
                     }
+                })
+        })
+    }
+
+    @ViewBuilder
+    private func displayPickerOrCamera(_ source: SourceType) -> some View {
```
etoledom: Sources could be other than picker or camera in the future, like the genmoji.

We could call this function something like `displayImagePicker(for source: SourceType)` . Or we could use `selector` instead of `picker` to avoid possible confusion. 🤔

etoledom (approved):
Looking good! 🎉

PR #377, opened 2024-08-29
Quick Editor: Add presentation styles

### LINKED ISSUE #376: Quick Editor: Custom height and detents
The QuickEditor doesn't have to be displayed at full height. We need to support different heights and detents(sticking points). Let's allow passing `presentationDetents` in the SDK.

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/376

### Description

~~We allow passing `presentationDetents`. Unfortunately they are supported only `+ iOS16` so I had to duplicate some code :(~~

Introduces different presentation styles.

**Pre iOS 16**

 - The system doesn't allow us to set the bottom sheet height and we can only display full size sheets. So we only allow for the vertically scrolling grid.

**iOS 16.0 +**

- **Vertical grid**
   Presentation options:
    - **large**: full size sheet
    - **expandableMedium**: starts with a fractional height but can be expanded to full height. the fraction is by default 0.7 but it can be adjusted by the 3rd party developer. The behavior of whether we should prioritize scrolling over expanding is also adjustable.

https://github.com/user-attachments/assets/9bd85995-9312-4d9b-bedd-a22f38c98449

- **Horizontal grid**
   Presentation options:
   - **intrinsicHeight**: as the name tells.

https://github.com/user-attachments/assets/718ceb3d-8adc-4255-b8b4-7b40ebe5862c

------

If the width is not compact(like in iPads) the system displays a full height sheet so we always show the vertical grid:

Simulator Screenshot - iPad mini (6th generation) - 2024-09-05 at 15 13 44

### Testing Steps

SwiftUI demo app > Avatar picker view > select different layout options.


etoledom (commented):
Some thoughts about the behavior of this:

I think this should be managed internally.
The user can select the layout, and the SDK will choose how ti present it.

Ideally we can fit the sheet height content size. This is easy in UIKit, but doesn't seem to be a thing on SwiftUI yet :(
I saw new API on iOS18 to make this easy SwiftUI, but we are far from it.
https://developer.apple.com/documentation/swiftui/view/presentationsizing(_:)

There are tutorials using GeometryReader, though seems to fail when used over scroll views as we do.

I really hope we can achieve this, since it feels as the most normal option used in many apps and the best UX.

Alternatively, we can chose to do 0.7 always with the option to go full screen if the content doesn't fit. 
Or 0.7 on horizontal layout (with the scroll to full screen) while going full screen on vertical layout.

Either way, if we present a sheet that does not fit the content automatically, we must give the option to scroll to full screen.

CleanShot 2024-08-29 at 18 54 05

pinarol:
I think this is ready for review. @etoledom

etoledom (approved):
Thank you for this PR!

As I mentioned a few times before, I'm very surprised that this behavior needs so much work, and still we can't avoid an eventual jump without a smooth animation.

Said that, this is a pretty good workaround, and as long as the estimated height is close enough to the real one, it feels pretty good 👍 

I have it some time trying to figure out a way to animate the jump without success. 

So let's move forward with this, since it is already better than the previous behavior.

PR #380, opened 2024-08-30
Automation for downloading current localizations from GlotPress, Localized Demo Project

### LINKED ISSUE #378: Add support for downloading localizations from GlotPress

### LINKED ISSUE #379: Add localization support to the Demo project
The Demo project needs to support every localization available in the SDK so that we can preview localizations with the Demo apps.  Localizing the UI of the demo apps is not necessary for this to work.  We only need to have support for all languages in the Demo app.

andrewdmontgomery:
Closes #378 
Closes #379 

### Description

Adds:
- Adds the ability to download "current" localizations from GlotPress
- Adds all supported languages to the Demo project (necessary to preview localized SDK views in the Demo apps)
- Refactors the fastlane file to more flexibly handle multiple localizable sources

Fixes:
- All `NSLocalizedString()` calls need to specify `bundle: Bundle.module` in the SDK so that their localizations can be found at run time in the SDK module, rather than the `Bundle.main` which is the default.

Note: The ability to `generate_strings` and `download_localized_strings` are meant to be used during a release.  Developers won't be responsible for handling these steps on PR's.

---
### A note about localization the Demo project
The Demo project now has localization support for all 16 languages that we support with the SDK.  This was necessary to preview localized elements of the SDK from within the Demo apps.

However, we do not currently localize any strings in the Demo project:
- `bundle exec fastlane generate_strings`: this will not add any localizable strings to the Demo project
- `bundle exec fastlane download_localized_strings`: this will not download any localized strings from GlotPress (there is no project on GlotPress for the Demo project)

If we choose to localize the Demo project, the fastlane automation is ready to handle the Demo project as an additional `LocalizableSource`.

### A note about requiring `bundle: .module` in `NSLocalizedString()` calls in the SDK
This presents a problem.  If we forget to specify `.module`, then that string will not be localized when it appears in an app that imports our SDK.

We could enforce this with SwiftLint. We would just have to add it to our repo, etc.  Note that, when we eventually upgrade to String Catalogs, and swap `NSLocalizedString()` with `String(localized:, defaultValue:, bundle:, comment:)` we will have the same requirement, and the same opportunity for unlocalized strings appearing in third-party apps.

I'm open to thoughts on this (@AliSoftware)

---

#### Add strings to code, generate an updated `en` Localizable.strings

1. Add localizable strings to the SDK.  Remember to specify the `bundle: .module` when calling `NSLocalizedString("key", bundle: .module, value: "english string", comment: "a comment")`
2. Update the `en` `Localizable.strings` for the GravatarUI module of the SDK:
   - `bundle exec fastlane generate_strings skip_commit:true` (this will run the command and will skip committing the new strings)
- [ ] OBSERVE: `Sources/GravatarUI/Resources/en.lproj/Localizable.strings` should contain your new string

Revert the change above, and re-run the command without `skip_commit:true`:
`bundle exec fastlane generate_strings`

- OBSERVE:
   - [ ] `Sources/GravatarUI/Resources/en.lproj/Localizable.strings` should contain your new string
   - [ ] A new commit was added:
      - Comment: `Update strings in base locale`
      - Only the `Localizable.strings` file was committed (your source file should not have been included in the commit)

#### Download localized strings from GlotPress
1. Delete the non-en `.lproj` folders from `Sources/GravatarUI/Resources`
2. Download localizations from GlotPress:
   - `bundle exec fastlane download_localized_strings skip_commit:true`
- [ ] OBSERVE: All languages are re-downloaded (note that all languages are currently empty)

Repeat the steps above, only this time, without `skip_commit:true`
- [ ] OBSERVE:
   - All languages are re-downloaded (again)
   - Since there are no changes, there will be no commit.  But note that the command runs without error.

#### Preview a localization using the Demo app
Use these example translations (generated by Google) for the steps below.

```
/* Title of a button that will close the Avatar Picker, appearing beneath a message that advises the user that their login session has expired. */
"AvatarPicker.ContentLoading.Failure.SessionExpired.Close.buttonTitle" = "fermer";

/* A message describing the error and advising the user to login again to resolve the issue */
"AvatarPicker.ContentLoading.Failure.SessionExpired.Close.subtext" = "Désolé, il semble que votre session ait expiré. Assurez-vous d'être connecté pour mettre à jour votre avatar.";

/* Title of a message advising the user that their login session has expired. */
"AvatarPicker.ContentLoading.Failure.SessionExpired.title" = "session expirée";
```

1. Open the Demo Project
2. Add the example text above to `Sources/GravatarUI/Resources/fr.lproj/Localizable.strings`
3. Select the scheme `Gravatar-SwfitUI-Demo`
4. Edit the scheme
5. Under `Run` --> `Options`, set `App Language` to `French`.  Close when finished.
6. Run the app in a simulator
7. Tap `Avatar picker view`
8. Clear your email address and/or token
9. Tap `Tap to open the Avatar Picker`
- [ ] OBSERVE: The view is localized

image

#### To Reset Your Environment
1. Open the Demo Project
2. Edit the `Gravatar-SwiftUI-Demo` scheme
3. Under `Run` --> `Options`, set `App Language` to `System Language`.  Close when finished.

CC'ing AppsInfra: @iangmaia @AliSoftware


andrewdmontgomery:
As always, Ruby feedback and suggestions are always quite welcome. 😁

andrewdmontgomery:
By the way, I still can't get SwiftUI Previews to appear localized.

https://developer.apple.com/documentation/xcode/previewing-localizations

I've tried:
- Clearing derived data
- Deleting everything from: `~/Library/Developer/Xcode/UserData/Previews/Simulator Devices/`
- `killall -9 com.apple.CoreSimulator.CoreSimulatorService`

In my testing, I added the strings above to the SDK's French localization (described above).  Then:
1. Open `AvatarPickerView` source file in Xcode (Either from within the Demo Project, or by opening the Swift Package)
2. Update the `#Preview("Load from network")` code block to include the necessary `environment()` (see example below)
3. Load the preview
4. Click the `Load from network` preview

```swift
#Preview("Load from network") {
    /// Enter valid email and auth token.
    AvatarPickerView(model: .init(email: .init(""), authToken: ""), isPresented: .constant(true))
        .environment(\.locale, .init(identifier: "fr"))
}
```

AliSoftware:
> This presents a problem. If we forget to specify .module, then that string will not be localized when it appears in an app that imports our SDK.
>
> We could enforce this with SwiftLint. We would just have to add it to our repo, etc. Note that, when we eventually upgrade to String Catalogs, and swap NSLocalizedString() with String(localized:, defaultValue:, bundle:, comment:) we will have the same requirement, and the same opportunity for unlocalized strings appearing in third-party apps.
>
> I'm open to thoughts on this (@AliSoftware)

One easy way to enforce this is to:
 - Create a custom free function (e.g. `SDKLocalizedString` or `_T` or however you wanna call it) that would act as a replacement to `NSLocalizedString` but enforcing `Bundle: .module`[^1]
 - Pass it as a custom routine name when calling `genstrings`
 - Enforce the use of `SDKLocalizedString` over `NSLocalizedString` using a SwiftLint rule or similar.

<details><summary>Some inspiration from WordPress-iOS</summary>

We use a similar approach in WPiOS, except that the reason with use a custom routine there is to make the app extensions (e.g. widgets) use the same strings as the main app.
 - [We have an `extension Bundle` providing a `static var app: Bundle` that gets the app bundle even from an app extension](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/WordPress/Classes/Utility/AppLocalizedString.swift#L16-L23)
 - Then [our custom `func AppLocalizedString` calls `Bundle.app.localizedString(forKey: key, value: value, table: nil)`](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/WordPress/Classes/Utility/AppLocalizedString.swift#L66-L68)
 - We then have [an elaborate `swift` script](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/Scripts/BuildPhases/LintAppLocalizedStringsUsage.swift) (invoked via [a `.sh` trampoline script to avoid recompiling it on each Xcode build](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/Scripts/BuildPhases/LintAppLocalizedStringsUsage.sh)) which parses the `.xcodeproj` to find the list of files in the BuildPhase of each target and [warns about `NSLocalizedString` usage](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/Scripts/BuildPhases/LintAppLocalizedStringsUsage.swift#L266-L287) in files belonging to an appex target

That being said, the setup in WPiOS is more complicated that we need it to be for GravatarSDK. The main reason why we created our own `swift` script to lint the files instead of using `swiftlint` is because we had an existing codebase with plenty of `NSLocalizedString` calls and only wanted to enforce the use of the `AppLocalizedString` routine for files that belonged to an appex target, while still being ok for devs and existing code to call `NSLocalizedString` in apps that were only included in the app target. This is why the script was needed to parse the whole `pbxproj` files (as a plist) and navigate in its structure to find the targets and their BuildPhase and the files listed in it to only lint the files of interest but not the others—an advanced filtering we couldn't implement in `swiftlint` alone.

</details>

For Gravatar we don't have as much subtlety as WP nor need to account for legacy/existing codebase, so we might as well use `swiftlint` to enforce it. That being said, part of this can still serve as inspiration, like [the implementation of `AppLocalizedString`](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/WordPress/Classes/Utility/AppLocalizedString.swift) and [how it is passed as a custom `routine` when calling `ios_generate_files_from_code` (our wrapper for `genstrings`) from our `Fastfile`](https://github.com/wordpress-mobile/WordPress-iOS/blob/41b8d0a87111db71b4994619ebc9cf3f0b743d5a/fastlane/lanes/localization.rb#L217).

It is worth noting that once we'll want to migrate from `genstrings` and legacy `.strings` files to `xcodebuild -exportLocalizations` and String Catalogs, iinm we won't be able to use a custom routine though.
Maybe at that later point we could come up with:
 - Either a more involved `swiftlint` rule that would directly validate the `bundle: .module` parameter passed to our future uses of `String(localized:defaultValue:bundle:comment:)` (could become a bit complicated to write with a single regex to account for e.g. multiline call sites though)
 - Or write a custom swift script that leverages SourceKit or SwiftSyntax (whichever would fit best, not sure) to parse the code at token level and thus be more resilient at detecting the calls to `String(localized:…)` in the source code (and the value of the parameters passed to that call), especially for multiline calls. Feels a bit too involved and maybe "bulldozer to kill a fly" kind of solution for the needs though. But could be a cool HACK week project too… 🙃 
 - Or maybe better, leverage Swift Macros, and create a `#L10n` / `#Localized` macro that would expand to `String(localized:…, bundle: .module, …)` automatically, making it impossible to pass the wrong `bundle` parameter. It could then be easier to enforce its usage via `swiftlint` (which I'm pretty sure would run on source code _before_ macro resolution). This would only work providing that the `xcodebuild -exportLocalizations` works on the code post-macro-expansion and not before, so that it would still be able to consider the usage of those `#Localized` macros as calls to `String(localized:…)` needing to be exported.

In any case, we can think about this once we get there? (It's pretty new territory at that point, so I'm ok with GravatarSDK being the test bed for experimenting with tooling ideas on this)

[^1]: As a side note, when implementing a free function for use as a custom routine for `genstrings`, be sure to have it implemented by calling `Bundle.localizedString(…)` method and not calling `NSLocalizedString`. This is because `NSLocalizedString` is specifically dedicated to only be used as a macro/wrapper to be detected by localization extraction automation (be it `genstrings` or `xcodebuild -exportLocalizations`) and should only accept a string literal constant as its `key` parameter, so forwarding the `key` parameter/variables via `NSLocalisedString(key, …)` from `func SDKLocalizedString(key: String, …)` would be an invalid usage.

fastlane/lib/localizable_source.rb
```diff
@@ -0,0 +1,68 @@
...
+#     localizations_root: '/path/to/localizations',
+#     gp_project_url: 'https://example.com/project'
+#   )
+#
+#   puts localizable.base_localization_strings
+#   # Output: /path/to/localizations/en.lproj/Localizable.strings
+#
+#   puts localizable.base_localization_strings(base_locale: 'fr')
+#   # Output: /path/to/localizations/fr.lproj/Localizable.strings
+#
```
AliSoftware: I love the thorough documentation here.

That being said, I'd suggest to use the YARD documentation convention to write such documentation. e.g.:

```ruby
# The `LocalizableSource` class represents a localizable source for generating
# strings and downloading localizations from GlotPress.
#
# Example:
#
#   localizable = LocalizableSource.new(
#     …
#    …
#
class LocalizableSource
  # [Array<String>] An array of paths where source files related to localization are stored.
  # This can include multiple directories where localized resources are kept.
  attr_accessor :source_paths

  …
  # Creates a new instance of `LocalizableSource`.
  # @param source_paths [Array<String>]An array of paths where source files are located.
  #        This argument is required and cannot be `nil`.
  # @param localizations_root [String] The root path for localization files.
  #        This argument is required and cannot be `nil`.
  # @param gp_project_url [String, nil] The URL to the GlotPress project or similar.
  #        This argument is optional.
  # @raise ArgumentError if either `source_paths` or `localizations_root` is `nil`.
  def initialize(source_paths:, localizations_root:, gp_project_url: nil)
    …
  end

  # Generates the file path to the `Localizable.strings` file for a given locale.
  # @param base_locale [String] The locale code (e.g., 'en', 'fr', 'es') for which to generate the path. Defaults to `'en'`.
  # @return [String] The full path to the `Localizable.strings` file for the specified locale.
  def base_localization_strings(base_locale: 'en')
    …
  end
```

fastlane/lib/lanes/localization.rb
```diff
@@ -1,5 +1,50 @@
 # frozen_string_literal: true
 
+require_relative '../localizable_source'
+
+default_platform(:ios)
+
+UI.user_error!('Please run fastlane via `bundle exec`') unless FastlaneCore::Helper.bundler?
```
AliSoftware: I like this safety net, but I'd expect this to be already tested by the root `Fastfile` that `require_relative`s this `.rf` file, and thus not needed to repeat this test here?

fastlane/lib/lanes/localization.rb
```diff
@@ -8,69 +53,91 @@
...
-            File.join('Sources', 'GravatarUI')
-          ],
+          paths: source.source_paths,
           output_dir: tempdir
         )
-  
+
         utf16_strings = File.join(tempdir, 'Localizable.strings')
-        utf8_strings = File.join("..", demo_en_lproj, 'Localizable.strings')
+        utf8_strings = File.join('..', source.base_localization_strings)
```
AliSoftware: A side note that, technically speaking, `genstrings` (and thus `ios_generate_strings_file_from_code`) can potentially generate more than one file. In particular, if the code uses `NSLocalizedString(key, value: …, table: "Other", bundle: .module)`, then it will generate an `Other.strings` alongside the `Localizable.strings` (which is the default table name).

So if you want to be technically correct, you might want to run your `utf16_to_utf8` helper on each `.strings` file found in the `tempdir`, not just `Localizable.strings`. That would also mean that you'd need to revisit your assumption about `def base_localization_strings` being useful there vs it being too locked with the assumption of `Localizable.strings`, or change that into a `def base_lproj` helper that only builds the path up to the `.proj`, or something similar.

That being said, only supporting `Localizable.strings` file and the default `Localizable` table name would probably be enough at least at that stage. Besides, supporting multiple `.strings` files would need more work anyway (e.g. you'd need separate projects in GlotPress for each `<tablename>.strings` file, or as a work around if that's not suitable you'd need to merge the content of all the `*.strings` file into one before uploading it in GlotPress, then split the result back into individual files when pulling them back (which is what we did for WP/JP iOS so that we wouldn't need a separate GlotPress project just for other `.strings` files in our project like the `InfoPlist.strings` file, or the `.strings` files of our app extensions or Siri intents… but that leads to complex merging and splitting logic and is better to be avoided if not needed like here)

fastlane/lib/lanes/localization.rb
```diff
@@ -8,69 +53,91 @@
...
+    next if skip_commit
+
+    git_commit(
+      path: paths_to_commit,
+      message: 'Update strings in base locale',
+      allow_nothing_to_commit: true
+    )
+  end
+
+  private_lane :utf16_to_utf8 do |options|
```
AliSoftware: This is totally opinionated and there's no super strong rule here, but for this helper I think it'd be fine to write it as a regular Ruby function (`def`) rather than a `private_lane`.

The only benefit of writing it as a `private_lane` over `def` is, I think, that `private_lane` being "visible" to the Fastlane runner, the calls to this `private_lane` will appear in the fastlane output (`Cruising to lane: utf16_to_utf8`, then `Cruising back to lane generate_strings`)… which in our case would not be super useful (and could actually be considered noise).

Of course if you change it from `private_lane … do` to `def`, don't forget to also change the calls to `next` with `return` instead, since `private_lane` takes blocks for their body while functions are… well, just functions.

And if you decide to still keep it as a `private_lane`, I'd suggest to at least [use the named keyword syntax for the parameters of the lane rather than accepting an arbitrary Hash of `|options|`](https://github.com/fastlane/fastlane/pull/21587), to make it clearer and self-documented which parameters it accepts.

fastlane/lib/lanes/localization.rb
```diff
@@ -8,69 +53,91 @@
...
+
+    File.open(source, 'rb:UTF-16') do |in_file|
+      utf16_content = in_file.read
+      utf8_content = utf16_content.encode('UTF-8')
+
+      File.open(destination, 'w:UTF-8') do |out_file|
+        out_file.write(utf8_content)
       end
     end
   end
```
AliSoftware: In other words, writing this as a regular Ruby function should look like this, iinm (writing from GitHub, Ruby syntax not validated in a REPL, so I'll let you validate it works):

```suggestion
  def utf16_to_utf8(source:, destination:)
    return unless File.exist?(source)

    utf16_content = File.read(source, 'rb:UTF-16') do |in_file|
    utf8_content = utf16_content.encode('UTF-8')
    File.write(destination, utf8_content, 'w:UTF-8')
  end
```

fastlane/lib/lanes/localization.rb
andrewdmontgomery: Yeah, I think this would be better as a function.  The visibility wasn't the goal, I was debugging a path problem (that was resolved with the `..` bit).  This is really an implementation detail.  I'll make the swap.

fastlane/lib/lanes/localization.rb
andrewdmontgomery: Nice.  I'm not able to use the `File.read(source, 'rb:UTF-16')` (or `File.read(source, encoding: 'UTF-16')`) syntaxes because these files are binary (unless there's a way to use them that I'm not aware of).

fastlane/lib/lanes/localization.rb
andrewdmontgomery: Oh my, good call out.

I refactored the `generate_strings` lane to handle arbitrary Tables.  But as you pointed out, downloading translations for tables would require some rethinking, at some point in the future.

fastlane/lib/lanes/localization.rb
AliSoftware: What about `File.read(source, mode: 'rb:UTF-16')`, or `File.read(source, binmode: true, encoding: 'UTF-16')`? 🤔  (and similarly `File.write(destination, utf8_content, mode: 'w:UTF-8')`)

andrewdmontgomery:
> It is worth noting that once we'll want to migrate from genstrings and legacy .strings files to xcodebuild -exportLocalizations and String Catalogs, iinm we won't be able to use a custom routine though.

I forgot to reply to this.  I was originally tempted to re-use the custom function and the `routine`'s of `genstrings`.  But I paused for the reason you've mentioned here.

I love the idea of a Swift Macro.  It could potentially start life as a substittute for `NSLocalizedString()` and later be refactored for `String()`.

Whatever we do, I'll do it in a follow up PR.

fastlane/lib/lanes/localization.rb
andrewdmontgomery: phew, that reads so much better.  Thanks.  I was tired but also unwilling to stop on Friday evening. 😁

andrewdmontgomery:
Alright, this should be ready for one last set of 👀

@AliSoftware

andrewdmontgomery:
> I love the idea of a Swift Macro. It could potentially start life as a substittute for NSLocalizedString() and later be refactored for String().

Ah, except I believe `genstrings` acts on raw source files.  So macros won't be expanded.

pinarol:
> One easy way to enforce this is to:
> Create a custom free function (e.g. SDKLocalizedString or _T or however you wanna call it) that would act as a replacement to NSLocalizedString but enforcing Bundle: .module`[1](https://github.com/Automattic/Gravatar-SDK-iOS/pull/380#user-content-fn-1-be4883e53e472a8b587c477c2246d0ca)
> Pass it as a custom routine name when calling genstrings
> Enforce the use of SDKLocalisedString over NSLocalizedString using a SwiftLint rule or similar.

Yep. This seems like the way to go. 👍

pinarol (approved):
LGTM 👍 We can tackle this on a separate PR I think. https://github.com/Automattic/Gravatar-SDK-iOS/pull/380#issuecomment-2328220523

AliSoftware:
> Or write a custom swift script that leverages […] SwiftSyntax […] to parse the code at token level and thus be more resilient at detecting the calls […] in the source code (and the value of the parameters passed to that call), especially for multiline calls. Feels a bit too involved and maybe "bulldozer to kill a fly" kind of solution for the needs though. But could be a cool HACK week project too…

[I actually ended up being nerd-sniped by this idea (of playing with SwiftSyntax to lint code)](https://github.com/AliSoftware/swift-syntax-linter-demo) and was pleasantly surprised how easy it ended up being to implement such a solution after all. All it took was [creating a subclass of `SyntaxVisitor`, override its `func visit(_ node: FunctionCallExprSyntax)`, then check the visited `node.calledExpression` and `node.arguments`](https://github.com/AliSoftware/swift-syntax-linter-demo/blob/main/Sources/NSLocalizedStringLinter/NSLocalizedStringVisitor.swift).

So just leaving that here for reference, in case we want to use it as inspiration to use as a custom linter to enforce `bundle: .module` in our `String(localized: …)` calls once we migrate away from `NSLocalizedString`, `.strings` files and `genstrings` in a later date.

_But in the meantime, as suggested and agreed above, while we're still using `genstrings`, a custom `SDKLocalizedString` routine—which would already take care of passing the appropriate `bundle: .module` internally—and a custom swiftlint rule—to enforce its use everywhere in the SDK source code—would be enough and way simpler indeed. So this POC is only for potential inspiration for later._

fastlane/lib/localizable_source.rb
```diff
@@ -0,0 +1,59 @@
...
+class LocalizableSource
+  # @return [Array<String>] An array of paths where source files related to localization are stored.
+  # This can include multiple directories where localized resources are kept.
```
AliSoftware: 💄  IMHO the terms "source files related to localization" and "localization resources" mades me think more about `.strings` files/resources than the `.swift` source files, so suggesting something like this instead:
```suggestion
  # @return [Array<String>] An array of paths to source files we want to scan for localization are stored.
  # This can include multiple directories where localized resources are kept.
```

fastlane/lib/localizable_source.rb
```diff
@@ -0,0 +1,59 @@
...
+  attr_accessor :source_paths
+
+  # @return [String] The root directory for localization files.
```
AliSoftware: 💄  doc nitpick
```suggestion
  # @return [String] The parent directory containing the `.lproj` folders.
```

fastlane/lib/localizable_source.rb
```diff
@@ -0,0 +1,59 @@
...
+  # If `nil`, base localization files can be generated, but no localizations can be downloaded.
+  attr_accessor :gp_project_url
+
+  # Initializes a new LocalizableSource instance.
+  #
+  # @param source_paths [Array<String>] The paths to the source files.
+  # @param localizations_root [String] The root directory for localization files.
+  # @param gp_project_url [String, nil] Optional URL to the translation management system or project repository.
+  # @raise [ArgumentError] if source_paths or localizations_root is nil.
+
```
AliSoftware: 💄 nitpick (YARD convention)
```suggestion
  #
```

fastlane/lib/localizable_source.rb
```diff
@@ -0,0 +1,59 @@
...
+  # Constructs the path to the base localization strings file.
+  #
+  # The method combines the localization root directory with the locale and table name to generate the path
+  # to the base localization strings file.
+  #
+  # @param table_name [String] The name of the table for localization strings. Defaults to "Localizable".
+  # @param base_locale [String] The base locale for localization. Defaults to 'en'.
+  # @return [String] The path to the base localization strings file.
+
```
AliSoftware: 💄 nitpick (YARD convention)
```suggestion
  #
```

fastlane/lib/lanes/localization.rb
```diff
@@ -8,69 +51,92 @@
...
+      ios_download_strings_files_from_glotpress(
+        project_url: source.gp_project_url,
+        locales: GLOTPRESS_TO_LPROJ_APP_LOCALE_CODES,
+        download_dir: source.localizations_root
+      )
+
+      next if skip_commit
+
+      git_add(path: source.localizations_root)
+      paths_to_commit << source.localizations_root
```
AliSoftware: 💡  This works… though an alternative approach (that I'd personally find simpler to read?) would be to not do the `git_add` here, and instead unconditionally add `paths_to_commit << source.localizations_root` inside the loop, and only do the `git_add(path: paths_to_commit)` outside the loop on line 79 (i.e. after the `next if skip_commit` that is outside the loop and just before the `git_commit` call) 🤷

fastlane/lib/lanes/localization.rb
```diff
@@ -8,69 +51,92 @@
...
+        process_generated_strings(
+          source: source,
+          generated_strings_dir: tempdir
         )
       end
+
+      next if skip_commit
+
+      git_add(path: source.localizations_root)
+      paths_to_commit << source.localizations_root
```
AliSoftware: Same idea to add `paths_to_commit << source.localizations_root` unconditionally and only do the `git_add(path: paths_to_commit)` outside the loop just before the call to `git_commit` 🤷‍♂️

fastlane/lib/lanes/localization.rb
```diff
@@ -8,69 +51,92 @@
...
-  
\ No newline at end of file
+
+  def utf16_to_utf8(source:, destination:)
+    return unless File.exist?(source)
+
+    utf16_content = File.read(source, mode: 'rb:UTF-16')
+    utf8_content = utf16_content.encode('UTF-8')
+    File.write(destination, utf8_content, mode: 'w:UTF-8')
+  end
```
AliSoftware: 💄  It took me a moment to understand what type the `source:` parameter was supposed to be; at first I thought it was supposed to be the source _file_ to process, feeling similar to the `source:` parameter of `utf16_to_utf8(source:, destination:)`. Then I realized it was supposed to be a `LocalizableSource` instance instead.

May I suggest to add a quick YARD doc comment to mention the expected type of the input parameters for those helpers then, to help future readers?

💭  Also, while at it: this is totally opinionated, but at that point I wonder if having `utf16_to_utf8` helper as a separate function is worth it, compared to move the 3 LoCs of its implementation directly into `process_generated_strings` 🤔  wdyt?

fastlane/lib/lanes/localization.rb
andrewdmontgomery: Much better.  Not sure what I was thinking with that.

fastlane/lib/lanes/localization.rb
andrewdmontgomery: Yeah, the more I look at this the less I like it.  The implementation of the conversion implicitly requires you to use a temporary directory of generated strings.

I'm going to refactor this.  I think `genstrings` should drop the UTF-16 files where they belong.  Then we can "process" them "in place", using a temp directory, and copy the re-encoded files back under the hood.  No need for the caller to stage their `genstrings` output in their own temp directory, etc.

And then I can just `process_generated_strings(source:)`

andrewdmontgomery:
> But in the meantime, as suggested and agreed above, while we're still using genstrings, a custom SDKLocalizedString routine—which would already take care of passing the appropriate bundle: .module internally—and a custom swiftlint rule—to enforce its use everywhere in the SDK source code—would be enough and way simpler indeed. So this POC is only for potential inspiration for later.

I have another branch that introduces `SDKLocalizedString()`.  I'll have a PR up for that shortly.

fastlane/lib/lanes/localization.rb
andrewdmontgomery: I cleaned this up, as I described.  I ended up adding some error-catching.  As a result, rubocop ended up giving me a 👎 for ABC complexity.  So I refactored it into smaller, single-purpose functions.

See what you think.

fastlane/lanes/localization.rb
```diff
@@ -0,0 +1,183 @@
...
+  # @param tempdir [String] temp directory for storing the encoded file
+  # @return [String, nil] returns the path to the converted file if the conversion succeeds, or `nil` if the file is not UTF-16
+  # @raise [StandardError] if a general error occurs during file conversion
+  #
+  def convert_file_to_utf8(strings_file:, tempdir:)
+    utf8_strings_file = File.join(tempdir, File.basename(strings_file))
+    possible_utf16_content = File.read(strings_file, mode: 'rb:UTF-16')
+    utf8_content = possible_utf16_content.encode('UTF-8')
+    UI.message("Converting: #{strings_file}")
+    File.write(utf8_strings_file, utf8_content, mode: 'w:UTF-8')
```
AliSoftware: The call to `encode` is not really necessary here: `File.write` already takes care of converting the content you tell it to write (`utf8_content`) from the internal encoding that said content string is using to the destination encoding you're asking to use in `File.write` (UTF-8).

 - Here you're converting the internal representation of `possible_utf16_content` from an internal representation of a string that was using UTF16, to the same string but using UTF-8 for its internal representation. And then you're telling it to `File.write` that string that is now already using UTF-8 internally, which means `File.write` doesn't need to do a conversion.
 - But without the explicit call to `encode`, `File.write` would already convert the string you're telling to write from the encoding it is using internally for its internal representation (UTF-16 if that string was read from a file using `File.read(…, mode: 'rb:UTF-16')`, as this `mode` tells Ruby to use UTF-16 as both the external encoding—the one to use to interpret the file content—and as internal encoding—the encoding Ruby will use internally to store the bytes of the string) and convert that string (whatever it uses for its internal_encoding) to the requested external encoding requested by `File.write(…, mode: …)` (UTF-8 here).

So TL;DR: `File.write` already handles the encoding conversion for you when you provide it explicitly in `mode:` or `encoding:` parameter.

fastlane/lanes/localization.rb
```diff
@@ -0,0 +1,183 @@
...
+  # Converts a UTF-16 `.strings` file to UTF-8 and writes it to the specified path.
+  #
+  # @param strings_file [String] the path to the original UTF-16 `.strings` file
+  # @param tempdir [String] temp directory for storing the encoded file
+  # @return [String, nil] returns the path to the converted file if the conversion succeeds, or `nil` if the file is not UTF-16
+  # @raise [StandardError] if a general error occurs during file conversion
+  #
+  def convert_file_to_utf8(strings_file:, tempdir:)
+    utf8_strings_file = File.join(tempdir, File.basename(strings_file))
+    possible_utf16_content = File.read(strings_file, mode: 'rb:UTF-16')
```
AliSoftware: Tip: You can use `BOM|UTF-16` instead of just `UTF-16` to tell Ruby to first try to use the Byte-Order-Mark at the start of the file to determine the encoding of the file being read (UTF8, UTF16LE, UTF16BE), and fallback to `UTF-16` if there was no BOM to help it guess.

fastlane/lanes/localization.rb
```diff
@@ -0,0 +1,183 @@
...
+    end
+  end
+
+  # Converts a file from UTF-16 to UTF-8 using a temp directory to ensure that the conversion is successfull
+  # before overwriting the original file
+  #
+  # @param strings_file [String] path to the `.strings` file to convert
+  # @return [void]
+  #
+  def convert_file(strings_file:)
```
AliSoftware: nitpick: since this is modifying the files in-place (well it's using a temp dir as intermediary, but it's still replacing the original file at the end), Ruby conventions would suggest to use `!` as a suffix in the function name to denote that (as in: `def convert_file!(strings_file:)`

fastlane/lanes/localization.rb
```diff
@@ -0,0 +1,183 @@
...
+  #
+  # @param source [LocalizableSource] An object that represents a localizable source
+  # @return [void]
+  #
+  # @example Convert all `.strings` files from UTF-16 to UTF-8.
+  #   convert_generated_strings(
+  #     source: LocalizationSource.new(source_paths: ['/source/path'], localizations_root: 'Localizations'),
+  #   )
+  #
+  def convert_generated_strings(source:)
```
AliSoftware: Same Ruby convention nitpick given this modifies the input files inplace: `def convert_generated_strings!(source:)`

fastlane/lanes/localization.rb
```diff
@@ -0,0 +1,183 @@
...
+  end
+
+  # Converts a file from UTF-16 to UTF-8 using a temp directory to ensure that the conversion is successfull
+  # before overwriting the original file
+  #
+  # @param strings_file [String] path to the `.strings` file to convert
+  # @return [void]
+  #
+  def convert_file(strings_file:)
+    Dir.mktmpdir do |tempdir|
```
AliSoftware, line 130: I get why you refactored it this way with many small functions due to `rubocop` being unhappy, but this design means that we'll create as many temp dirs as there are files to convert, instead of creating a single temp dir and do the conversions of all our files to be processed in that single dir. Not that big a deal, but not ideal either? 🤷

AliSoftware (commented):
Wow the error handling is very detailed now, trying to catch all the type of errors even the most unlikely ones 😅  I like the effort and the commitment to good error handling and details… but that also makes the implementations and the splitting in minimalistic helpers quite verbose…? Though this might just be a personal opinion.

---

@andrewdmontgomery  I've prepared https://github.com/wordpress-mobile/release-toolkit/pull/591 on the side to support a custom output encoding to be specified in the call to `ios_generate_strings_file_from_code` and handle the conversion implementation directly in the action, which would hopefully help then drastically simplifying the call site by not having to implement such logic in the repo's `Fastfile` itself.

What I'd suggest to move forward:
 - Either remove the custom logic for encoding conversion from this PR for the time being (and accept that the generated `.strings` files will be UTF-16 for the time being, pending the next steps below), or keep it around and don't really take my feedback about the code into account (since we'll remove that logic in next steps)
 - Merge this PR, so we can move forward
 - Prepare a separate PR to adopt the new version of `release-toolkit` (removing the custom logic for encoding conversion there if you ended up not removing it in this PR), so you can just use the new `output_encoding: 'UTF-8'` parameter. At first it can point to the WIP branch from https://github.com/wordpress-mobile/release-toolkit/pull/591 and be a Draft[^1]
 - Once PR https://github.com/wordpress-mobile/release-toolkit/pull/591 has landed, we can ship a new version of `release-toolkit` and you can update your new PR to point to it instead of the WIP branch.

WDYT?

[^1]: Note that currently this repo uses `release-toolkit 11.x` while we shipped a `12.0` since, so what is pending in https://github.com/wordpress-mobile/release-toolkit/pull/591 will ship in a version `12.1`; this means you'll have to ensure the breaking changes introduced in version `12.0` of `release-toolkit` won't affect any action you use in the `Fastfile` of `Gravatar-SDK-iOS` before migrating to `12.x`; though I don't expect any of [the breaking changes from 12.x (listed in this `MIGRATION.md`)](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md) to affect this repo though

fastlane/lanes/localization.rb
andrewdmontgomery: whoa, cool.

fastlane/lanes/localization.rb
andrewdmontgomery: Ah yes, thanks.

fastlane/lanes/localization.rb
andrewdmontgomery: Yeah, I went back and forth (commit: [17ad43f53d](https://github.com/Automattic/Gravatar-SDK-iOS/pull/380/commits/17ad43f53dee12ffa769b43796a89150ad94ad24)).

I was trying to push the implementation detail down.  But really the whole thing just needs to be encapsulated in the call site: `convert_generated_strings!()`.

Yeah, I'll revert that commit and move that tempdir back up.

andrewdmontgomery:
> Wow the error handling is very detailed now, trying to catch all the type of errors even the most unlikely ones 😅 I like the effort and the commitment to good error handling and details… but that also makes the implementations and the splitting in minimalistic helpers quite verbose…? Though this might just be a personal opinion.

`rescue Encoding::InvalidByteSequenceError` --> This is actually the reason I introduced error handling.

At the moment, our `Demo` project is localized, but we are not actually localizing any strings in the Demo app source code.  This was necessary in order to allow the Demo app to show the localized views of the SDK.

This means that:
- We do not generate a new `.strings` file for the Demo project (which means it remains in the default encoding: UTF-8)
- But we are still attempting to read the file as UTF-16 when we convert all `.strings` files to UTF-8
- So the encoding/writing of that file throws an error

```shell
lanes/localization.rb:159:in `write': [!] incomplete "\n" on UTF-16 (Encoding::InvalidByteSequenceError)
```

Since I was breaking up this function into single-purpose functions, and since I was already adding basic error handling, I figured there was room to be consistent.  But you're right; handling the file copy error is overkill.  I'm just catching it to write a fastlane error.  That's probably not very helpful.

> ... or keep it around and don't really take my feedback about the code into account (since we'll remove that logic in next steps)

Yeah, I like this. Thanks for the work on the release-tools.  Your feedback has been very helpful for me, thank you.  I did implement your feedback.  It's better.

I'll merge this, since it's working.  And I opened an issue (#389) to update our release-tooling to 12.0, in preparation for 12.1.

PR #382, opened 2024-09-04
Update openapi spec

etoledom:
### Description

This PR updates the open api spec.

Some changes to note:

- Fetch Identity is not longer required, so it has been removed.
- Avatars request now returns the selected avatar. This replaces the identity fetch.
- Image upload error model for status 400 (i.E: `uncropped_image` and `unsupported_image`).
  - The UI has been updated to not prompt a retry in these cases. A cross will be shown which will remove the failed image when tapped. 
- Avatar selection endpoint has been updated to the latest spec. 

CleanShot 2024-09-20 at 17 49 50

*Note*: I don't particularlt like that all the openapi structs are public. Many of these we use them internally only.

### Testing Steps

Test all avatar picker flows:
- Loading avatars.
- Selecting different avatar.
- Upload avatar.

Reproduce upload avatar error:
- This will reproduce an image-not-square error
  - Go to `AvatarPickerViewModel.swift:248`
    - Comment the `guard` statement.  
  - Go to `AvatarPickerViewModel.swift:261`
    - Change one of the sizes (adding an extra number, for example:
      -  `let squareSize = CGSize(width: squareSide, height: squareSide + 10)`
  -  Build and run
  - Try to upload an image.
    - Check that the error appear and there's no re-upload option.
    - Tap on the X on the avatar, the avatar should disappear.
- Check a different error as uploading avatar without internet connection.
  - Check that re-upload option is present.


andrewdmontgomery:
> Note: I don't particularly like that all the openapi structs are public. Many of these we use them internally only.

I believe we can accomplish this.  We would update the `make generate` command:

1. Run `run-in-docker.sh generate -i openapi.yaml \
    --global-property models...` as we currently do
2. Run it again, but this time:
   - `--global-property models="InternalStruct1:InternalStruct2"`
   -  `--additional-properties=nonPublicApi=true`

I'll put up a draft PR to demonstrate.

pinarol:
The UI changes in this PR is outdated. And the specs might be outdated also. It can be easier to make a separate PR just with the open-api spec changes. 🤔

pinarol (approved):
Okay it seems like the specs are up to date and that's the most important part. So I'd like to merge the PR and make the UI changes on top of that.

pinarol:
I checked with @etoledom and I'll push the merge button.

PR #385, opened 2024-09-05
Remove async from ImageCaching

kean:
This PR removes the `async` requirement from the `ImageCaching` protocol so that the clients could use it from synchronous contexts.


kean:
Converting to draft as the tests implement the caching protocol using an actor, which has to be asynchronous.

pinarol:
Fixed the CI issue here https://github.com/Automattic/Gravatar-SDK-iOS/pull/386, so i am closing this one in favor of that PR.

PR #387, opened 2024-09-06
Use `ImageCache.shared` as the default value

pinarol:
Closes # 

### Description

ImageDownloadService defaults to `ImageCache()` when the passed cache is `nil`. It should be `ImageCache.shared`. Fortunately we pass here `ImageCache.shared` in many places so there's no real bug reported about this. But this is still worth a patch release.

### Testing Steps

Smoke testing the UIKit demo app.

### DIFF
--- a/Sources/Gravatar/Network/Services/ImageDownloadService.swift
+++ b/Sources/Gravatar/Network/Services/ImageDownloadService.swift
@@ -17,7 +17,7 @@ public struct ImageDownloadService: ImageDownloader, Sendable {
     ///   - cache: A type which will perform image caching operations.
     public init(client: HTTPClient? = nil, cache: ImageCaching? = nil) {
         self.client = client ?? URLSessionHTTPClient()
-        self.imageCache = cache ?? ImageCache()
+        self.imageCache = cache ?? ImageCache.shared
     }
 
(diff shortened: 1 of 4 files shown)


Sources/Gravatar/Network/Services/ImageDownloadService.swift
```diff
@@ -17,7 +17,7 @@ public struct ImageDownloadService: ImageDownloader, Sendable {
     ///   - cache: A type which will perform image caching operations.
     public init(client: HTTPClient? = nil, cache: ImageCaching? = nil) {
         self.client = client ?? URLSessionHTTPClient()
-        self.imageCache = cache ?? ImageCache()
+        self.imageCache = cache ?? ImageCache.shared
```
pinarol, line 20: I opened a separate issue to make ImageCache's init private so same thing won't happen again: https://github.com/Automattic/Gravatar-SDK-iOS/issues/388 
It is technically a breaking change so I didn't want to include here.

etoledom (approved):
Good catch!

Any particular reason the OR is on a Draft?

pinarol:
> Any particular reason the OR is on a Draft?

Nope. Just forgotten.

PR #390, opened 2024-09-06
Add `SDKLocalizedString` global function

andrewdmontgomery:
Closes #

### Description

This PR introduces a new way to handle localized strings: `SDKLocalizedString(::::)`.  It is meant to be used instead of the equivalent `NSLocalizedString(::::)` calls.

This work was inspired by a similar solution in the WordPress-iOS repository: [AppLocalizedString](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/WordPress/Classes/Utility/AppLocalizedString.swift)

#### The problem
Every `NSLocalizedString()` instance in the SDK must be called with `bundle: .module` set explicitly.  Any string that doesn't explicitly set that will fail to localize:
- The string will be added to the `.strings` file
- That string will then be localized
- And we will download those localizations to the SDK
- But any app that uses our SDK will search for the localizations in the default location: the `.main` app bundle.  It obviously won't find our localizations there.  So the string will use the default value: English

This global function wraps that setting, making sure that every localized string specifies `.module`.

#### The implementation

This PR  introduces three changes:
- adds a global function in the `GravatarUI` module for handling localized strings: `SDKLocalizedString(::::)`
- updates the call to `genstrings` to include this `SDKLocalizedString` function call as a "routine", which will cause it to use the function to generate strings for the `.strings` file
- Introduces SwiftLintPlugin, with a rule that will raise a build error if `NSLocalizedString` is used in the SDK

---

#### Only until we support String Catalogs
We only need this while we are using `.strings` files that are generated by `genstrings`.  When we move to String Catalogs, we will also move to using `String(localized::::)` instead of `NSLocalizedString(::::)`, and we will stop using `genstrings` to generate our strings.

#### Added to `GravatarUI` only (for now)
I added this global function to `GravatarUI` because this is the only place we need it currently.  But if we decide to add more modules with UI strings, we could move this to `Gravatar` and set the access to `package`.

#### Linting in CI
I was going to add SwiftLint to the CI linting step.  But since SwiftLint is being called during a build, it is already being called in the linting step because `make lint` is calling `swiftlint`, which apparently causes build tools to run.

#### Specifying `.macOS(.v12)` in the Package.swift
To make our Package compatible with SwiftLint, I needed to explicitly configure the minimum version of macOS to `.macOS(.v12)`:

```swift
platforms: [
    // Platforms specifies os version minimums. It does not limit which platforms are supported.
    .iOS(.v15),
    .macOS(.v12)  // The SDK does not support macOS, this satisfies SwiftLint requirements
]
```

Originally, we thought that this line specifies that our Package supports macOS.  But that was a misunderstanding.  The `platforms:` parameter specifies a _minimum version_.  There is no way to limit a Package to a specific platform.

You can see that implicitly.  Try removing the `.macOS(.v12)` platform from the list, and then run `make swiftformat`.  You will get the following error:

```shell
error: the library 'Gravatar' requires macos 10.13, but depends on the product 'SwiftLintBuildToolPlugin' which requires macos 12.0; consider changing the library 'Gravatar' to require macos 12.0 or later, or the product 'SwiftLintBuildToolPlugin' to require macos 10.13 or earlier.
error: the library 'GravatarUI' requires macos 10.13, but depends on the product 'SwiftLintBuildToolPlugin' which requires macos 12.0; consider changing the library 'GravatarUI' to require macos 12.0 or later, or the product 'SwiftLintBuildToolPlugin' to require macos 10.13 or earlier.
error: fatalError
```

> error: the library 'Gravatar' requires macos 10.13 ...
> error: the library 'GravatarUI' requires macos 10.13 ...

If we do not specify a minimum macOS version, the Package uses the default: `.macOS(.v13)`

So by specifying `.v12`, we are not advertising support for macOS.  We are only adjusting the minimum version our Package supports.

---

#### Add an `NSLocalizedString` to the SDK
1. Add an `NSLocalizedString()` to a source file within either SDK module
2. Build the project (or package)
- [ ] OBSERVE: you see an in-line build error for the line you added, with an informative message

#### Add an `NSLocalizedString` to the Demo project
1. Add an `NSLocalizedString()` to a source file within either SDK module
2. Build the project (or package)
- [ ] OBSERVE: you DO NOT SEE an in-line build error for the line you added (the rule should only be triggered in the SDK)

cc: @AliSoftware

### DIFF
--- a/Sources/GravatarUI/Utility/SDKLocalizedString.swift
+++ b/Sources/GravatarUI/Utility/SDKLocalizedString.swift
@@ -1,15 +1,5 @@
 import Foundation
 
-/// Use this to express *intent* on your API that the string you are manipulating / returning is intended to already be localized
-/// and its value to have been provided via a call to `NSLocalizedString` or `SDKLocalizedString`.
-///
-/// Semantically speaking, a method taking or returning a `LocalizedString` is signaling that you can display said UI string
-/// to the end user, without the need to be treated as a key to be localized. The string is expected to already have been localized
-/// at that point of the code, via a call to `NSLocalizedString`, `SDKLocalizedString` or similar upstream in the code.
-///
-/// - Note: Remember though that, as a `typealias`, this won't provide any compile-time guarantee.
-typealias LocalizedString = String
-
 /// Use this function instead of `NSLocalizedString` to reference localized strings **from the library module**.
 ///
 /// You should use this `SDKLocalizedString` method in place of `NSLocalizedString` for all localized strings in the SDK.
@@ -34,6 +24,6 @@ typealias LocalizedString = String
 /// - Returns: A localized version of the string designated by `key` in the table identified by `tableName`.
 ///   If the localized string for `key` cannot be found within the table, `value` is returned.
 ///   (However, `key` is returned instead when `value` is `nil` or the empty string).
-func SDKLocalizedString(_ key: String, tableName: String? = nil, value: String? = nil, comment: String) -> LocalizedString {
+func SDKLocalizedString(_ key: String, tableName: String? = nil, value: String? = nil, comment: String) -> String {
     Bundle.module.localizedString(forKey: key, value: value, table: tableName)
 }
--- a/Sources/GravatarUI/Utility/SDKLocalizedString.swift
+++ b/Sources/GravatarUI/Utility/SDKLocalizedString.swift
@@ -1,0 +1,39 @@
+import Foundation
+
+/// Use this to express *intent* on your API that the string you are manipulating / returning is intended to already be localized
+/// and its value to have been provided via a call to `NSLocalizedString` or `SDKLocalizedString`.
+///
+/// Semantically speaking, a method taking or returning a `LocalizedString` is signaling that you can display said UI string
+/// to the end user, without the need to be treated as a key to be localized. The string is expected to already have been localized
+/// at that point of the code, via a call to `NSLocalizedString`, `SDKLocalizedString` or similar upstream in the code.
+///
+/// - Note: Remember though that, as a `typealias`, this won't provide any compile-time guarantee.
+typealias LocalizedString = String
+
+/// Use this function instead of `NSLocalizedString` to reference localized strings **from the library module**.
+///
+/// You should use this `SDKLocalizedString` method in place of `NSLocalizedString` for all localized strings in the SDK.
+/// This ensures that an app target that imports this module will perform localization lookup in the module, and not in the main app bundle,
+/// which is the default when using `NSLocalizedStrings()` without specifying `bundle = .module`.
+///
+/// - Note:
+///   Tooling: Be sure to pass this function's name as a custom routine when parsing the code to generate the main `.strings` file,
+///   using `genstrings -s SDKLocalizedString`, so that this helper method is recognized. You will also have to
+///   exclude this very file from being parsed by `genstrings`, so that it won't accidentally misinterpret that routine/function definition
+///   below as a call site and generate an error because of it.
+///
+/// - Parameters:
+///   - key: An identifying value used to reference a localized string.
+///   - tableName: The basename of the `.strings` file **in the app bundle** containing
+///     the localized values. If `tableName` is `nil`, the `Localizable` table is used.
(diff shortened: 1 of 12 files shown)


AliSoftware:
> We only need this while we are using `.strings` files that are generated by `genstrings`.

Nitpick, but just to be extra clear, while that wording might suggest that we won't need _any_ enforcement anymore once we migrate to String Catalogs, in practice we'll _still_ need some way to ensure that all calls to `String(localized::::)` in the SDK would specify `bundle: .module` _even_ once we migrate away from `genstrings`; it's just that we'll need to use a _different_ approach to enforce the adjusted requirement when it will come to that.

Sources/GravatarUI/Utility/SDKLocalizedString.swift
```diff
@@ -0,0 +1,39 @@
...
+
+/// Use this to express *intent* on your API that the string you are manipulating / returning is intended to already be localized
+/// and its value to have been provided via a call to `NSLocalizedString` or `SDKLocalizedString`.
+///
+/// Semantically speaking, a method taking or returning a `LocalizedString` is signaling that you can display said UI string
+/// to the end user, without the need to be treated as a key to be localized. The string is expected to already have been localized
+/// at that point of the code, via a call to `NSLocalizedString`, `SDKLocalizedString` or similar upstream in the code.
+///
+/// - Note: Remember though that, as a `typealias`, this won't provide any compile-time guarantee.
+typealias LocalizedString = String
```
AliSoftware: tbh, this was a nice idea I had back when I introduced those helpers in WPiOS… but in reality:
 - I'm pretty sure no WPiOS developers have used the `LocalizedString` typealias when writing feature code since I've introduced it in practice. I thought this would be useful but I'm pretty sure it's mostly ignored by devs (I'd even suspect most of them don't even know this typealias exists in the first place)
 - It is also debatable if the name of that `typealias` is as clear as what I initially though it was. My rationale was that by naming it `LocalizedString` it would make it clear in function declarations using it that the string was _already_ localized… but when you think of it as a type that has almost the same name as `NSLocalizedString`/`SDKLocalizedString`… I wonder if the name `LocalizedString` for this `typealias` will not make some devs think "this is a string that _needs_ to / is _intended_ to be localized and displayed in UI to end users, so I thus _need_ to make sure to pass that `LocalizedString` to `SDKLocalizedString` for it to be localized"… while this would be the opposite of the meaning I initially intended to convey (naming is hard!)

So in the end I wonder if it's worth keeping this `typealias` to begin with, especially if devs end up not using it after all and would instead always manipulate the `String` type. Especially given the conventions your team established as making those localized strings be defined as constants in some `enum Localized` + `static let` declarations, and thus the call sites when one needs a localized string would already look like `Localized.buttonTitle` and thus already be clear enough that this would be an already-localized string and not a key…

AliSoftware (approved):

AliSoftware:
[CI fails with](https://buildkite.com/automattic/gravatar-sdk-ios/builds/1192#0191c8a8-bf51-4042-a0f8-87477cb4e463/453-456):
```
error: “SwiftLintBuildToolPlugin” must be enabled before it can be used
```

Seems we'll need to add `-skipPackagePluginValidation` to the `xcargs` parameter when calling `run_tests`/`build_app`/`scan` from the `Fastfile` to fix this.

---

EDIT: I took the liberty to [push a commit directly to your branch and this PR to do exactly that](https://github.com/Automattic/Gravatar-SDK-iOS/pull/390/commits/50a6a886d95f1f84e799352562ffdb1ad6f13c0b)

Sources/GravatarUI/Utility/SDKLocalizedString.swift
etoledom: > Especially given the conventions your team established as making those localized strings be defined as constants in some enum Localized + static let declarations, and thus the call sites when one needs a localized string would already look like Localized.buttonTitle and thus already be clear enough that this would be an already-localized string and not a key…

I'd agree that this is enough context.

A type should help the compiler enforce a behavior, but as a type alias, it does little more than a different read, which in this case could also be a bit confusing (as @AliSoftware explained).

andrewdmontgomery: Good call

PR #400, opened 2024-09-11
Localize more strings

andrewdmontgomery:
Closes #

### Description

I went through the codebase looking for static strings that weren't localized yet.  Most of them I localized.  In `QuickEditor.swift` I just left a `TODO` item.

### Testing Steps

1. Open the Demo Project
2. Edit both the `UIKit` and `SwiftUI` schemes
3. For each scheme, go to `Run` --> `Options`, and enable `Show non-localized strings`
Screenshot 2024-09-11 at 3 30 40 PM

With this enabled, all **_non-localized_** strings will be shown in ALL CAPITAL LETTERS.

- Smoke test `AvatarPickerView`
   - Attempt to upload photo
   - OBSERVE: the menu options "Choose a Photo" or "Take a Photo" appear as expected (not CAPITALIZED)
- Smoke test `Profile View`
   - OBSERVE: the `Palette` menu options' text appears as expected (not CAPITALIZED)
- Smoke test 'Profile editor with oauth`
   - OBSERVE: The `Done` button in the navigation bar appears as expected (not CAPITALIZED) 
image

### DIFF
--- a/Sources/GravatarUI/DesignSystem/Palette.swift
+++ b/Sources/GravatarUI/DesignSystem/Palette.swift
@@ -126,7 +126,7 @@ public enum PaletteType {
 extension Palette {
     static var system: Palette {
         .init(
-            name: "System Default",
+            name: Localized.Name.system,
             foreground: .init(
                 primary: UIColor(
                     light: light.foreground.primary,
@@ -175,7 +175,7 @@ extension Palette {
 
     public static var light: Palette {
         .init(
-            name: "Light",
+            name: Localized.Name.light,
             foreground: .init(
                 primary: .black,
                 primarySlightlyDimmed: .gravatarBlack,
@@ -198,7 +198,7 @@ extension Palette {
 
     static var dark: Palette {
         .init(
-            name: "Dark",
+            name: Localized.Name.dark,
             foreground: .init(
                 primary: .white,
                 primarySlightlyDimmed: .white,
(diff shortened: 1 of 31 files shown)


Sources/GravatarUI/DesignSystem/Palette.swift
```diff
@@ -231,3 +231,25 @@ extension Palette {
...
+                "Palette.Name.light",
+                value: "Light",
+                comment: "The 'light mode' appearance of this app"
+            )
+            static let dark = SDKLocalizedString(
+                "Palette.Name.dark",
+                value: "Dark",
+                comment: "The 'dark mode' appearance of this app"
+            )
+        }
```
pinarol: The palette names here are intended as a sort of "identifier" so i suggest not localizing them. If we want to provide a localized name for each palette one day, we can have a separate "localizedName". Currently we only show these in the demo app and so it shouldn't be a problem to not localize them.

pinarol: Not for this PR but it could be better to have these names in a separate enum like the one in Orangina 🤔. 

```
    private enum ColorPaletteNames: String {
        case system = "System Default"
        case trueBlue = "True Blue"
        case darkMode = "Dark Mode"
        case lowContrast = "Low Contrast"
        case gothRave = "Goth Rave"
        case pride = "Pride"
}
```

andrewdmontgomery: Ah, I see.

Plus, the ability for developers to create a `custom` palette makes providing a localized name trickier.

Thanks, I'll revert.

And good call on the enum.

pinarol (approved):
LGTM. 👍

PR #402, opened 2024-09-12
Implement scopes

etoledom:
Closes #

### Description

This PR implements the new scopes from the backend.

### Testing Steps

- Run the SwiftUI demo app.
- Open the Profile editor with oauth page.
- Log out if needed.
- Try to log in.
  - Check that log in succeed and everything works as before.

### DIFF
--- a/Sources/Gravatar/Network/Services/ProfileService.swift
+++ b/Sources/Gravatar/Network/Services/ProfileService.swift
@@ -118,7 +118,7 @@ public struct Avatar: Decodable, Sendable {
     }
 
     public var url: String {
-        "https://gravatar.com\(imageUrl)?size=256"
+        imageUrl
     }
 }
 
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
@@ -155,6 +155,20 @@ private struct OAuthURLParams: Encodable {
     let blogID: String
     let redirectURI: String
     let userEmail: String
+    var scope1: String
+    var scope2: String
+    var scope3: String
+
+    public enum CodingKeys: String, CodingKey, CaseIterable {
+        case clientID
+        case responseType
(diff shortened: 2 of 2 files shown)


Sources/Gravatar/Network/Services/ProfileService.swift
```diff
@@ -118,7 +118,7 @@ public struct Avatar: Decodable, Sendable {
     }
 
     public var url: String {
-        "https://gravatar.com\(imageUrl)?size=256"
+        imageUrl
```
etoledom, line 121: The response of the avatar list now returns full URL.

I haven't checked yet how this affect other parts, and it might break something

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -155,13 +155,30 @@ private struct OAuthURLParams: Encodable {
...
+
+    public enum CodingKeys: String, CodingKey, CaseIterable {
+        case clientID
+        case responseType
+        case blogID
+        case redirectURI
+        case userEmail
+        case scope1 = "scope[1]"
+        case scope2 = "scope[2]"
+        case scope3 = "scope[3]"
```
pinarol, line 170: interesting, should the index start from 1 really?

Sources/Gravatar/Network/Services/ProfileService.swift
etoledom: I've checked this now, and the other model where we use `image_url`, which is the image upload response, it's also returning the updated full url, so we don't have any broken behavior.

pinarol (approved):
Looks and works good 🎉

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
etoledom: Curious indeed. I based it on the example by @diegoaer

PR #405, opened 2024-09-12
Adding secrets config for CI (and a8c devs)

### LINKED ISSUE #395: Fill out "Secrets" for the internal distribution build during CI
Our "Secrets" are empty for the internal distribution build. The OAuth related cases can't be tested this way. Let's address this and make sure those are testable by A8C folks.

etoledom:
Closes #395 

### Description

This PR adds the needed configuration for CI to use the oauth secrets.

#### Pre-requisites:
- Have `.mobile-secrets` set up, and up to date (fetch+pull latests changes).

#### On the local project:
- Remove the local copy of `Demo/Demo/Secrets.swift`.
- Run `make setup-secrets`
  - Check that the `Secrets.swift` file gets created.
- Build and run.
  - Check the oauth works.

#### On CI:
- Download the build on a device from https://github.com/Automattic/Gravatar-SDK-iOS/pull/405#issuecomment-2347351534
- Open the downloaded demo app
  - Check that oauth works.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,3 +7,6 @@ gem 'fastlane', '~> 2.222'
 gem 'fastlane-plugin-appcenter', '~> 2.1'
 gem 'fastlane-plugin-wpmreleasetoolkit', '~> 12.0'
 gem 'rubocop', '~> 1.65'
+
+plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
+eval_gemfile(plugins_path) if File.exist?(plugins_path)
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,3 @@ gem 'fastlane', '~> 2.222'
 gem 'fastlane-plugin-appcenter', '~> 2.1'
 gem 'fastlane-plugin-wpmreleasetoolkit', '~> 12.0'
 gem 'rubocop', '~> 1.65'
-
-plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
-eval_gemfile(plugins_path) if File.exist?(plugins_path)
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ bundle-install:
 fetch-code-signing: bundle-install
 	bundle exec fastlane configure_code_signing
 
-setup-secrets:
+setup-secrets: bundle-install
 	bundle exec fastlane run configure_apply
 
 swiftformat: # Automatically find and fixes lint issues
--- a/Makefile
+++ b/Makefile
@@ -50,12 +50,12 @@ build-demo-swiftui: bundle-install
 
 build-demo-for-distribution: build-demo-for-distribution-swiftui build-demo-for-distribution-uikit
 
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -72,6 +72,9 @@ bundle-install:
 fetch-code-signing: bundle-install
 	bundle exec fastlane configure_code_signing
 
+setup-secrets:
```
AliSoftware:
```suggestion
setup-secrets: bundle-install
```

Makefile
```diff
@@ -50,12 +50,12 @@ build-demo-swiftui: bundle-install
 
 build-demo-for-distribution: build-demo-for-distribution-swiftui build-demo-for-distribution-uikit
 
-build-demo-for-distribution-swiftui: fetch-code-signing check-build-number
+build-demo-for-distribution-swiftui: fetch-code-signing check-build-number setup-secrets
```
AliSoftware, line 53: Because `make build-demo-for-distribution-swiftui` will now require the _fastlane_ gem to be installed, you might want to add a call to `install_gems` in the corresponding steps of the `pipeline.yml` file before calling these make targets.

It would still work if you don't (provided that you apply [my suggestion above](https://github.com/Automattic/Gravatar-SDK-iOS/pull/405#discussion_r1758560875) of making `setup-secrets` target depend on `bundle-install`), because then `make build-demo-for-distribution-swiftui` will depend on `setup-secrets` which will depend on `bundle-install` which will install the gems like _fastlane_.

But the bonus of calling `install_gems` on CI at the start of any step that requires gems to be installed—instead of only relying on your `make` target already calling `bundle install`—is that our `install_gems` script has caching built-in. So it fetches the pre-built gems from our CI cache, making the subsequent `bundle install` calls be faster on CI.

etoledom: Thank you! I have added `install_gems` ti the `pipeline.yml` steps 👍

Gemfile
```diff
@@ -7,3 +7,6 @@ gem 'fastlane', '~> 2.222'
 gem 'fastlane-plugin-appcenter', '~> 2.1'
 gem 'fastlane-plugin-wpmreleasetoolkit', '~> 12.0'
 gem 'rubocop', '~> 1.65'
+
+plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
+eval_gemfile(plugins_path) if File.exist?(plugins_path)
```
AliSoftware: We don't use a `Pluginfile` to declare fastlane plugins separately (and instead we declare them direclty in the `Gemfile` as `gem 'fastlane-plugin-…'` like above), so those lines are not necessary.

It's actually better to remove them even if those lines are no-op (because `File.exists?(plugins_path)` would be false as the file doesn't exist), because just having an `eval` instruction in the `Gemfile` like here will make dependabot refuse to properly analyze RubyGems dependencies to find security issues in our gems or prepare PRs to fix them.

```suggestion
```

etoledom: Done 👍 🙏

AliSoftware (approved):

pinarol (approved):
Works like a charm 🎉

PR #406, opened 2024-09-13
Patch URLComponents to properly encode literal `+` characters in `URLQueryItem`s

### LINKED ISSUE #404: Email addresses with `+` cannot be used for Oauth
When attempting to login with OAuth, email addresses containing a `+` can't be used because the `+` is converted to a space.

| enter email address | OAuth prompt |
| --- | --- |
|  Image |  Image |

This appears to be caused by how `URLComponents` encodes `URLQueryItem`:
https://stackoverflow.com/a/27724627/10486646

Essentially, the `+` is a valid character in a query.  But the server will convert that character to a space.  So if the character is meant to be literally, it needs to be percent-encoded.

andrewdmontgomery:
Closes #404 

### Description

This introduces a method for using our own encoding when adding `URLQueryItem`s to a `URLComponents` object.

It defines a custom `CharacterSet` named `.restAPI` that includes all of the non-reserved characters, which will cause all other characters to be encoded.  The character set matches the definition of URL Encoding from [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt)

**Non-reserved characters** _(not encoded)_:
```
ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_.~
```

**Reserved characters** (encoded):
```
!*'();:@&=+$,/?%#[]
```

The encoding is applied to both the `name` and `value` of a `URLQueryItem`.

### Handling Encoding Ourselves

Now that we are performing our own encoding, and applying those encoded `URLQueryItem`s directly to the `.percentEncodedQueryItems` property of a `URLComponent`, we are taking responsibility for that encoding being compatible.  For example, if we attempt to assign a `URLQueryItem` with a `name` that contains characters that are not properly encoded, a runtime exception is thrown.

### Testing Steps

See #404 for details

- [ ] All unit tests should pass
- [ ] Smoke test avatars
- [ ] Try `Profile editor with oauth` in the `SwiftUI` demo app.  Enter an email address that contains a `+` character, and observe that the auto-populated email address in the OAuth web view contains that character (and not a space).

### DIFF
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
@@ -201,10 +201,7 @@ extension Encodable {
             let data = try encoder.encode(self)
             let dictionary = try JSONSerialization.jsonObject(with: data) as? [String: String]
             return dictionary?.map {
-                print("key: \($0.key) | value: \($0.value)")
-                let queryItem = URLQueryItem(name: $0.key, value: $0.value)
-                print("URLQueryItem: \(queryItem)")
-                return queryItem
+                URLQueryItem(name: $0.key, value: $0.value)
             } ?? []
         }
     }
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
@@ -72,13 +72,12 @@ public struct OAuthSession: Sendable {
     }
 
     private func oauthURL(with email: Email, secrets: Configuration.OAuthSecrets) throws -> URL {
+        let baseUrlComponents = URLComponents(string: "https://public-api.wordpress.com/oauth2/authorize")!
+
         do {
             let queryItems = try OAuthURLParams(email: email, secrets: secrets).queryItems
-            let urlComponents = URLComponents(
-                string: "https://public-api.wordpress.com/oauth2/authorize",
-                queryItems: queryItems
-            )!
-            guard let finalURL = urlComponents.url else {
+            let components = baseUrlComponents.withQueryItems(queryItems)
+            guard let finalURL = components.url else {
                 assertionFailure(
                     "Error encoding oauth secrets. Check the config in `Configuration.shared.configure(with:oauthSecrets:)` and try again"
                 )
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
@@ -157,6 +157,20 @@ private struct OAuthURLParams: Encodable {
     let blogID: String
     let redirectURI: String
     let userEmail: String
+    var scope1: String
+    var scope2: String
+    var scope3: String
+
+    public enum CodingKeys: String, CodingKey, CaseIterable {
+        case clientID
+        case responseType
+        case blogID
+        case redirectURI
+        case userEmail
+        case scope1 = "scope[1]"
+        case scope2 = "scope[2]"
+        case scope3 = "scope[3]"
+    }
 
     init(email: Email, secrets: Configuration.OAuthSecrets) {
         self.clientID = secrets.clientID
@@ -164,6 +178,9 @@ private struct OAuthURLParams: Encodable {
         self.blogID = "0"
         self.redirectURI = secrets.redirectURI
         self.userEmail = email.rawValue
+        self.scope1 = "gravatar-profile:read"
+        self.scope2 = "gravatar-profile:manage"
+        self.scope3 = "auth"
     }
 }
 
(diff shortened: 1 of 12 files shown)


pinarol (approved):
LGTM 👍

andrewdmontgomery:
After chatting with @pinarol , I refactored this to fully handle encoding, rather than just performing the single character swap.

The PR description has been updated to reflect this change.

andrewdmontgomery:
@diegoaer Does our choice of characters to encode (above) make sense from the backend?

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -199,7 +201,10 @@ extension Encodable {
             let data = try encoder.encode(self)
             let dictionary = try JSONSerialization.jsonObject(with: data) as? [String: String]
             return dictionary?.map {
-                URLQueryItem(name: $0.key, value: $0.value)
+                print("key: \($0.key) | value: \($0.value)")
+                let queryItem = URLQueryItem(name: $0.key, value: $0.value)
+                print("URLQueryItem: \(queryItem)")
+                return queryItem
```
etoledom: Not sure if we need to keep these `print`s here

etoledom (approved):
Looking good! Thank you 🙏

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
andrewdmontgomery: Oops, no.  Thanks!

andrewdmontgomery:
> Now that we are performing our own encoding, and applying those encoded URLQueryItems directly to the .percentEncodedQueryItems property of a URLComponent, we are taking responsibility for that encoding being compatible. For example, if we attempt to assign a URLQueryItem with a name that contains characters that are not properly encoded, a runtime exception is thrown.

Over the weekend, this started to make me a little nervous.  So I added a unit test that uses our new encoding function to assign `URLQueryItem`s with all kinds of non-ascii strings.  This should allow us to catch any fatalError's during unit testing.

But during my investigation, I found this note in the `Foundation` source:

```swift
    /// Returns an array of query items for this `URLComponents`, in the order in which they appear in the original query string. Any percent-encoding in a query item name or value is retained
    ///
    /// The setter combines an array containing any number of `URLQueryItem`s, each of which represents a single key-value pair, into a query string and sets the `URLComponents` query property. This property assumes the query item names and values are already correctly percent-encoded, and that the query item names do not contain the query item delimiter characters '&' and '='. Attempting to set an incorrectly percent-encoded query item or a query item name with the query item delimiter characters '&' and '=' will cause a `fatalError`.
    @available(macOS 10.13, iOS 11.0, tvOS 11.0, watchOS 4.0, *)
    public var percentEncodedQueryItems: [URLQueryItem]?
```

This confirms what I already found in testing:
> Attempting to set an incorrectly percent-encoded query item or a query item name with the query item delimiter characters '&' and '=' will cause a `fatalError`

So a name or value that is "incorrectly percent-encoded", or a name that contains "the query item delimiter characters '&' and '=' will cause a fatal error.

## RFC 3986
I also found something interesting in the `Query` section of the URI encoding RFC:
https://datatracker.ietf.org/doc/html/rfc3986/#section-3.4

First, the list of allowed characters seems to exactly match Apple's `CharacterSet.urlQueryAllowed`:
> query       = *( pchar / "/" / "?" )

Those are the allowed characters in the query section of a URI:
- `pchar`
   - Unreserved Chars
      - Uppercase, lowercase, decimal numbers
      - `Sub-Delims`: `! $ & ' ( ) * + , ; =`
      - `: @`
   - `/ ?`

All of those characters are valid in the query section.  And those seem to be exactly the characters in `CharacterSet.urlQueryAllowed`.

Second, some of the characters we are choosing to encode now (e.g. `://` and `@`) are perfectly valid, and are easier for humans if they are left un-encoded:
> However, as query components are often used to carry identifying information in the form of "key=value" pairs and one frequently used value is a reference to another URI, it is sometimes better for usability to avoid percent-encoding those characters.

So this PR is being too aggressive with it's encoding now.

So if we are going to use our own custom encoding, what we technically need is a `CharacterSet` that includes all of `CharacterSet.urlQueryAllowed`, and subtracts `+`, `&`, and `=`.

But the default behavior of `URLComponoents` already takes care of that... except for the `+` character.

This is a lot of code to cover that one character.

Which brings us back to Apple's recommendation:
```swift
components.percentEncodedQuery = components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B")
```

1. Allow `URLComponents` to handle all of the encoding
2. Update the `percentEncodedQuery` to replace _only_ the one character that we need to encode

I'm going to create a separate branch with that solution, so we can see what it looks like.

andrewdmontgomery:
Of all the characters that are **_allowed_** in a Query String, there is only one character whose meaning is defined by the implementation (us) – the `+` character's meaning is up to us.  It is valid to use it as an encoded `space`.  It is also valid to use it as a literal `+` character.

It would therefore be a bug for Apple to automatically encode the `+` character.

It is up to us to decide whether the `+` character is literally, and if so, to encode it.

Since this is the only character that we need to handle manually (based on our own needs), it is simpler and safer to update the `percentEncodedQuery` property, to encode `+` characters, than to bypass the built-in encoding and implement our own.

Issue #381, opened 2024-09-02
Quick Editor: Use PHPickerViewController


pinarol:
Let's look into using PHPickerViewController for picking the image.

PHPickerViewController:
- Designed with privacy in mind; it does not require full access to the photo library.
- The user selects only the photos or videos they want to share with the app, and the app only gets access to those specific items.

etoledom:
We only can start using `PHPickerViewController` when we implement a custom image cropper, as it doesn't offer this option.

pinarol:
Good point. 👍

pinarol:
Another possibility: Allowing to inject the whole image picker together with the cropper 🤔. So in case other repos already have PHPickerViewController + their own cropper, they can continue using it. This way at least we wouldn't be lowering their existing privacy level.

PR #407, opened 2024-09-13
Add image cropper

pinarol:
Closes part of https://github.com/Automattic/Gravatar-SDK-iOS/issues/381 . This is necessary to be able to use PHPickerViewController. 

### Description

Adding an image cropper to be used with PHPhotoPicker since it doesn't give us a cropper out of the box.

- Uses UIScrollView to zoom and traverse the image.
- Generates perfectly square images.
- Downsizes the images to the given limit (2024 x 2024 pixels for now). The max limit can be adjusted once tested with the real upload flow. 

https://github.com/user-attachments/assets/11765ec2-4796-47de-8b3c-faf3dcd17596

### Testing Steps

- UIKit Demo app > Image cropper (Only visible in DEBUG mode since the cropper is not public.)

### DIFF
--- a/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
+++ b/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
@@ -1,3 +1,4 @@
+import CoreGraphics
 import UIKit
 
 class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
@@ -6,6 +7,7 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
         static let backgroundColor = UIColor.black
         static let scrollViewHorizontalPadding: CGFloat = .DS.Padding.medium
         static let cropSize: CGFloat = 320
+        static let maxOutputImageSizeInPixels: CGFloat = 1024
     }
 
     // ScrollView for zooming and panning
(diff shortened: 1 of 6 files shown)


etoledom:
On the UI side, I like that it looks close to the system cropper 🎉 

I'd suggest two small changes, which are personal preferences (so feel free to disagree 😅 )

I think it would be nice to have the action buttons on a toolbar at the bottom, as it is in the system cropper:

CleanShot 2024-09-13 at 15 22 02@2x

And the second one: This one is a bit more tricky, since this behavior is defined in the Demo app.
We can achieve the presentation of the cropper to be at the top of the picker (again, as in the system cropper).
This I believe feels more fluid than dismissing the picker and presenting the cropper as starting a different flow.

Something like this:

CleanShot 2024-09-13 at 15 31 19

I was thinking that ideally, we could have a controller which creates one single flow composing the picking with `PHPickerViewController` and our custom cropper. In this way we can control the presentation animation internally, and this probably will also make it easier to bring it to SwiftUI as just one component which behaves as we want.

What do you think?

etoledom:
One small detail:
When I open a portrait image with the device oriented on landscape, it seems that the image is out of center:




---

This one is curious 🤔  
There seems to be a difference on the scale depending on the device orientation. The resulting images seem to be the same though, so maybe is not a problem. I'm just curious to see the difference.

Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+            comment: "Cancel the crop"
+        )
+    }
+}
+
+@MainActor
+extension UIImage {
+    // Resize the UIImage fitting within a specified maximum size.
+    func square(maxLength maxLengthInPixels: CGFloat) -> UIImage? {
+        let scale = UIScreen.main.scale
```
etoledom: `UIScreen.main` is being deprecated. 

To get the scale, this could be a good replacement option:
`UITraitCollection.current.displayScale`

Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+
+    func setImageToCrop(image: UIImage) {
+        imageView.image = image
+        scrollView.layoutIfNeeded()
+        scrollView.clipsToBounds = false
+        imageView.frame = CGRect(x: 0, y: 0, width: image.size.width, height: image.size.height)
+        scrollView.contentSize = image.size
+        let scaleHeight = scrollView.frame.size.height / scrollView.contentSize.height
+        let scaleWidth = scrollView.frame.size.width / scrollView.contentSize.width
+        let minScale = max(scaleWidth, scaleHeight) // chosing the "max" to avoid having empty black space inside the crop frame
```
etoledom, line 122: This is a very nice improvement over the system cropper 🎉

pinarol:
> When I open a portrait image with the device oriented on landscape, it seems that the image is out of center:

Should be fixed now.

Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
pinarol: Ah right. Done.

pinarol:
> And the second one: This one is a bit more tricky, since this behavior is defined in the Demo app. We can achieve the presentation of the cropper to be at the top of the picker (again, as in the system cropper).

We'll need to do that when we integrate this cropper to the avatar picker together with the PHPhotoPicker. The image cropper doesn't contain any present&dismiss calls. So it is all up to the presenting party.

pinarol:
> There seems to be a difference on the scale depending on the device orientation.

The cropping rect can be slightly smaller in landscape mode, layout constraints make it so that the cropper always fits in the view. When there's not much space left, the cropping frame will shrink a bit. Which might result in a small difference in the cropped area. The important thing is, does it actually crop the area inside the crop frame...

As for the difference between scale:

scale: 1 means we had to adjust the image. this can happen if:
 - width & height is unequal: this happens quite randomly according to what `cropping(to:...)` returns
 - cropped image's dimensions exceed max limit

scale: 3 means we didn't need to adjust it. it was already a perfect square and its dimensions are under the limit.

pinarol:
> I think it would be nice to have the action buttons on a toolbar at the bottom, as it is in the system cropper:

Yeah let's do this on a follow up PR. I created an issue for that.
https://github.com/Automattic/Gravatar-SDK-iOS/issues/408

etoledom (approved):
The cropper on its own is working prefectly now 🎉 

Let's move forward with this as a base and improve on the way 👍

PR #409, opened 2024-09-16
Properly encode literal `+` characters in `URLQueryItem`s

### LINKED ISSUE #404: Email addresses with `+` cannot be used for Oauth

andrewdmontgomery:
Closes #404 

This PR is meant to replace #406.  See that PR for a detailed discussion of why this PR exists.

### Description

This introduces a method for adding `URLQueryItem`'s to a `URLComponents` object, and encapsulates the encoding of `+` literal characters.

This is a simplified approach.  See #406 for a more comprehensive, and more fragile, solution.

### Testing Steps

See #404 for details

- [ ] All unit tests should pass
- [ ] Smoke test avatars
- [ ] Try `Profile editor with oauth` in the `SwiftUI` demo app.  Enter an email address that contains a `+` character, and observe that the auto-populated email address in the OAuth web view contains that character (and not a space).

### DIFF
--- a/Sources/Gravatar/AvatarURL.swift
+++ b/Sources/Gravatar/AvatarURL.swift
@@ -55,7 +55,7 @@ extension URL {
         guard var components = URLComponents(url: self, resolvingAgainstBaseURL: false) else {
             return nil
         }
-        components.queryItems = options.queryItems
+        components.setQueryItems(options.queryItems)
 
         if components.queryItems?.isEmpty == true {
             components.queryItems = nil
--- a/Sources/Gravatar/AvatarURL.swift
+++ b/Sources/Gravatar/AvatarURL.swift
@@ -11,12 +11,15 @@ public struct AvatarURL {
     public init?(url: URL, options: AvatarQueryOptions = AvatarQueryOptions()) {
         guard
             Self.isAvatarURL(url),
-            let components = URLComponents(url: url, resolvingAgainstBaseURL: false)?.sanitizingComponents(),
-            let sanitizedURL = components.url,
-            let url = sanitizedURL.addQueryItems(from: options)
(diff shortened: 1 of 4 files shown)


andrewdmontgomery:
If we agree that this is a safer approach, I will open a PR to adjust our fix in PocketCasts: https://github.com/Automattic/pocket-casts-ios/pull/1770

As it is now, that fix should not crash.  But that approach encodes too much, and will make it possible for a future developer to introduce a change that will result in runtime fatalErrors.

andrewdmontgomery:
https://github.com/Automattic/Gravatar-SDK-iOS/pull/406#issuecomment-2356457161

> Of all the characters that are allowed in a Query String, there is only one character whose meaning is defined by the implementation (us) – the + character's meaning is up to us. It is valid to use it as an encoded space. It is also valid to use it as a literal + character.
>
> It would therefore be a bug for Apple to automatically encode the + character.
>
> It is up to us to decide whether the + character is literally, and if so, to encode it.
>
> Since this is the only character that we need to handle manually (based on our own needs), it is simpler and safer to update the percentEncodedQuery property, to encode + characters, than to bypass the built-in encoding and implement our own.

etoledom (approved):
Looks good!

As we discussed, let's go with this simpler solution which solves the original problem already 👍

PR #410, opened 2024-09-17
Moving cropper action buttons to the bottom of the screen

### LINKED ISSUE #408: Quick Editor: Move the buttons of the cropper to the bottom area
Follow up https://github.com/Automattic/Gravatar-SDK-iOS/pull/407#issuecomment-2348976991

CleanShot 2024-09-13 at 15 22 02@2x

etoledom:
Closes #408

### Description
Moving the action buttons from the navigation bar to a toolbar at the bottom of the screen:

### Testing Steps

- Run the UIKit demo app.
- Navigate to `Image Cropper` demo.
  - **NOTE:** This won't be available on the CI builds .
- Choose and crop an image.
  - Check that the action buttons are visible and they work as expected.

### DIFF
--- a/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
+++ b/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
@@ -83,7 +83,7 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
         setToolbarItems([
             UIBarButtonItem(systemItem: .cancel, primaryAction: cancelAction),
             UIBarButtonItem(systemItem: .flexibleSpace),
-            UIBarButtonItem(systemItem: .done, primaryAction: doneAction)
+            UIBarButtonItem(systemItem: .done, primaryAction: doneAction),
         ], animated: false)
 
         navigationController?.isToolbarHidden = false
--- a/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
+++ b/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
@@ -38,8 +38,16 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
     }()
 
     let inputImage: UIImage
-    var onCompletion: ((UIImage, Bool) -> Void)?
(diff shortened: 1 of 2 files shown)


Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
```diff
@@ -63,23 +71,32 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
...
 
         view.backgroundColor = Constants.backgroundColor
         scrollView.delegate = self
 
         setImageToCrop(image: inputImage)
+        setupToolbar()
+    }
+
+    private func setupToolbar() {
+        setToolbarItems([
```
andrewdmontgomery, line 83: Just because I happen to be looking at localization stuff, this handles the right-to-left layout mirroring perfectly (as you'd hope). 👍

Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
```diff
@@ -63,23 +71,32 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
...
+            UIBarButtonItem(systemItem: .cancel, primaryAction: cancelAction),
+            UIBarButtonItem(systemItem: .flexibleSpace),
+            UIBarButtonItem(systemItem: .done, primaryAction: doneAction),
+        ], animated: false)
```
andrewdmontgomery, line 87: Just because I happen to be looking at localization stuff, this handles the right-to-left layout mirroring and localization perfectly (as you'd hope). 👍

andrewdmontgomery (approved):
👍 works as expected (and localizes nicely)

Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
etoledom: Exactly, that's was the idea of using toolbar instead of anything else. 
With the default buttons we get all system localization for free, but we are stuck with `Done` as the button title.

We can of course have a custom title, but I like the idea of reusing system labels if we can, and `Done` is clear enough I think.

PR #411, opened 2024-09-18
Using internal cropper as default

etoledom:
Part of #381

### Description

This PR uses the image cropper from https://github.com/Automattic/Gravatar-SDK-iOS/pull/407 as the default image cropper on the avatar picker.

After this, we are free to use `PHPickerViewController`.

### Testing Steps

- Run the SwiftUI demo app.
- Open the `Avatar picker view` demo screen.
- Open the avatar picker.
- Upload a new image by `Chose a photo`.
- Select an image 
  - Check that the picker is presented with title `Resize & Crop`.
  - Check that the image is cropped as expected.
- Go back to the `Avatar picker view` demo screen (closing the avatar picker).
- Switch ON `Custom image cropper`.
- Repeat the previous step to chose an image.
  - Check that the dummy image cropper is shown.


andrewdmontgomery (approved):
👍  Works as advertised

PR #412, opened 2024-09-18
Merge Localizations

andrewdmontgomery:
Closes #

### Description

- Merges localizations
- Fixes an instance where we weren't using the localized string

Note: in the Photo Picker, the `Choose a Photo` and `Take a Photo` strings are not yet localized.  Those are handled in a separate PR: #400 

#### Necessary Tests

- [ ] CI is 🟢 

#### For the Curious
If you're curious, you can check out various localizations.  However, any localization bugs will be handled in a separate PR.

1. Update the `App Language` for each scheme
image
2. Run using each scheme
3. Check out various views

Note: Most of the time, you should ignore missing localizations for any Navigation bar items.  They will only be localized if the strings are localized by the parent view, which is usaully (but not always) the Demo app.  We do not provide any localized strings in the Demo app.  There are times, though, where we we use system labels (such as `.done`).  In those cases, they will appear localized.

Note: The exception to that rule is the `Profile editor with oauth` view.  The `Done` button here is provided by the SDK, and should be localized.

### DIFF
--- a/Sources/GravatarUI/Resources/ar.lproj/Localizable.strings
+++ b/Sources/GravatarUI/Resources/ar.lproj/Localizable.strings
@@ -1,5 +1,65 @@
-/* Translation-Revision-Date: +0000 */
+/* Translation-Revision-Date: 2024-09-12 11:54:03+0000 */
 /* Plural-Forms: nplurals=6; plural=n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n%100>=3 && n%100<=10 ? 3 : n%100>=11 && n%100<=99 ? 4 : 5; */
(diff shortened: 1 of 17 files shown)


andrewdmontgomery:
@etoledom I updated the description to limit the required testing.  It just needs to pass CI.

Before we release, we'll want to QA the layouts, etc.  But this PR is just to merge in the localizations.  The QA should be handled separately.

etoledom (approved):
Looking good!

I noticed we are missing the image source text translations, but they seems to be correctly localized on trunk.

CleanShot 2024-09-19 at 21 58 02@2x

PR #415, opened 2024-09-19
Add Accept-Language Headers + Refactor

### LINKED ISSUE #399: Send Accept-Language header
We should send Accept-Language header in the REST API calls like:

> Accept-Language: en-US, fr-CA, es-ES

It seems to me like we can take the values from `NSLocale.preferredLanguages`.

andrewdmontgomery:
Closes #399 

### Description

This adds a dynamically generated `Accept-Language` header to all `URLRequest`'s.

This also does a little cleanup:
- The `Authorization` header code lived in two different places, now it lives in one.
- A useful `HeaderField` enum (for storing the header names) that was private (for `ProfileService`), is now `internal` and stores header names for more headers.
- Renamed the `.settingAuthorizationHeaderField(with:)` --> `.settingAuthorization(bearerToken:)`

Lastly, this adds some unit testing for the new header-setting functions, as well as the existing `.settingAuthorization()` functions.

### Further cleanup

Our use of headers is starting to grow.  It might be worth continuing to clean up how we handle them.

### Testing Steps

- 1 Run each Demo app
- 2 Use your favorite SSL Proxy to sniff the headers

For each function in each app:
- [ ] Each request contains a `Accept-Language` header with a value:
Example: `Accept-Language: fr-US, en-US;q=0.9`

### Reading the `Accept-Language` header values

Your device will have one or more preferred languages configured.  We assemble a value for the `Accept-Language` header from this array of languages.  The format looks like this.  Let's say your device has the following languages, in order of preference:

- French (France)
- English (UK)

`[fr-FR, en-GB]` --> `"fr-FR, en-GB;q=0.9"`

Which means:
"I prefer French (France), but will accept British English"

It is up to the service to decide how to handle this request.  From my testing, if a request contains `fr-FR, en-GB;q=0.9`, the server responds with `fr`, which is what you would expect:

- "I asked for `fr-FR`, which implies that `fr` is preferred over the rest of the options"


andrewdmontgomery:
I'm closing this PR in favor of a simpler approach: #416

PR #416, opened 2024-09-20
Add Custom Accept-Language Headers

### LINKED ISSUE #399: Send Accept-Language header

andrewdmontgomery:
Closes #399 

### Description

This adds a dynamically generated `Accept-Language` header to all `URLRequest`'s.

This also adds some unit testing for the new header-setting functions, as well as the existing `.settingAuthorization()` functions.

### About the `Accept-Language` header

By default, iOS appears to add a simple `Accept-Language` header, consisting of just the primary preferred language:

```
Accept-Language: fr-FR,fr;q=0.9
```

This PR replaces that header with a more comprehensive list, by using the top 6 preferred languages configured on the device.

That's a subtle difference.

### Further cleanup

Our use of headers is starting to grow.  It might be worth continuing to clean up how we handle them.

### Testing Steps

1. Add at least one more language to your devices list of languages (it isn't necessary to change your preferred language)
2. Run each Demo app 
3. Use your favorite SSL Proxy to sniff the headers

For each function in each app:
- [ ] Each request contains a `Accept-Language` header with a value:
Example: `Accept-Language: fr-US, en-US;q=0.9`

### Reading the `Accept-Language` header values

Your device will have one or more preferred languages configured.  We assemble a value for the `Accept-Language` header from this array of languages.  The format looks like this.  Let's say your device has the following languages, in order of preference:

- French (France)
- English (UK)

`[fr-FR, en-GB]` --> `"fr-FR, en-GB;q=0.9"`

Which means:
"I prefer French (France), but will accept British English"

It is up to the service to decide how to handle this request.  From my testing, if a request contains `fr-FR, en-GB;q=0.9`, the server responds with `fr`, which is what you would expect:

- "I asked for `fr-FR`, which implies that `fr` is preferred over the rest of the options"


andrewdmontgomery:
We are going to close this PR.

A basic `Accept-Language` header is already being added.  It does not specify a hierarchy of preferred languages.  But it accurately reflects the app's preferred language.  The lack of a list of preferred languages (if multiple are specified) is probably a security feature related to fingerprinting (see below).

If I have the system set with:
1. French
2. English

And the app is set to use the default (French), then the default header included by Apple is:
```
Accept-Language: fr-FR,fr;q=0.9
```
_"I prefer French as spoken in France, but if that isn't available, any French is acceptable"_

If I leave the system settings the same, and  I change the app to prefer English:
```
Accept-Language: en-US,en;q=0.9
```
_"I prefer English as spoken in the US, but if that isn't available, any English is acceptable."_


This is sufficient for our needs for now.

### Notes
Note that if an app prefers a non-English language that the Gravatar backend doesn't support, `en` is not explicitly specified as a fallback.  It is common for a backend to make this assumption, though.  The only way to avoid making that assumption would be to return an error.

## Future Considerations
We could choose to send a list of preferred languages.  But the work in this PR is not sufficient:

### This branch is insufficient
In our existing implementation, there are too many places where we need to add this header.  Since we currently allow third parties to use their own `HTTPClient`, we would need another abstraction to set default headers.  As this branch stands now, there are still URL requests that are missing this header.

### Don't send all preferred languages
We should be mindful of how many languages we send.  Language preferences can be an important part of fingerprinting a user without their permission.
   - https://www.rfc-editor.org/rfc/rfc9110.html#name-accept-language
   - https://www.rfc-editor.org/rfc/rfc9110.html#fingerprinting

This branch/PR currently sends up to six (6) preferred languages.  That is way too many.

### Language preferences
If we choose to send fallback languages, we will want to be explicit about preferences.  Consider the case where an app uses their System's default language, and the device is configured with the following languages:

1. French (France)
2. German (Germany)

By default, if we use the output from `Locale.preferredLanguages` to produce this string, we will end up with something like this:

```
fr-FR,de-DE;q=0.9
```
_I prefer French as spoken in France, but I will accept German as spoken in German._

Notice that this does NOT explicitly allow for `fr` or `de`.  It only allows for `fr-FR` and `de-DE`.  It is common for a backend to make the assumption that `fr-FR` really means `fr-FR,fr;q=0.9`.  But it is not guaranteed.  If we are going to provide a custom `Accept-Language`, we should be explicit (including adding `en` as the final fallback):

```
fr-FR, fr;q=0.9,de-DE;q=0.8,de;q=0.7,en;q=0.6
```

andrewdmontgomery:
From a discussion with @diegoaer:

> We only accept some locales (list below))
> We only accept exact matches, so we accept es but not es-ES

> Everything defaults to en and if we receive fr-FR (for example) we treat it as unsupported

```
	'ar'     => 'ar.gravatar.com', // Arabic (should be RTL)
	'az'     => 'az.gravatar.com', // Azerbaijani
	'bg'     => 'bg.gravatar.com', // Bulgarian
	'bn'     => 'bn.gravatar.com', // Bengali
	'br'     => 'br.gravatar.com', // Portuguese (Brazil)
	'bs'     => 'bs.gravatar.com', // Bosnian
	'ca'     => 'ca.gravatar.com', // Catalan
	'cn'     => 'cn.gravatar.com', // Chinese (China)
	'cs'     => 'cs.gravatar.com', // Czech
	'cy'     => 'cy.gravatar.com', // Welsh
	'da'     => 'da.gravatar.com', // Danish
	'de'     => 'de.gravatar.com', // German
	'el'     => 'el.gravatar.com', // Greek
	'es'     => 'es.gravatar.com', // Spanish
	'fa'     => 'fa.gravatar.com', // Persian (RTL?)
	'fi'     => 'fi.gravatar.com', // Finnish
	'fr'     => 'fr.gravatar.com', // French
	'fr-ca'  => 'fr-ca.gravatar.com', // French (Canada)
	'ga'     => 'ga.gravatar.com', // Irish
	'ge'     => 'ka.gravatar.com', // Georgian (kartuli)
	'gl'     => 'gl.gravatar.com', // Galacian
	'he'     => 'he.gravatar.com', // Hebrew (should be RTL)
	'hu'     => 'hu.gravatar.com', // Hungarian
	'id'     => 'id.gravatar.com', // Indonesian
	'it'     => 'it.gravatar.com', // Italian
	'ja'     => 'ja.gravatar.com', // Japanese
	'ka'     => 'ka.gravatar.com', // Georgian (kartuli)
	'km'     => 'km.gravatar.com', // Khmer
	'kn'     => 'kn.gravatar.com', // Kannada
	'ko'     => 'ko.gravatar.com', // Korean
	'lt'     => 'lt.gravatar.com', // Lithuanian
	'ms'     => 'ms.gravatar.com', // Malay
	'nb'     => 'nb.gravatar.com', // Norwegian (Bokmal)
	'nl'     => 'nl.gravatar.com', // Dutch
	'oc'     => 'oc.gravatar.com', // Occitan
	'pl'     => 'pl.gravatar.com', // Polish
	'po'     => 'po.gravatar.com', // Greek (Polytonic)
	'pt'     => 'pt.gravatar.com', // Portuguese (Portugal)
	'ro'     => 'ro.gravatar.com', // Romanian
	'ru'     => 'ru.gravatar.com', // Russian
	'sk'     => 'sk.gravatar.com', // Slovak
	'sl'     => 'sl.gravatar.com', // Slovenian
	'sq'     => 'sq.gravatar.com', // Albanian
	'sr'     => 'sr.gravatar.com', // Serbian
	'sv'     => 'sv.gravatar.com', // Swedish
	'skr'    => 'skr.gravatar.com', // Saraiki
	'th'     => 'th.gravatar.com', // Thai
	'tr'     => 'tr.gravatar.com', // Turkish
	'uk'     => 'uk.gravatar.com', // Ukrainian
	'ur'     => 'ur.gravatar.com', // Urdu
	'vi'     => 'vi.gravatar.com', // Vietnamese
	'zh-tw'  => 'zh-tw.gravatar.com', // Chinese (Taiwan)
```

If we decide to include our own headers, then we will need to be explicit about our fallbacks.

Issue #129, opened 2024-03-22
Improve Demo App UI


etoledom:
Many demo screens are laking some care. It would be good to make them look and feel good since they are a kind of presentation card

E.g. keyboard dismissal, saving latest used email, everywhere, scrolling content for smaller screens

PR #419, opened 2024-09-24
Image Squaring Stategy

### LINKED ISSUE #364: Quick Editor: Upload image processor
We have a "squared()" method to compensate with the Apple's default cropper's errors. I think we can include it in the upload function but in a way that is easily replaceable, like the image processor we have for the image download operation.

andrewdmontgomery:
Closes #364 

### Description

After a discussion, we have decided that this will be an internal implementation, and not configurable by users of the SDK.

This updates the `AvatarService` `upload(:::)` function to automatically square images that are nearly square.

### Testing Steps
Use the following demo image:
506 x 500 (squarable image)
506x500

550 x 500 (non-squarable image)
550x500

In the Demo app, bypass the custom image squaring that it does:
https://github.com/Automattic/Gravatar-SDK-iOS/blob/b6fd4689160737c3a137fa72b4f0d294652e51d7/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift#L185

Return the `croppedUIImage` instead of the squared version:
```diff
let croppedUIImage = UIImage(cgImage: croppedCGImage, scale: UITraitCollection.current.displayScale, orientation: .up)

guard let result = croppedUIImage.square(maxLength: Constants.maxOutputImageSizeInPixels) else { return }
- onCompletion?(result)
+ onCompletion?(croppedUIImage)
```

#### Test squarable image is squared
1. Launch the Demo app
2. Select the "Upload Image" demo
3. Enter the appropriate email/token
4. Select the squarable image (506 x 500)
5. Upload
   - [ ] OVSERVE that the upload is successful
6. Check your Gravatar profile on the web:
   - [ ] OBSERVE that the image was cropped (the 1px-wide red bars on left/right have been cropped out)

#### Test non-squarable image fails
1. Launch the Demo app
2. Select the "Upload Image" demo
3. Enter the appropriate email/token
4. Select the non-squarable image (550 x 500)
5. Upload
   - [ ] OVSERVE that the upload returns an error
6. Check your Gravatar profile on the web:
   - [ ] OBSERVE that the image was not uploaded


pinarol:
I like this idea. I think 3rd parties can really benefit from the small squaring touch we do here. Because the iOS method [cropping(to: ...)](https://developer.apple.com/documentation/coregraphics/cgimage/1454683-cropping) can return slightly uneven edges and I think it is the "go to" method when it comes to cropping. Our default strategy cuts the longer edge only if the diff is very small.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -141,10 +141,11 @@ class AvatarPickerViewModel: ObservableObject {
         }
     }
 
-    func upload(_ image: UIImage, shouldSquareImage: Bool) async {
+    func uploadImage(_ image: UIImage, imageSquaring: ImageSquaringStrategy?) async {
```
pinarol: IMO the `.default` case can be the default value for this param. It won't do anything if the image is already a square. And it will only kick in when the diff is really small.

pinarol:
Also cc: @etoledom

Sources/Gravatar/UIImage/ImageSquaring/ImageSquaringStrategy.swift
```diff
@@ -0,0 +1,17 @@
+import UIKit
+
+public enum ImageSquaringStrategy: Sendable {
+    case custom(cropper: ImageSquaring)
+    case `default`
```
pinarol: Let's add some docs to explain how the `default` strategy works.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

pinarol:
We only need the `ImageSquaringStrategy` in the `AvatarService.upload(...)` method. Not in the other UI parts. Our cropper already creates perfect squares. This is more for devs who use `AvatarService` directly. 

Because most devs use either their own cropper with [cropping(to: ...)](https://developer.apple.com/documentation/coregraphics/cgimage/1454683-cropping) or the `UIImagePickerController`'s cropper and both can generate imperfect squares(I mean just a few pixels diff between edges). This can be difficult to notice during development because it doesn't always happen, and fixing this is not much fun. So we are just compensating for that. That's why we only intended to fix the image when the diff is really small. Noticeable changes to the images are better done with a visual cropper. 

So instead of having an enum like `ImageSquaringStrategy` we can also have a parameter like `edgeDifferenceFixingThreshold: CGFloat = 0.02` in the `AvatarService.upload(...)`. And this would replace the `minorDifferenceThreshold` constant we already have. They can pass 0 if they don't want us to intervene. This is probably more suitable for the purpose.

andrewdmontgomery:
Yeah, the more I worked on this yesterday, the more weird it got.

> We only need the ImageSquaringStrategy in the AvatarService.upload(...) method. Not in the other UI parts. Our cropper already creates perfect squares. This is more for devs who use AvatarService directly.

That's clarifying.  Going into this, I thought we were trying to give devs complete control over the squaring check (like we do with the custom editor).  But the more I tried to make that work, the weirder it got.  If you are going to provide your own editor, it makes sense to make you responsible for delivering a square image from your editor.  And if your editor isn't producing square images, under what circumstances would you rather apply a second "squaring" function instead of just updating your editor?

Then I saw this comment in the `AvatarPickerView.swift`:

```swift
// If there's a custom image editor, it should take care of squaring.
```

If that's true of the custom image editor, it should also be true of ours.  So, I refactored the image editing flow to guarantee that it produces perfectly square images by:
- applying the squaring strategy immediately before the `onImageEdited()` callback
- giving developers the ability to set their own squarer
- giving developers the option of using the default squarer, but with a custom background

But again, why would developers use a custom squarer like that instead of updating their editor?

What you are proposing here is simpler.  Instead of refactoring the image editor flow to guarantee square images, we can just add an `assert(image.isSquare)` just before we call `onImageEdited()` callback, so that we give developers a warning.

andrewdmontgomery:
> So instead of having an enum like `ImageSquaringStrategy` we can also have a parameter like `edgeDifferenceFixingThreshold: CGFloat = 0.02` in the `AvatarService.upload(...)`. And this would replace the `minorDifferenceThreshold` constant we already have. They can pass 0 if they don't want us to intervene. This is probably more suitable for the purpose.

In our current implementation, that threshold defines whether the non-square image uses `aspect fit` or `aspect fill`.  All differences above that threshold will `fit` with background color added (`.black`).

I think setting that threshold to `0` (which would mean all non-square images will use `aspect fit`) would be a legitimate choice.

So I went with the argument name `cropToFitThreshold`, and I've made it optional with our default value of `0.02`:
- If set, the threshold defines which images are `fit` and `fill`
- If `nil`, no squaring is applied

pinarol (approved):
Changes to the SDK looks good 🎉 

Could you explain the Demo app changes a bit? I am a bit confused about what to expect from the new segmented control. 🤔 

Also, to be able to enter decimal numbers I think we need to make the keyboard type `.decimalPad`.

And this page overall doesn't look nice when dark mode is enabled (In Simulator "Settings > Developer > Dark Appearance"). For example the text color is not visible. It can be nice to use system defined colors so things can adjust automatically. If this looks like a big effort feel free to open a new issue so we can handle separately.

andrewdmontgomery:
> Could you explain the Demo app changes a bit? I am a bit confused about what to expect from the new segmented control.

Yeah, I really don't love this UX.  We just have no extra room.  This is the last row we can add to the setting before we start pushing things off the bottom of an iPhone SE.  So, just to make this testable, I squished and abbreviated.

I'm going to give this UX some more thought today, with a clearer head than last week.  Then, I was going to write up in the description whatever we decide on.

For now...

The segmented control decides where squaring happens:

- Does the Image Picker continue to handle squaring the image with its own logic, as it currently does?
- Or, does the Image Picker just deliver an image, and then squaring will be handled during the upload (via AvatarService)?

The problem with the existing Image Upload is that it calls `.squared()` on the selected image before returning it.  So the new squaring logic in the upload can't be used.  This setting allows us to bypass that if we want.

> Also, to be able to enter decimal numbers I think we need to make the keyboard type .decimalPad.

Oops, yes.  Thanks.

> And this page overall doesn't look nice when dark mode is enabled (In Simulator "Settings > Developer > Dark Appearance"). For example the text color is not visible. It can be nice to use system defined colors so things can adjust automatically. If this looks like a big effort feel free to open a new issue so we can handle separately.

Thanks for catching this.

andrewdmontgomery:
I pushed those changes.  Dark mode is readable now, and the decimal pad is used.  I'll update the description.

This UX needs to be revisited, though.  But #129 may be a better place to do that, perhaps with sub-issues.

etoledom:
@andrewdmontgomery - Thanks for improving the looks of the demo app!

> And this page overall doesn't look nice when dark mode is enabled

I've been having this in mind for a very long time. 

This ticket (https://github.com/Automattic/Gravatar-SDK-iOS/issues/129) is meant to handle these kind of UI issues on the demo 👍 

cc @pinarol

Sources/Gravatar/Extensions/UIImage.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    package func squared(cropToFitThreshold: CGFloat? = 0.02) -> UIImage {
+        if isSquare() {
+            return self
+        }
+
+        guard let cropToFitThreshold else {
+            return self
+        }
+
```
etoledom: If I understand correctly, calling `squared(cropToFitThreshold: nil)` will do nothing, right?

I think I'd expect it to do the squaring without any threshold instead (to square any given image) 🤔
Alternatively we could make it not nullable and force a threshold. But I prefer the previous option.

pinarol: > I think I'd expect it to do the squaring without any threshold instead (to square any given image)

Aha interesting point. This makes me think maybe we shouldn't make this value optional because `nil` is not an intuitive value in this context and it can be translated differently.

etoledom: Using `nil` sounds to me like like "use no threshold".
Not optional could be, but we will miss the option to "square anything", unless users use a very big threshold number 🤔 

Maybe it would be more clear if the default value is `nil`, so calling `image.square()` will do just that for any image ratio. And if users wants some threshold, then it is explicitly requested. But I understand that for internal use we want the threshold by default 🤔 🤔

andrewdmontgomery: Interesting.  Since the threshold value is a percentage (`0.0 ... 1.0`), all values represent reasonable outcomes:
- `0.0`: any difference at all is above the threshold, and so, the image would be `cropToFit` (with background added)
- `1.0`: any difference at all is below the threshold, and so, the image would be `cropToFill`

Since both of those are reasonable outcomes for a developer to request, then neither of those values can be used to disable cropping altogether.  So I used `nil` to fill that purpose.  That seemed better than using a negative number (which is the only other value that doesn't represent a real outcome).

But I think you are right about that being unintuitive.

This might work better as an enum with an associated value.  I'll create a branch so you can see what I'm thinking.

andrewdmontgomery: I just created a branch of this, to demonstrate an enum approach: #560

etoledom: > then neither of those values can be used to disable cropping altogether. So I used nil to fill that purpose.

This is what I don't understand. If I don't want/need the image to be cropped, I wouldn't call the `crop` function at all, right? 

The Enum option looks good to me, but I'd have the same question about the `.none` case.
What is the advantage of offering this over the user just not calling the `crop` function on the image?

pinarol: > If I don't want/need the image to be cropped, I wouldn't call the crop function at all, right?

Well, as a 3rd party dev it wouldn't be up to you because we call `squared(...)` from inside the upload(...) method. 

Maybe the check for not calling the `squared(...)` method can be inside the `upload(...)` method. But we still need to figure out how 3rd parties can skip calling `squared(...)` during the upload. 

Am I missing something?

pinarol: Oh well I think I am realizing a different issue with the formula:

Currently it is:

> abs(width - height) / min(width, height)

But, it should better be:

> abs(width - height) / *max*(width, height)

This way the result will always be smaller than 1, which suits us better imo. 

The valid range for `cropToFitThreshold` becomes: 0..1

0 -> aspectFill (cut the longer edge)
1 -> aspectFit (make the shorter edge bigger by filling in black)

But yeah we still need to provide a way to skip the cropping completely.

pinarol: I am thinking things all over again and I think our formula is not easy to understand.

I think we should just go with something like "edge ratio" instead.

Edge Ratio = a / b, where a≤b (division of the smaller edge by the larger edge of a rectangle, it is like aspectRatio but this one ensures that the value is smaller than 1)

For example when I say if `edgeRatio >= 0.98` I understand easily that the rectangle is very close to a square.

pinarol:
```
enum SquaringStrategy {
    case aspectFill
    case aspectFit
    case aspectFillAboveEdgeRatio(_ ratio: CGFloat = 0.98)
    case none
}
```

We can use a different naming than `aspectFillAboveEdgeRatio` though, feel free to suggest. 

`aspectFillOrFitBasedOnEdgeRatio` can be another option. It is quite long though... 

OR `aspectFitBelowEdgeRatio` 🤔

pinarol: Oh and _technically_, we don't need separate cases for `aspectFill` and `aspectFit`.
```
 enum SquaringStrategy {
    /// Given that Edge Ratio = a / b, where a≤b (division of the smaller edge by the larger edge of a rectangle). 
    /// If Edge Ratio of the image is >= `edgeRatioThreshold` then aspectFill will be applied, otherwise aspectFit.
    /// Pass 1 to always using `aspectFill`, and 0 for `aspectFit`.
    case aspectFillOrAspectFit(accordingToEdgeRatioThreshold: CGFloat = 0.98)
    case none
}
```

Again, still not feeling strongly about the naming. 😬

andrewdmontgomery: > I am thinking things all over again and I think our formula is not easy to understand.

Good call, I like this.

> technically, we don't need separate cases for aspectFill and aspectFit.

True.  Having them, I thought, made the choices easier to understand.  Instead of having to explain "Pass 1 to.... Pass 0 to....", we just let them choose them.  I think the choices become easier to understand, and the calling site is easier to read when developers use `0.0` or `1.0`.

andrewdmontgomery: Ok, simpler.  I tried two different naming schemes:

- `edgeRatioDeterminesFitOrFill(edgeRatio: CGFloat)`
- `squarenessDeterminesFitOrFill(squarenessThreshold: CGFloat)`

The first one is technically accurate: we are using the edge ratio.  But it obfuscates the goal, which is to make a decision based on how close to square the image is.

I'm hoping that the second one communicates the purpose.  What do you think?  Does `squareness` make sense?  I have documented the meaning on the `squareness` function.

```swift
public enum SquaringStrategy {
    case aspectFill
    case aspectFit
    case squarenessDeterminesFitOrFill(squarenessThreshold: CGFloat)
    case `default`
    case none

    package func square(_ image: UIImage) -> UIImage {
        switch self {
        case .aspectFit:
            ImageSquarer(squarenessThreshold: .aspectFitThreshold).square(image)
        case .aspectFill:
            ImageSquarer(squarenessThreshold: .aspectFillThreshold).square(image)
        case .squarenessDeterminesFitOrFill(let squarenessThreshold):
            ImageSquarer(squarenessThreshold: squarenessThreshold).square(image)
        case .default:
            ImageSquarer(squarenessThreshold: .defaultThreshold).square(image)
        case .none:
            image
        }
    }
}

extension CGFloat {
    fileprivate static let defaultThreshold: CGFloat = 0.98
    fileprivate static let aspectFitThreshold: CGFloat = 1.0
    fileprivate static let aspectFillThreshold: CGFloat = 0.0
}
```

As you've pointed out, technically we only _need_ two cases:
- `.squarenessDetereminsFitOrFill()`
- `.none`

The others are derived.  But I think having them makes this all easier to understand and reason about:

- It's easy to reverse the logic of `fit` vs `fill`.  I reversed them multiple times during this refactor.  We can document that, of course.  But when debugging at the call site, being able to see `.aspectFill` instead of `.squarenessThreshold: 0.0` is really helpful.  We could just make those CGFloat `static let` entries public and use those: `squarenessThreshold: .aspectFillThreshold`.  But doesn't read as clearly, and it's harder to discover.
- The `default` case makes the calling site easier to understand.  And enums with associated values that have default values are awkward with autocomplete.

I'm not married to making these extra cases explicit.  If we want to prune down to just the base two cases, I can be convinced.

andrewdmontgomery: I simplified and pushed to the other branch:
https://github.com/Automattic/Gravatar-SDK-iOS/pull/560#discussion_r1841021847

As I mentioned over there, as long as we like this general direction (enum instead of trying to make ` CGFloat?` work), then I can just merge that other branch into this one, and we can continue here.

pinarol: I liked your proposal. I put a small suggestion in the other PR but feel free to merge and we can continue from this one.

andrewdmontgomery:
All changes are in this branch now

Sources/Gravatar/UIImage/SquaringStrategy.swift
```diff
@@ -0,0 +1,57 @@
...
+/// The Gravatar API will return an error when an image that is not square is uploaded, even if the difference is a single row of pixels.  `SquaringStrategy` is
+/// a safeguard, making sure that an image is properly squared before uploading.
+///
+/// ## Squareness
+/// `Squareness` is similar to `Aspect Ratio`, except that all values are in the range `0...1`
+/// - A square UIImage (`100 x 100`) has a `squareness` of `1`
+/// - A UIImage with a `size` of `100 x 200` has a `squareness of `0.5`
+public enum SquaringStrategy: Sendable {
+    /// The default `SquaringStrategy`. Uses a high `aspectFillMinSquareness` value to limit which images are cropped to `aspectFill`
+    public static let `default`: SquaringStrategy = .squarenessDeterminesFitOrFill(aspectFillMinSquareness: .defaultMinSquareness)
```
pinarol: We can also explain what it means to have "Uses a high `aspectFillMinSquareness`". 

> If the image is nearly square, the aspectFill strategy will crop the longer edge. If not, aspectFit will add a black background to the shorter edge instead.

Demo/Demo/Gravatar-Demo/DemoUploadImageViewController.swift
```diff
@@ -34,6 +34,57 @@ class DemoUploadImageViewController: UIViewController {
...
+        label.font = .preferredFont(forTextStyle: .body)
+        label.textColor = .label
+        label.adjustsFontSizeToFitWidth = true
+        return label
+    }()
+    
+    let cropToFitThreshold: UITextField = {
+        let textField = UITextField()
+        textField.translatesAutoresizingMaskIntoConstraints = false
+        textField.text = "0.02"
```
pinarol: This value needs an update now right? It's not the default anymore.

andrewdmontgomery: Yes, great catch.

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -42,28 +42,69 @@ public struct AvatarService: Sendable {
...
+    /// within a square whose sides are the length of the longest side, and a background color of `UIColor.black` will be added to fill the rest of the square
+    /// - **Disable image squaring:** When the threshold is set to `nil`, no image squaring will be applied.
+    ///
+    /// If image squaring is disabled, images that are not square will raise an assertion.
+    ///
     /// - Parameters:
     ///   - image: The image to be uploaded.
     ///   - selectionBehavior: How to handle avatar selection after uploading a new avatar
     ///   - accessToken: The authentication token for the user. This is a WordPress.com OAuth2 access token.
+    ///   - cropToFitThreshold: (Optional) Controls the maximum difference between the lenght of the sides of an image, measured as a percentage, where the
```
andrewdmontgomery: This needs to be updated for the new function signature

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -42,28 +42,69 @@ public struct AvatarService: Sendable {
...
+    /// - **Disable image squaring:** When the threshold is set to `nil`, no image squaring will be applied.
+    ///
+    /// If image squaring is disabled, images that are not square will raise an assertion.
+    ///
     /// - Parameters:
     ///   - image: The image to be uploaded.
     ///   - accessToken: The authentication token for the user. This is a WordPress.com OAuth2 access token.
     ///   - avatarSelection: How to handle avatar selection after uploading a new avatar
+    ///   - cropToFitThreshold: (Optional) Controls the maximum difference between the lenght of the sides of an image, measured as a percentage, where the
+    /// image will be cropped to fit in a square.  Set to `nil` to disable image squaring.  See the `Discussion` for details.
```
andrewdmontgomery: This needs to be updated for the new function signature

pinarol (approved):
The code looks good to me but when I search for "cropToFit" I see some variable names and docs that needs updating.

Also let's hear from @etoledom as well.

andrewdmontgomery:
`CropToFit` references are removed, and I updated some of the docs.

The UI in the Demo app is definitely going to need some love:
image

That whole `Square:` line is too cramped.  And the `Min:` label barely makes senes even if you know what it is.  There just isn't any more room on an SE screen.

Sources/Gravatar/Network/Services/ImageUploadService.swift
```diff
@@ -18,6 +18,11 @@ struct ImageUploadService: ImageUploader {
         avatarSelection: AvatarSelection = .preserveSelection,
         additionalHTTPHeaders: [HTTPHeaderField]?
     ) async throws -> (data: Data, response: HTTPURLResponse) {
+        assert(
+            image.isSquare(),
+            "This image is not square, and uploading it will fail. Consider using `SquaringStrategy.default` when calling `AvatarService.upload(::::)`."
+        )
```
andrewdmontgomery: Is this a good idea?  It would help developers discover non-square images in development (DEBUG).  But will it interfere with development somehow?

Should we drop this in favor of logging something to Console?

Sources/Gravatar/Extensions/UIImage.swift
```diff
@@ -0,0 +1,32 @@
...
+    ///
+    /// `Squareness` is similar to `Aspect Ratio`, except that all values are in the range `0...1`
+    /// - A square UIImage (`100 x 100`) has a `squareness` of `1`
+    /// - A UIImage with a `size` of `100 x 200` has a `squareness of `0.5`
+    ///
+    /// - SeeAlso: ``SquaringStrategy``
+    var squareness: CGFloat {
+        if isSquare() { // This catches 0x0 images, which would cause a divide-by-zero error
+            return 1
+        }
```
etoledom: Is it fine to return 1 (perfect square) on a image without dimensions? (0x0)

Technically it's probably kinda undefined, but maybe I would have chosen zero. 
Probably not a big issue anyway. This is for internal use only anyway.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -155,7 +155,9 @@ class AvatarPickerViewModel: ObservableObject {
         // SwiftUI doesn't update the UI if the grid is empty.
         // objectWillChange forces the update.
         objectWillChange.send()
-        let squareImage = shouldSquareImage ? image.squared() : image
+        let squareImage = shouldSquareImage ? SquaringStrategy.default.square(image) : image
```
etoledom: I still don't understand when `SquaringStrategy.none.square(image)` would be useful.
Here we choose the untouched image directly as I'd expect.

But if both of you think is good to have it, I'll go with it ¯\\\_(ツ)_/¯

etoledom (approved):
Looks good to me also.

I keep my question from before, but since you both seem to agree that it's a good thing, let's go with it. (https://github.com/Automattic/Gravatar-SDK-iOS/pull/419/files#r1843637494)

Sources/Gravatar/Network/Services/ImageUploadService.swift
etoledom: Maybe is better to return the error from the server in this case. 
We don't know if/when the backend will start allowing non-squared images, so it's probably better to let the image reach the backend and it can decide? 🤔

Sources/Gravatar/UIImage/SquaringStrategy.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Applies the `SquaringStrategy` to a `UIImage`
+    /// - Parameter image: a `UIImage` that should be squared
+    /// - Returns: a `UIImage` with the `SquaringStrategy` applied.
+    package func square(_ image: UIImage) -> UIImage {
+        switch self {
+        case .squarenessDeterminesFitOrFill(let aspectFillMinSquareness):
+            ImageSquarer(aspectFillMinSquareness: aspectFillMinSquareness).square(image)
+        case .none:
+            // TODO: Check for squareness and log non-square images
```
pinarol: Is the TODO a leftover? No need to do anything here.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
pinarol: I am a bit confused since the code says `SquaringStrategy.default.square(image)` and you say `SquaringStrategy.none.square(image)`. Can you clarify which one did you intend to quote?

Sources/Gravatar/Network/Services/ImageUploadService.swift
pinarol: Now that we have a clear error message coming from the backend, i think it is fine to defer this to backend and remove this assert.

Sources/Gravatar/Extensions/UIImage.swift
pinarol: I think it is up to us to treat 0x0 as a square in this context. It should be fine.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
andrewdmontgomery: Ah, I think I see the question.

```suggestion
        let squareImage = shouldSquareImage
            ? SquaringStrategy.default.square(image)
            : SquaringStrategy.none.square(image)
```

That's also valid.  I wasn't sure if it made things more or less clear.  But since this is an internal implementation detail, I went with the more concise change.

But I think your question really points out that `.none` isn't necessary in this case.  In fact, none of this is necessary at this point in the call stack.  The purpose of this PR is to:
- Move the "squaring" function into the `AvatarService.upload()` function (which this function will eventually call in `doUpload()`).
- Allow third-party devs to disable this squaring

But now that I'm looking at this, that approach is problematic.  The `doUpload(squareImage: ...)` function assumes it receives a square image:
- It calls `AvatarService.upload()` with the image (which will now square the image)
- But it also caches the `squareImage`

Hmm... I will think about this.

I think the question remains though: under what circumstances would a developer want to disable squaring?  First, they would have to supply their own editor, since ours doesn't produce square images.

And if they _do_ supply a square image, then ALL strategies will simply return the original image – untouched.  They literally return the same object.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -156,6 +156,8 @@ class AvatarPickerViewModel: ObservableObject {
         // objectWillChange forces the update.
         objectWillChange.send()
         let squareImage = shouldSquareImage ? image.squared() : image
```
pinarol: I think I added this line at the time but we can remove it altogether imo, the small squaring fix will be applied in the upload function anyway.

Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
```diff
@@ -236,12 +236,12 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
 @MainActor
 extension UIImage {
     // Resize the UIImage fitting within a specified maximum size.
-    func square(maxLength maxLengthInPixels: CGFloat) -> UIImage? {
```
pinarol: Can we not touch `ImageCropperViewController` at all for this change? These are developed solely for the cropper and there are quite some logic here I added to handle some edge cases.

Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
```diff
@@ -236,12 +236,12 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
...
-    }
+//    func square(maxLength maxLengthInPixels: CGFloat) -> UIImage? {
+//        let scale = UITraitCollection.current.displayScale
+//        let smallerEgde = min(size.width * scale, size.height * scale)
+//        let squareEdge = floor(min(maxLengthInPixels, smallerEgde))
+//        return downsize(to: squareEdge)
+//    }
 
     // Downsize to targetSquareEdgeInPixels
     private func downsize(to targetSquareEdgeInPixels: CGFloat) -> UIImage? {
```
pinarol: downsize(to... has lots of logic necessary within this class' context. Let's keep things as is here.

Sources/Gravatar/Extensions/UIImage.swift
```diff
@@ -0,0 +1,103 @@
...
+        max(self.size.width, self.size.height)
+    }
+
+    package func squared(
+        withinTolerance squarenessTolerance: CGFloat = .defaultSquarenessTolerance,
+        maxSize: ImageSize? = nil
+    ) -> UIImage {
+        self
+            .squared(withinTolerance: squarenessTolerance)
+            .resized(toMaxSize: maxSize)
```
pinarol: No resize needed. Let's keep it simple.

Demo/Demo/Gravatar-Demo/DemoUploadImageViewController.swift
```diff
@@ -34,6 +38,57 @@ class DemoUploadImageViewController: UIViewController {
         return textField
     }()
     
+    let squaringLabel: UILabel = {
```
pinarol: None of the demo app changes needed anymore?

Sources/Gravatar/Extensions/UIImage.swift
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    package func squared(withinTolerance squarenessTolerance: CGFloat) -> UIImage {
+        guard !self.isSquare, self.deviationFromSquare <= squarenessTolerance.clamped(to: 0 ... 1.0) else { return self }
```
pinarol: `self.deviationFromSquare <= squarenessTolerance.clamped(to: 0 ... 1.0)` is way too complex for what we are doing here.

why not just check like below?

> squareness >= 0.98

Sources/Gravatar/Utility/Clampable.swift
```diff
@@ -0,0 +1,14 @@
...
+    func clamped(to range: ClosedRange<Bound>) -> Bound
+}
+
+extension Clampable where Self: FloatingPoint {
+    func clamped(to range: ClosedRange<Self>) -> Self {
+        Self.minimum(Self.maximum(self, range.lowerBound), range.upperBound)
+    }
+}
+
+extension CGFloat: Clampable {}
```
pinarol: Do we really need this?

pinarol (changes requested):
Let's keep changes minimal. I think we don't need 90% of changes in this PR anymore. It can be easier to start from `trunk` at this point.

Demo/Demo/Gravatar-Demo/DemoUploadImageViewController.swift
andrewdmontgomery: Yep.  I just removed it.

Sources/Gravatar/Utility/Clampable.swift
andrewdmontgomery: I'm constantly having to carefully reason through the `min(max())` stuff.  So I found it helpful to encapsulate it in a function and write some tests for it.

But it could just be a function in an extension of `CGFloat` (or more generically, the `FloatingPoint` protocol).  I'll simplify.

👍

Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
andrewdmontgomery: This will mean maintaining two separate squaring functions:
- One to fix our own cropper
- One to fix uploads

But you make a good point.  If there are edge cases we aren't (or can't easily) test, then this isn't safe, and we should keep using this custom squarer.

I'll revert.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
andrewdmontgomery: Thanks for calling this out.  Yeah, agreed.

andrewdmontgomery: Ah, wait.  I remember the problem.

This image gets passed to `doUpload(squareImage:)`.  The name of that parameter (`squareImage`) communicates a dependency: that function caches whatever image we pass to it.   The squaring function added in this PR happens in the `AvatarService.upload()`.  That makes the squaring a side-effect of uploading, which isn't a great pattern.

We could introduce a `package func prepareImageForUpload()` function in `AvatarService`, and then call that:
1. manually, in the `doUpload()` function
2. Also call it from the `AvatarService.upload()` function

It would get called twice from the AvatarPicker.  But the second call would be a no-op, since it will already be squared.

That won't help users of the public `upload()` function, though.  The only way the will be able to cache the cropped image is to re-fetch after uploading.

I'm not sure `AvatarPickerViewModel` (`doUpload()`) should be responsible for setting the `ImageCache.shared`.  The `ImageDownloadService` uses the `ImageCache`.  The `ImageUploadService` could, too.

That would address this problem.

If that sounds like a good approach, we could remove this squaring from here.  That will leave us with a small bug (cached images may not be cropped).  Then we could address that bug by moving where we handle caching.

Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
andrewdmontgomery: And if we aren't going to reuse this for our own image picker, this `squared(aboveThreshold: CGFloat, maxSize: ImageSize)` no longer needs to handle `maxSize`.

Do we want to square and reduce?  That's beyond the scope of what we wanted.  But I have the logic (and the unit tests).

andrewdmontgomery: Note: as written, `maxSize` is `nil`, so it isn't reducing.  If we wanted to reduce uploads automatically, we'd set it to `.pixel(2048)`.

And if the API someday provided a `config` route, we could enforce whatever maximum it specifies.

andrewdmontgomery: For now I have removed the `maxSize` logic.  See [commit](https://github.com/Automattic/Gravatar-SDK-iOS/commit/b82e64b2a69a02e4b23d5545c43416d2026761b5) for the removal.

Sources/Gravatar/Extensions/UIImage.swift
andrewdmontgomery: Yeah, last night I thought using a "tolerance" value (instead of a "threshold") might be more intuitive.  So I tried it (which implied calculating the deviation).

But it was much less intuitive.  So I reverted this morning.

andrewdmontgomery:
> Let's keep changes minimal. I think we don't need 90% of changes in this PR anymore. It can be easier to start from `trunk` at this point.

Agreed.

Take one more look. I've pulled a ton out:

- The new squaring function is only called during `AvatarService.upload()`, and is scoped `internal` to `Gravatar`
- No public changes to the API
- All of the new `maxSize` logic has been pulled (See [comment](#discussion_r1850828864) above)
- Clamping of `CGFloat` is just a function extended from `CGFloat`

Other change
- `AvatarPickerViewModel` no longer squares images as it used to.   This will cause a caching bug as described above.  (See [comment](#discussion_r1850785183) above)

Sources/Gravatar/Utility/Clampable.swift
andrewdmontgomery: it's just a function on `CGFloat` now

andrewdmontgomery:
The PR description has been updated for the simplified scope, including testing steps.

Sources/Gravatar/Extensions/UIImage.swift
```diff
@@ -0,0 +1,66 @@
...
+        let squareSideLength = floor(self.shortEdge)
+
+        let squareSize = CGSize(width: squareSideLength, height: squareSideLength)
+        let imageOrigin = CGPoint(
+            x: (squareSize.width - width) / 2,
+            y: (squareSize.height - height) / 2
+        )
+
+        let format = UIGraphicsImageRendererFormat()
+        format.scale = self.scale // Respect original image scale
```
andrewdmontgomery: In the original version of this function (moved from `AvatarPickerViewModel`), we set the `scale` to `1`.

This causes a problem for images with a scale greater than `1`.  The `height` and `width` of the image are in points, not pixels.  The same is true when we create the `CGSize` for the new square.

So when the original scale is `2`, for example, then the **_pixel_** dimensions for the new square image will be halved, since `pixels = points * scale`:

- Original width/height: 99 x 100 points (@ 2x) --> pixels: 198 x 200
- Square width/height: 99 x99 points (@1x) --> pixels: 99 x 99

If we halve the scale, we halve the pixels.

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -63,7 +63,12 @@ public struct AvatarService: Sendable {
     @discardableResult
     package func upload(_ image: UIImage, accessToken: String, selectionBehavior: AvatarSelection) async throws -> Avatar {
         do {
-            let (data, _) = try await imageUploader.uploadImage(image, accessToken: accessToken, avatarSelection: selectionBehavior, additionalHTTPHeaders: nil)
+            let (data, _) = try await imageUploader.uploadImage(
+                image.squared(),
```
andrewdmontgomery, line 67: Here's where we call the new `squared()` method.

Sources/Gravatar/Extensions/UIImage.swift
```diff
@@ -0,0 +1,66 @@
...
+// MARK: - Internal
+
+extension UIImage {
+    /// Crops a `UIImage` to be square if it's `squareness` is above a given `squarenessThreshold`. Images with a `squareness` below the threshold will not be
+    /// squared.
+    /// - Parameters:
+    ///   - squarenessThreshold: The threshold over which images should be squared.
+    /// - Returns: A `UIImage` that has been squared according to the threshold
+    func squared(aboveThreshold squarenessThreshold: CGFloat = .defaultSquarenessThreshold) -> UIImage {
+        guard !self.isSquare, self.squareness > squarenessThreshold.clamped(to: 0 ... 1.0) else { return self }
```
pinarol: Sorry I wasn't clear, when I asked if this was necessary I meant the whole clamping operation, whether it's a function or a protocol...

I don't think that's how we should handle a value that's out of the range. This looks like we accept all possible CGFloat values because we internally fix the value. Instead, we could validate the value and throw and error, or, we could use an assertion error in debug config and return self without doing anything on the image. That way we teach the caller party that we expect this value to be between 0 and 1.

Also, this code is internal now, so we don't need to worry about out of range values. We can just document the expected value and move forward without any kind of clamping.

So what I suggest is basically getting rid of the whole clamping thing.

pinarol: I suggest we use `>=`instead of `>`. I feel like we would want to fix 100x98.

andrewdmontgomery: The challenge is with the wording of the parameter: `aboveThreshold`:
- `>`: above threshold
- `>=`: at or above threshold

Maybe I'm overthinking it, though?  Would you be surprised to discover that `aboveThreshold` applies to values _at the threshold_?

If you're ok with it, I'm ok with it.

The other options would be:
- Change the wording: `atOrAboveThreshold` (that feels unnecessarily complicated)
- Change the threshold: `0.97` (so values `> 0.97` are squared)

andrewdmontgomery: Gotcha.  Fair points.  I'll remove it.

andrewdmontgomery: Pushed the change (and updated testing to cover this case).

I think this is fine.  Even if the english isn't precise, the meaning makes more intuitive sense.

andrewdmontgomery: Removed

Sources/Gravatar/Extensions/UIImage.swift
andrewdmontgomery: Ah, but you [took care of this](https://github.com/Automattic/Gravatar-SDK-iOS/blob/2e7a9f570e159f672462881773d46a228b0b9144/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift#L256-L258):

```swift
let newSize = CGSize(width: targetSquareEdgeInPixels, height: targetSquareEdgeInPixels)
let format = UIGraphicsImageRendererFormat()
format.scale = 1
```

You create that `newSize` with `pixel` values, implying `points == pixels`.  So the `scale` needs to be `1`.

Is this is to avoid fractional pixels, which could be rounded up (and above the `targetSquareEdgeInPixels`)?

Nice.

Do we want the same strategy here?  This `squared()` function doesn't need to `downsize` to a specific size.  As long as we don't return an image that's _bigger_ (even by a pixel) than what we received, I think we're clear.

pinarol: I think as soon as we are consistent, it should be fine. If we do everything in points then we should set format.scale to `self.scale`, if in pixels then `1`.

Sources/Gravatar/Extensions/UIImage.swift
```diff
@@ -0,0 +1,67 @@
...
+            x: (squareSize.width - width) / 2,
+            y: (squareSize.height - height) / 2
+        )
+
+        let format = UIGraphicsImageRendererFormat()
+        format.scale = self.scale // Respect original image scale
+
+        return UIGraphicsImageRenderer(size: squareSize, format: format).image { _ in
+            // Draw the image in the center of the new square context
+            self.draw(in: CGRect(origin: imageOrigin, size: self.size))
```
pinarol: Oh I just noticed. I think we have to use `squareSize` instead of `self.size` here.

Sources/Gravatar/Extensions/UIImage.swift
andrewdmontgomery: ~I see what's happening.  I'm going to use your method above (scale = 1).~

I'm adding tests for this.

andrewdmontgomery: > I think as soon as we are consistent, it should be fine. If we do everything in points then we should set format.scale to `self.scale`, if in pixels then `1`.

Yep, I added some tests and either works.  I'll leave it as points, and keep the tests.

Sources/Gravatar/Extensions/UIImage.swift
andrewdmontgomery: Swift Testing tests are so flexible. 🎉   I updated these tests to test at all scales.  I found two small issues.

First, I was using `floor(image.shortEdge)`, mostly because we use `floor()` in the existing `squared()` function.  But at larger scales, this has the effect of reducing the final square image by one pixel in both directions.

The other issue is just an issue of floating point precision for `UIImage`'s at different scales:
- ✅ `@1x`: Calling `squared()` on a `98 x 100` test image works great.  The image is square to `98 x 98`. 
- ✅ `@2x`: Calling it on the same image (`49 x 50` points) works fine, too.
- ❌ `@3x`: Calling it on the same image (`32.667 x 33.333` points) produces a `squareness` of `0.9799999999999999`

We can go three ways with this:
1. We can leave it and live with it
2. We can shift our threshold down from `0.98` to `0.97`
3. We can introduce a comparison for CGFloats that supports a "tolerance" (e.g. Apple Numerics, roll our own, etc.)

I'm leaning toward leaving it.

We could experiment with lowering the threshold.  But the issue only really impacts small images, since a single extra row of pixels results in a large change in `squareness`.

I don't think we need or want the added complexity of handling CGFloat comparisons more accurately.  This threshold is arbitrary.

pinarol: > ❌ @3x: Calling it on the same image (32.667 x 33.333 points) produces a squareness of 0.9799999999999999

Ohh I see. Since 1 is not "exactly" divisible by 3... That's why I like to use pixels in these kinds of things.

Try calculating `squareness` as `(shortEdge * scale) / (longEdge * scale)` and it should be fixed.

pinarol: To elaborate, I think it is because representing pixels in points is tricky. When the scale is 3, 1pt has 3pixels. But what if the number of pixels aren't dividable by 3? In case of 98pixels, we have only 2pixels left for the last point then it's represented by 2/3 points. Same for the 100px edge (100-99 = 1) and 1px == 1/3 points. So just because we represent things in points we end up with some infinite fractional numbers and they need to be rounded at some point. When we calculate `squareness` in pixels it'll be more close to reality. That's my understanding.

andrewdmontgomery: Floating point precision is fun.

Yep, you're right.  Multiplying by the scale works.  It makes sense, now.  By multiplying by the scale first, we are "reverting" the division (of our pixel value) that happens for each scale before doing more division.

image

andrewdmontgomery: Fixed.  I added a test for this case, too.

andrewdmontgomery:
Since the Demo app no longer needs changes for this PR, this PR no longer fixes the Dark Mode issues with the Upload demo.  But that can be handled in the UX improvements issue (#129)

pinarol (approved):
Looks good to me 🎉 And thanks for your patience 🙏 This was such a roller coaster.

andrewdmontgomery:
> Looks good to me 🎉 And thanks for your patience 🙏 This was such a roller coaster.

Thanks for your diligence and guidance. 🙏

PR #420, opened 2024-09-24
Show upload fail error dialog

### LINKED ISSUE #373: Quick Editor: Detailed upload errors
Once this issue is resolved https://github.a8c.com/Automattic/gravatar/issues/108692 we need adapt our side. We should show more descriptive error messages. And we should distinguish retry-able errors from non-retry-able ones.

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/373

### Description

Adds upload fail dialog. 

Some errors are not suitable for retrying, so we don't show the retry option for them. These are the HTTP 400 errors for now. We know from REST API specs that these are validation errors about the image. Maybe we should make this assumption for all 4XX errors. I feel like it is better to show the Retry option in a false positive situation than not show it when it is actually needed.

Others look like this:

### Testing Steps

SwiftUI demo app > Avatar Picker View
- Try for both horizontal and vertical layouts
- Try uploading an image when you are offline.
- Observe: The error dialog appears with "Remove" and "Retry" options.

(It is a bit difficult to generate HTTP 400 errors, i had to comment out some image cropping code to get that not-square error ☝️. )

### DIFF
--- a/Sources/Gravatar/Network/APIErrorPayload.swift
+++ b/Sources/Gravatar/Network/APIErrorPayload.swift
@@ -1,0 +1,15 @@
+import Foundation
+
+/// Error payload for the REST API calls.
+public protocol APIErrorPayload: Sendable {
+    /// A business error code that identifies this error. (This is not the HTTP status code.)
+    var code: String { get }
+    /// Error message that comes from the REST API.
+    var message: String? { get }
+}
+
+extension ModelError: APIErrorPayload {
+    public var message: String? {
+        error
+    }
+}
--- a/Sources/Gravatar/Network/Services/URLSessionHTTPClient.swift
+++ b/Sources/Gravatar/Network/Services/URLSessionHTTPClient.swift
@@ -65,13 +65,8 @@ extension HTTPClientError {
             return .URLSessionError(error: error)
         case .invalidHTTPStatusCodeError(let response, let data):
             if response.statusCode == 400 {
-                do {
-                    let error: ModelError = try data.decode()
-                    return .invalidHTTPStatusCode(response: response, errorPayload: error)
-                } catch {
-                    // if parsing the error fails then return invalidHTTPStatusCode without the payload.
-                    return .invalidHTTPStatusCode(response: response)
-                }
+                let error: ModelError? = try? data.decode()
+                return .invalidHTTPStatusCode(response: response, errorPayload: error)
             } else {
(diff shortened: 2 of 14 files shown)


Sources/Gravatar/Network/APIErrorPayload.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// Error payload for the REST API calls.
+public protocol APIErrorPayload: Sendable {
+    /// A business error code that identifies this error. (This is not the HTTP status code.)
+    var code: String { get }
+    /// Error message that comes from the REST API.
+    var message: String? { get }
+}
```
pinarol, line 9: I checked the Gravatar backend, the errors have a standard format, a code and a message. The message is optional. So i think we can represent all the business errors with this protocol.

Sources/Gravatar/Network/Services/URLSessionHTTPClient.swift
```diff
@@ -62,11 +62,21 @@ extension HTTPClientError {
...
         case .invalidHTTPStatusCodeError(let response, let data):
-            .invalidHTTPStatusCode(response: response, data: data)
+            if response.statusCode == 400 {
+                do {
+                    let error: ModelError = try data.decode()
+                    return .invalidHTTPStatusCode(response: response, errorPayload: error)
+                } catch {
+                    // if parsing the error fails then return invalidHTTPStatusCode without the payload.
+                    return .invalidHTTPStatusCode(response: response)
+                }
```
etoledom: Since we are skipping handling the parsing error, maybe we can avoid using try-catch?

```suggestion
                let error: ModelError? = try? data.decode()
                return .invalidHTTPStatusCode(response: response, errorPayload: error)
```

etoledom (approved):
Looking good!

Added a small suggestion on code (up to you if it looks good).

And one other small detail to check before merging:

CleanShot 2024-09-25 at 09 18 25@2x

I'm not sure if the title should have a full stop here.

Sources/Gravatar/Network/Services/URLSessionHTTPClient.swift
pinarol: Nice. 👍

pinarol:
> I'm not sure if the title should have a full stop here.

Hmm now that you pointed out I feel like it shouldn't. I'll remove the `.` 🤔

PR #423, opened 2024-09-24
Use Implicit OAuth (token response)

### LINKED ISSUE #421: Start using the “token” response type in OAuth flow
we are still using the “code” response type for oauth flow, we need to use "token" and make any adjustments to the code where necessary.
To be able to test this we need to use the `wpcom-local-dev` custom schema in the redirectURI defined in the Secrets.swift. Something like:

    static let redirectURI: String = "wpcom-local-dev://authorization-callback"

Normally, the BE doesn't allow us to use token endpoint with a custom schema but wpcom-local-dev is a special schema that's white listed at the backend for testing purposes.

If I recall right, we might need a code change for parsing the access token from the redirect url after switching to "token".

andrewdmontgomery:
Closes #421 

### Description

This swaps the OAuth implementation from "Server Code  (requesting `code` response type) to Implicit OAuth (requesting `token` response type).

The implementation of Implicit OAuth is simpler, as the token is supplied in the URL fragment of the callback from the `authorization` endpoint.

This introduces an internal type called `AccessToken`, mostly for unit testing.  Eventually, we could update the `AccessToken` to include the `Email`, and then encapsulate the keychain operations.

### Testing Steps

- [ ] Test OAuth in the Demo app
   - Run the SwiftUI app
   - Log out if needed
   - Try to log in
   - Check that login succeeds and everything works as before

### DIFF
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
@@ -45,29 +45,11 @@ public struct OAuthSession: Sendable {
     }
 
     private func getToken(from callbackURL: URL, secrets: Configuration.OAuthSecrets) async throws -> String {
-        let queryItems = URLComponents(string: callbackURL.absoluteString)?.queryItems
-        guard let code = queryItems?.filter({ $0.name == "code" }).first?.value else {
+        guard let accessToken = AccessToken(from: callbackURL) else {
             throw OAuthError.couldNotParseAccessCode(callbackURL.absoluteString)
         }
 
-        return try await requestAccessToken(code: code, secrets: secrets)
-    }
-
-    private func requestAccessToken(code: String, secrets: Configuration.OAuthSecrets) async throws -> String {
-        do {
-            let request = try accessTokenRequest(with: code, secrets: secrets)
-            let (data, response) = try await URLSession.shared.data(for: request)
-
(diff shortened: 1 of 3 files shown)


andrewdmontgomery:
I was going to parse the `expires_in` field in the response, also, and include it in `AccessToken`.  But for now, the only thing we could do with it would be to return a message "Session Expired".  And we already do that.

pinarol (approved):
Looks good 🎉

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -36,38 +36,20 @@ public struct OAuthSession: Sendable {
...
             return token
         } catch {
             throw OAuthError.from(error: error)
         }
     }
 
-    private func getToken(from callbackURL: URL, secrets: Configuration.OAuthSecrets) async throws -> String {
-        let queryItems = URLComponents(string: callbackURL.absoluteString)?.queryItems
-        guard let code = queryItems?.filter({ $0.name == "code" }).first?.value else {
+    private func tokenResponse(from callbackURL: URL) async throws -> AccessToken {
```
pinarol: this function is not async anymore so we should remove `async`

andrewdmontgomery: nice catch

PR #425, opened 2024-09-25
Using PHPicker

### LINKED ISSUE #381: Quick Editor: Use PHPickerViewController

etoledom:
Closes #381 

### Description

Using `PHPickerViewController` instead of `UIImagePickerController` for picking on-device photos.
`UIImagePickerController` is still used for the device camera, and both (camera and device photos) use the custom Image Cropper as photo editor.

### Testing Steps

- Run the SwiftUI demo app.
- Open the Avatar Picker demo screen .
- Present the Avatar Picker.
- Add a new image.
  - Check that both options (`Take photo` and `Choose image`) work as expected.


pinarol (approved):
Looks good 👍

PR #426, opened 2024-09-25
Small update to v3 endpoint

etoledom:
Closes #

### Description

`selected_email` query param became `selected_email_hash`

Plus updating spec to latest.

### Testing Steps

- Run SwiftUI demo app.
- Open the Avatar Picker.
  - Check that there is an avatar pre-selected, and the profile card at the top shows the selected avatar.


pinarol (approved):
LGTM 👍

PR #427, opened 2024-09-25
Check if the correct user is authorized during OAuth

### LINKED ISSUE #370: Quick Editor: Check if correct user is authenticated
Backend task: https://github.a8c.com/Automattic/gravatar/issues/108703

We need to check if the correct user is authenticated after the OAuth flow. So after the OAuth flow we need to call this new endpoint and compare the emails. If not, we should show a proper error.

We only need this after the OAuth flow since the OAuth flow allows to login with a different email than intended.

In case the 3rd party app already manages their own auth token (like the Jetpack app) we can assume they pass the correct email and the token. If they fail to do that then it is a bug on that end and needs fixing. No need to call the `/me` endpoint on our side for this every time we show the picker.

pinarol:
Closes #370 

### Description

Calls `/v3/me/associated-email` to check if the correct user is authenticated right after oauth. Displays an error message otherwise:

### Testing Steps

SwiftUI Demo app > Profile editor with oauth
Login with email A first
Then enter a different email in the SwiftUI app and enter the OAuth flow
Tap "Approve" (since the browser remembers your previous login)


etoledom (approved):
Looks and works great 🎉

PR #430, opened 2024-09-26
Use a thread safe cache in TestImageCache to fix a crash

pinarol:
Closes #

### Description

Fixes a crash that happens in CI from time to time:

Screenshot 2024-09-26 at 12 08 55

I think the crash happened because the mock cache in the TestImageCache was not thread safe so i replace it with `ImageCache` which is thread safe.

### Testing Steps

Try running the failing test case many times `testNotCurrentSourceTaskResult`.


etoledom (approved):
Thank you 🙏 
I didn't see the crash any more after running the test many times

PR #432, opened 2024-09-26
QE OAuth: Ignore the cancelled error so it doesn't override our current error

### LINKED ISSUE #431: QE: Make OAuth flow ignore the cancelled state
When [this error](https://github.com/Automattic/Gravatar-SDK-iOS/pull/427) happens and the user starts the OAuth flow, if the user cancels the flow and returns back we reset the error. In this case we want to keep the error.

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/431

### Description

We start to detect if the user has cancelled the oauth flow and treat it just like a dismissal of the OAuth flow so we ignore the error and the previous state remains same. 

### Testing Steps

SwiftUI Demo app > Profile editor with OAuth
Follow the steps here https://github.com/Automattic/Gravatar-SDK-iOS/pull/427 to see the wrong email error
Restart the oauth flow but this time tap cancel
Observe: The same wrong email error remains

cc: @hamorillo

### DIFF
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
@@ -80,7 +80,7 @@ enum OAuthError: Error {
     case notConfigured
     case couldNotCreateOAuthURLWithGivenSecrets
     case couldNotParseAccessCode(String)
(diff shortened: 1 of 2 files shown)


etoledom (approved):
Good catch! 🎉

PR #441, opened 2024-09-30
Image Download: Add cache buster if forceRefresh == true

### LINKED ISSUE #366: "Cache buster" parameter for downloading the most recent version of the avatar
Backend does not immediately give us the recently uploaded avatar. We need to add a cache buster query param like:

```
"&_=\(Int(NSDate().timeIntervalSince1970))"
```
This was the solution we used in the Pocket Casts too.

We can probably do this in the image downloader if `forceRefresh == true` and the domain is "gravatar.com". 

We should still use the original url for the cache key.

Another option is to implement a more sophisticated cache key mechanism to allow cache key to be different from the absolute URL 🤔 .

pinarol:
Closes #366 

### Description

Unfortunately Gravatar doesn't support cache-control headers. So i am adding a cache buster query parameter if `forceRefresh == true`. I also updated the cache policy to `.reloadIgnoringLocalAndRemoteCacheData` if `forceRefresh == true`, it doesn't work yet and ignored by the backend. But there's a systems request about it and if the server side starts respecting it, all we need to do will be to remove the cache buster param.

### Testing Steps

UIKit demo app > UIImageViewExtension
Switch on "Ignore Cache"
Change your avatar else where
Tap on "Fetch Avatar" on the UIKit demo app
It should fetch the most up to date avatar

### DIFF
--- a/Sources/Gravatar/Network/Services/ImageDownloadService.swift
+++ b/Sources/Gravatar/Network/Services/ImageDownloadService.swift
@@ -95,7 +95,15 @@ public struct ImageDownloadService: ImageDownloader, Sendable {
 
 extension URLRequest {
     fileprivate static func imageRequest(url: URL, forceRefresh: Bool) -> URLRequest {
-        var request = forceRefresh ? URLRequest(url: url, cachePolicy: .reloadIgnoringLocalCacheData) : URLRequest(url: url)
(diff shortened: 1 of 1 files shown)


etoledom (approved):
Works as described 👍

PR #442, opened 2024-09-30
Set "Accept : application/json" header

### LINKED ISSUE #422: Quick Editor: Add `Accept`:`application/json` header to REST API requests
Our BE friends asked to add this header.

pinarol:
Closes #422 

### Description

Sets the `Accept=application/json` header on the URLSessionConfiguration. This value can be overridden at the URLRequest level so our image download request also works fine. 

### Testing Steps

Check with a tool like proxyman and observe the header.


andrewdmontgomery:
Doing this in our own `URLSessionsHTTPClient` will mean that this header won't be set if a third-party supplies their own `HTTPClient`.

We chatted before, and you basically said what I was thinking; supporting third-party `HTTPClient` injection may be more trouble than it's worth.

So 👍

etoledom (approved):

PR #444, opened 2024-09-30
Decrease the max crop size

### LINKED ISSUE #438: QE: Decrease the max crop size
Our new cropper's max output size is 2048 x 2048 pixels. Having tried things a bit, it can take a bit long to upload such image. Let's try a smaller value like 1500 or so.

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/438

### Description

The existing max crop size(2048 x 2048) was selected arbitrarily. Now that i had time to test the uploads, it can take a bit long to upload such image, it seems like real device pictures can be big. I think this new value is better but let me know what you think.

### Testing Steps

You can test the "UIKit demo app > Image cropper" example page(only visible in DEBUG config) to crop an image bigger than (1536 x 1536) and observe that the cropped image is 1536 x 1536 (scale: 1).

You can also try "SwiftUI Demo app > avatar picker" to upload an image and see if the upload time feels better.

### DIFF
--- a/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
+++ b/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
@@ -4,7 +4,7 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
     private enum Constants {
         static let backgroundColor = UIColor.black
         static let croperFrameSize: CGFloat = 320
-        static let maxOutputImageSizeInPixels: CGFloat = 1536
+        static let maxOutputImageSizeInPixels: CGFloat = 1280
     }
 
     // ScrollView for zooming and panning
--- a/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
+++ b/Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
@@ -4,7 +4,7 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
     private enum Constants {
         static let backgroundColor = UIColor.black
         static let croperFrameSize: CGFloat = 320
-        static let maxOutputImageSizeInPixels: CGFloat = 2048
+        static let maxOutputImageSizeInPixels: CGFloat = 1536
     }
 
     // ScrollView for zooming and panning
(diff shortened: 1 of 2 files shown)


Sources/GravatarUI/ImageCropper/ImageCropperViewController.swift
```diff
@@ -4,7 +4,7 @@ class ImageCropperViewController: UIViewController, UIScrollViewDelegate {
     private enum Constants {
         static let backgroundColor = UIColor.black
         static let croperFrameSize: CGFloat = 320
-        static let maxOutputImageSizeInPixels: CGFloat = 2048
+        static let maxOutputImageSizeInPixels: CGFloat = 1536
```
pinarol: Why 1536? It is just in the middle of 1024 and 2048.

etoledom: If the backend accepts files of 2048 in dimension, I guess the client should allow it also? 🤔 

It's gonna be slower upload as the file size is bigger, but that's normal.

I tested some images on trunk:
- 2024px dimensions was 820KB in size.
- 2000px dimensions was 1.5MB in size.
- 2024px dimensions was 2MB in size.

Same images with the changes on this PR:
- 1536px dimensions was 470KB in size.
- 1536px dimensions was 1MB in size.
- 1536px dimensions was 1.2MB in size.

As an alternative, from trunk but reducing JPEG compression to 0.8:
- 2024px dimensions was 600KB in size.
- 2000px dimensions was 1.3MB in size.
- 2048px dimensions was 1.6MB in size.

Of course it will vary with the image content but I don't think it would get much bigger. 

**Note:** I didn't notice any difference in quality in any of the tests mentioned above.

Maybe in the future we can improve the cropper to add an "HD" and Lower size options instead, so it's the user decision if they want to upload fast or better quality.

pinarol: We discussed things off PR a bit. Seems like the JPEG images have high processing times on the BE side. So we are reducing the size to 1280, quality to 0.8. Still not as fast as we want but it was taking ~10-12 seconds before, so a nice improvement. We'd like to reduce the BE processing times in the meantime.

Screenshot 2024-10-02 at 14 14 21

etoledom (approved):
Thank you! 🙏

PR #445, opened 2024-09-30
Add color scheme picker to the SwiftUI demo app

pinarol:
Closes #

### Description

Just wanted to add a color scheme option to the QE demo pages for easier testing.

### Testing Steps

SwiftUI demo app, these pages now have color scheme options:

- Avatar Picker View
- Profile editor with oauth


etoledom (approved):
Looking good!

As an idea:

To avoid using `nil` with a message.
```swift
@State private var selectedScheme: ColorScheme? = nil // nil means it will use the system default
```

We could use `UIUserInterfaceStyle` instead.
`UIUserInterfaceStyle.unspecified` already declares that `Choose this option when you want to follow the system's interface style.`

https://github.com/Automattic/Gravatar-SDK-iOS/compare/wppinar/add-color-option-to-demo-app...etoledom/wppinar/add-color-option-to-demo-app?expand=1

Not a blocker though, so approving anyway 👍

pinarol:
Nice suggestion. I merged your branch into this one. Thanks!

PR #448, opened 2024-10-02
Handle expired token (401 error) on avatar selection and upload

### LINKED ISSUE #447: QE: Handle Avatar selection with expired token
Selecting an avatar with an expired token gives a generic error and no ability to re-auth.

andrewdmontgomery:
Closes #447 #453

### Description

If the QE is open while a token expires, the failure message is generic and doesn't provide a way for the user to re-authenticate.

This PR updates the UX, and error message handling, so that the user is informed that their session has expired, and they are presented with the ability to re-authenticate.

This PR also makes the following change:

```diff
    func map() -> ResponseErrorReason {
        switch self {
        case .URLSessionError(let error):
            return .URLSessionError(error: error)
        case .invalidHTTPStatusCodeError(let response, let data):
-          if response.statusCode == 400 {
+          if response.statusCode >= 400 {
                let error: ModelError? = try? data.decode()
                return .invalidHTTPStatusCode(response: response, errorPayload: error)
            } else {
                return .invalidHTTPStatusCode(response: response)
            }
        case .invalidURLResponseError(let response):
            return .invalidURLResponse(response: response)
        }
    }
```

This is _technically_ accurate (all 400+ codes are "invalid".  But there may be a good reason why we defined this as `==`.  So I wanted to call this out.

This PR also localizes one of the toast messages.  It looks like other toast messages still need to be localized.  I'll handle those in a separate PR.

### Updating the OpenAPI spec

This PR now includes an updated OpenAPI spec that addresses the issue mentioned here.

~_This PR temporarily updates the OpenAPI.spec – but only as a demonstration.  This change to the OpenAPI spec should not be merged unless and until we have discussed it with the Backend team and Android._~

~When an expired token is used on the endpoint 
`/me/avatars/{imageId}/email`, the json response only includes the `error` key/value (with no business `code` provided):~

```json
{
  "error": "You are not authorized to perform this action."
}
```

~We are currently using `ModelError` to represent this response, which is a model generated from the `Error` schema in the OpenAPI spec.  That spec lists both the `code` and `error` as "required".~

~As a result, this JSON data in the response fails to decode as a `ModelError`, and we cannot pass the message to the client.~

~If using `ModelError` (`Error` schema) is appropriate for this endpoint, then either:~
- ~The backend needs include a `code`~
- ~Or the OpenAPI spec needs to specify the `code` as optional~

~If using the `ModelError` is not appropriate, then we have some refactoring to do.~

#### Testing Avatar Selection
1. Launch the SwiftUI demo app
2. Tap `Profile editor with oauth`
3. Perform Oauth if necessary
4. On your computer:
   a. Go here (with your test app Gravatar account): https://wordpress.com/me/security/connected-applications
   b. Click "Disconnect" for your test app
5. Back on your device, attempt to choose a different avatar
6. OBSERVE:
   - [ ] A toast message appears saying, `You are not authorized to perform this action.` (This string is provided by the backend in the response, and I believe it should be localized)
   - [ ] The grid of avatars is replaced with the error message "Session expired" and a "Log in" button

Simulator Screenshot - iPhone SE (3rd generation) - 2024-10-01 at 22 07 23

#### Testing Avatar Upload
1. Follow the steps 1-4 above
2. Back on your device, attempt to upload a new image
3. OBSERVE:
   - [ ] A toast message appears saying, `You are not authorized to perform this action.` (This string is provided by the backend in the response, and I believe it should be localized)
   - [ ] The grid of avatars is replaced with the error message "Session expired" and a "Log in" button


pinarol:
> This is technically accurate (all 400+ codes are "invalid". But there may be a good reason why we defined this as ==. So I wanted to call this out.

That check is for deciding to decode the error into `ModelError` and we can only assume this is possible if the specs say so. Since we treat the decoding operation as "optional" we can probably move fwd with >=400 as well. But we can probably make it ">=400 and <500", i don't think 5XX errors are any fit for the `ModelError` anyway.

pinarol:
> This PR temporarily updates the OpenAPI.spec – but only as a demonstration. This change to the OpenAPI spec should not be merged unless and until we have discussed it with the Backend team and Android.

👍

andrewdmontgomery:
The proposed change to the `Error` schema was accepted.  A more comprehensive change is pending: https://github.a8c.com/Automattic/gravatar/pull/109013

I will wait for that PR to merge, and apply those changes to this PR.

etoledom:
> The proposed change to the `Error` schema was accepted. A more comprehensive change is pending: https://github.a8c.com/Automattic/gravatar/pull/109013
> 
> I will wait for that PR to merge, and apply those changes to this PR.

It has been merged 🎉

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -104,6 +104,12 @@ class AvatarPickerViewModel: ObservableObject {
             selectedAvatarResult = .success(response.imageId)
         } catch APIError.responseError(let reason) where reason.cancelled {
             // NoOp.
+        } catch APIError.responseError(let .invalidHTTPStatusCode(response, errorPayload)) where response.statusCode == 401 {
+            handleUnrecoverableClientError(
+                errorMessage: errorPayload?.message ?? Localized.avatarUpdateFail,
```
pinarol: We shouldn't show the server error to the user here because it is neither localized nor user friendly. Only upload errors are localized (and that's only when statusCode == 400). We can just pass `Localized.avatarUpdateFail`.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -185,6 +191,13 @@ class AvatarPickerViewModel: ObservableObject {
                 supportsRetry: false,
                 errorMessage: errorPayload?.message ?? Localized.genericUploadError
             )
+        } catch ImageUploadError.responseError(reason: let .invalidHTTPStatusCode(response, errorPayload)) where response.statusCode == 401 {
+            // If the status code is 401, then it means the token is not valid and we should prompt the user accordingly.
+            handleUnrecoverableClientError(
+                errorMessage: errorPayload?.message ?? Localized.genericUploadError,
```
pinarol: Same. We can just pass `Localized.genericUploadError`.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -104,6 +104,12 @@ class AvatarPickerViewModel: ObservableObject {
             selectedAvatarResult = .success(response.imageId)
         } catch APIError.responseError(let reason) where reason.cancelled {
             // NoOp.
+        } catch APIError.responseError(let .invalidHTTPStatusCode(response, errorPayload)) where response.statusCode == 401 {
```
pinarol: nit: instead of hardcoding 401 you can use `HTTPStatus.unauthorized.rawValue`

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -185,6 +191,13 @@ class AvatarPickerViewModel: ObservableObject {
                 supportsRetry: false,
                 errorMessage: errorPayload?.message ?? Localized.genericUploadError
             )
+        } catch ImageUploadError.responseError(reason: let .invalidHTTPStatusCode(response, errorPayload)) where response.statusCode == 401 {
```
pinarol: same nit. `HTTPStatus.unauthorized.rawValue`

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -211,6 +224,22 @@ class AvatarPickerViewModel: ObservableObject {
         grid.replaceModel(withID: imageID, with: newModel)
     }
 
+    private func handleUnrecoverableClientError(errorMessage: String? = nil, response: HTTPURLResponse, errorPayload: APIErrorPayload? = nil) {
+        if let errorMessage {
+            toastManager.showToast(errorMessage, type: .error)
```
pinarol: Actually, do we even need to show the toast for this case? There's no need imo.

andrewdmontgomery: I wasn't sure.  I wanted to see how it looked.  But yeah, I agree.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
andrewdmontgomery: You mentioned below not using a toast for these, and I like that.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
andrewdmontgomery: Yeah, the message from the backend is terrible.  I had thought these were eventually going to be improved and localized, which is why I tried these toasts.  I hadn't realized that was only happening for the `400` responses.

But as you pointed out below, the toasts are redundant anyway.

pinarol (approved):
LGTM 👍

PR #450, opened 2024-10-02
Adding a configuration for the global quick editor experience

etoledom:
Closes #

### Description

Adding a configuration for the global quick editor experience
So far we are only applying a configuration to override the system color scheme

### Testing Steps

- Run the UIKit demo app
- Chose the Quick Editor demo screen
- Select a different color scheme from the current system 
- Open the Quick Editor
  - Check that the Quick Editor respect the selected color scheme.


pinarol (approved):
Looks and works good 🎉

PR #452, opened 2024-10-02
Updating open-api specs

etoledom:
Closes #

### Description

Updating to the latest specs

### Testing Steps

The change could affect the `CheckTokenAuthorizationService`. To test, follow the steps from: https://github.com/Automattic/Gravatar-SDK-iOS/pull/427

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,7 @@ generate: $(OUTPUT_DIRECTORY) # Generates the open-api model
 	cp "$(OPENAPI_YAML_PATH)" "$(OPENAPI_GENERATOR_CLONE_DIR)"/openapi.yaml
 	mkdir -p "$(OPENAPI_GENERATOR_CLONE_DIR)"/templates
 	cp "$(MODEL_TEMPLATE_PATH)"/*.mustache "$(OPENAPI_GENERATOR_CLONE_DIR)"/templates/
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -112,6 +112,7 @@ generate: $(OUTPUT_DIRECTORY) # Generates the open-api model
 	cp "$(OPENAPI_YAML_PATH)" "$(OPENAPI_GENERATOR_CLONE_DIR)"/openapi.yaml
 	mkdir -p "$(OPENAPI_GENERATOR_CLONE_DIR)"/templates
 	cp "$(MODEL_TEMPLATE_PATH)"/*.mustache "$(OPENAPI_GENERATOR_CLONE_DIR)"/templates/
+	rm -rf "$(OPENAPI_GENERATOR_CLONE_DIR)"/generated/OpenAPIClient/Classes/OpenAPIs/Models/*
```
pinarol, line 115: This is intended right?

pinarol (approved):
LGTM 👍

Makefile
etoledom: Yes, forgot to comment here.

I noticed that the `AssociatedEmail200Response` file wasn't removed the first time I run the script. This change is intended to fix this.

Since it is not anymore part of the spec, the not-longer generated file should be also removed.

PR #455, opened 2024-10-03
Add common headers for Platform, SDK version, Source app

### LINKED ISSUE #451: Add some additional headers for all requests
`X-Platform: ios`
`X-SDK-Version: 2.1.0`

if it's easy let's also include:

`X-Source: <app bundle name>` 

Also, let's remove the existing `Client-Type` header, it's not used.

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/451

### Description

Adding headers:

`X-Platform: ios`
`X-SDK-Version: <version>`
`X-Source: <app bundle name>`

I created a follow-up issue to find a better way about injecting the SDK version into the iOS runtime. https://github.com/Automattic/Gravatar-SDK-iOS/issues/456

### Testing Steps

SwiftUI demo app > QE 
Check with a tool like proxyman and confirm the headers are added.

### DIFF
--- a/Sources/Gravatar/Bundle+ResourceBundle.swift
+++ b/Sources/Gravatar/Bundle+ResourceBundle.swift
@@ -1,0 +1,18 @@
+import Foundation
+
+#if !SWIFT_PACKAGE
+private class BundleFinder: NSObject {}
+extension Bundle {
+    static var module: Bundle {
+        let defaultBundle = Bundle(for: BundleFinder.self)
+        // If installed with CocoaPods, resources will be in GravatarUI.bundle
+        if let bundleURL = defaultBundle.resourceURL,
+           let resourceBundle = Bundle(url: bundleURL.appendingPathComponent("Gravatar.bundle"))
+        {
+            return resourceBundle
+        }
+        // Otherwise, the default bundle is used for resources
+        return defaultBundle
+    }
+}
+#endif
--- a/Sources/Gravatar/Bundle+ResourceBundle.swift
+++ b/Sources/Gravatar/Bundle+ResourceBundle.swift
@@ -5,7 +5,10 @@ private class BundleFinder: NSObject {}
 extension Bundle {
     static var module: Bundle {
         let defaultBundle = Bundle(for: BundleFinder.self)
-        // If installed with CocoaPods, resources will be in GravatarUI.bundle
+        // If installed with CocoaPods, resources will be in Gravatar.bundle
+        // The name of the bundle "Gravatar.bundle" (without the .bundle file extension)
+        // needs to match the key in the respective Gravatar.podspec:
+        // `s.resource_bundles = { 'Gravatar' => ['Sources/Gravatar/Resources/*.plist'] }`
         if let bundleURL = defaultBundle.resourceURL,
            let resourceBundle = Bundle(url: bundleURL.appendingPathComponent("Gravatar.bundle"))
         {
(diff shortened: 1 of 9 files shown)


etoledom (approved):
I see the correct parameters on the Headers 🎉 
CleanShot 2024-10-03 at 19 17 53@2x

Sources/Gravatar/Network/Services/URLSessionHTTPClient.swift
```diff
@@ -14,6 +14,9 @@ struct URLSessionHTTPClient: HTTPClient {
         let configuration = URLSessionConfiguration.default
         configuration.httpAdditionalHeaders = [
             "Accept": "application/json",
+            "X-Platform": "ios",
```
andrewdmontgomery, line 17: This reminds me of a separate question: do we want to track iOS/iPadOS?  If so, I think we'd still treat them both as `X-Platform ios`, and add a separate header for device form factor.  That way both platforms could record this.

We could add a separate header like `"X-Device-Type: tablet"`.

Sources/Gravatar/Bundle+ResourceBundle.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+#if !SWIFT_PACKAGE
+private class BundleFinder: NSObject {}
+extension Bundle {
+    static var module: Bundle {
+        let defaultBundle = Bundle(for: BundleFinder.self)
+        // If installed with CocoaPods, resources will be in GravatarUI.bundle
+        if let bundleURL = defaultBundle.resourceURL,
+           let resourceBundle = Bundle(url: bundleURL.appendingPathComponent("Gravatar.bundle"))
```
andrewdmontgomery, line 13: Nice.

For those visiting this in the future:

The name of the bundle `"Gravatar.bundle"` (without the `.bundle` file extension) needs to match the key in the respective `Gravatar.podspec`:

```ruby
 s.resource_bundles = {
    'Gravatar' => ['Sources/Gravatar/Resources/*.plist']
  }
```

We happen to be using the same name ("Gravatar") for the module and the pod, too.  So it's easier to get right.  But the actual dependency is on the name of the resource bundle we define in the podspec.

Sources/Gravatar/Bundle+ResourceBundle.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+#if !SWIFT_PACKAGE
+private class BundleFinder: NSObject {}
+extension Bundle {
+    static var module: Bundle {
+        let defaultBundle = Bundle(for: BundleFinder.self)
+        // If installed with CocoaPods, resources will be in GravatarUI.bundle
```
andrewdmontgomery:
```suggestion
        // If installed with CocoaPods, resources will be in Gravatar.bundle
```

andrewdmontgomery (approved):
Just a small change to a comment.  Looks good.

Sources/Gravatar/Network/Services/URLSessionHTTPClient.swift
pinarol: For now we don't want to track iOS/iPadOS and yes i agree that would be a separate header if needed.

Sources/Gravatar/Bundle+ResourceBundle.swift
pinarol: Thanks for elaborating. This is valuable info so i added it in the code comment.

Sources/Gravatar/Bundle+ResourceBundle.swift
pinarol: Done.

PR #460, opened 2024-10-04
Prevent stacking of toast messages

### LINKED ISSUE #457: QE: Prevent "Avatar updated! ..." toasts from stacking
A request from @Luchadores 

When we change avatars quickly "Avatar updated! ..." toast messages stack too much. Can we make it so that "if one confirmation is already presented, we don’t show another one until it disappears"?

pinarol:
Closes #457 

### Description

Upon a request from our designer, i implemented a mechanism to prevent the toasts from stacking. If the new toast has the same message, the previous one will be dismissed first. 

Oct-04-2024 13-51-47

### Testing Steps

SwiftUI demo app > Profile editor with oauth
Switch between different avatars and observe the you see only 1 success toast at a time.
Disconnect from internet, try the same thing and observe it for the fail message as well

### DIFF
--- a/Sources/GravatarUI/SwiftUI/Toast/ToastManager.swift
+++ b/Sources/GravatarUI/SwiftUI/Toast/ToastManager.swift
@@ -4,11 +4,25 @@ import SwiftUI
 class ToastManager: ObservableObject {
     @Published var toasts: [ToastItem] = []
 
-    func showToast(_ message: String, type: ToastType = .info) {
-        let toast = ToastItem(message: message, type: type)
+    func showToast(_ message: String, type: ToastType = .info, stackingBehavior: ToastStackingBehavior = .dismissExistingWithSameMessage) {
+        let toast = ToastItem(message: message, type: type, stackingBehavior: stackingBehavior)
+        dismissExistingIfNeeded(upcomingToast: toast)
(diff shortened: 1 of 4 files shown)


Sources/GravatarUI/SwiftUI/Toast/ToastManager.swift
```diff
@@ -30,8 +44,16 @@ enum ToastType: Int {
     case error
 }
 
+enum ToastStackingBehavior: Equatable {
+    /// Dismisses the toast with the same message before showing the new one.
+    case dismissExistingWithSameMessage
+    /// Doesn't dismiss.
+    case doNotDismiss
+}
```
etoledom: Nit: As the enum name refers to `stacking`, maybe the cases also should refer to stacking. For example:
```
case avoidStackingWithSameMessage
case alwaysStack
// possible future
case neverStack
```

Not a blocker. Up to you to change if you like the idea.

etoledom (approved):
Looking good!

Sources/GravatarUI/SwiftUI/Toast/ToastManager.swift
pinarol: makes sense let me check.

PR #462, opened 2024-10-04
Making some openapi structs internal

etoledom:
Closes #

### Description

Making some of the openapi structs internal.

`Avatar` struct is being used by:
```swift
public func upload(_ image: UIImage, accessToken: String) async throws -> Avatar
```

I made a second commit replacing it with a public protocol, to consider and discuss.

### Testing Steps
- CI should be 🍏 
- Avatar picker should work as expected

### DIFF
--- a/Sources/Gravatar/Network/Services/AvatarService.swift
+++ b/Sources/Gravatar/Network/Services/AvatarService.swift
@@ -55,12 +55,25 @@ public struct AvatarService: Sendable {
     /// - Parameters:
     ///   - image: The image to be uploaded.
     ///   - accessToken: The authentication token for the user. This is a WordPress.com OAuth2 access token.
-    /// - Returns: An asynchronously-delivered `AvatarModel` instance, containing data of the newly created avatar.
+    /// - Returns: An asynchronously-delivered `AvatarType` instance, containing data of the newly created avatar.
     @discardableResult
-    public func upload(_ image: UIImage, accessToken: String) async throws -> Avatar {
+    public func upload(_ image: UIImage, accessToken: String) async throws -> AvatarType {
+        let avatar: Avatar = try await upload(image, accessToken: accessToken)
+        return avatar
+    }
+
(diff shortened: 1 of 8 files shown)


Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -20,6 +20,18 @@ public struct AvatarService: Sendable {
...
+    ///
+    /// Optionally, you can pass a custom type conforming to ``URLSessionProtocol``.
+    /// Similarly, you can pass a custom type conforming to ``ImageCaching`` to use your custom caching system.
+    /// - Parameters:
+    ///   - session: A type which will perform basic networking operations. By default, a properly configured URLSession instance will be used.
+    ///   - cache: A type which will perform image caching operations.
+    public init(session: URLSessionProtocol? = nil, cache: ImageCaching? = nil) {
+        let client = URLSessionHTTPClient(urlSession: session)
+        self.init(client: client, cache: cache)
+    }
```
etoledom, line 33: This will help tests to not need to access the internal `URLSessionHTTPClient` type.

Since the internal and public `upload()` methods have to be literally typed to avoid confusion, this change allows to avoid `@testable` import of Gravatar.

A simpler solution would be to add the type to the test, but I thought is better to avoid `@testable` import if possible.

I don't particularly like that users are able to bypass our session configuration (specially the custom headers), but users already were able to use a different session or any other arbitrary way to do networking via injecting an `HTTPClient` conforming type.

pinarol: I agree and I think we can make HTTPClient internal. Things won't work with a different HTTPClient implementation properly anyway. For example: We rely on the error handling that happens in `URLSessionHTTPClient`. 

Allowing to pass a custom URLSession is the way to go i think.

pinarol (approved):
Looks good 🎉

PR #459, opened 2024-10-03
Avatar selected callback

etoledom:
Closes #

### Description

This PR adds a callback to signal the parent app of a new avatar selection by the user.

#### UIKit:
- Go to Quick Editor example page
- Open the Avatar Picker and select a different avatar
  - Check that the avatar on the demo screen updates acordingly

#### SwiftUI:
- Go to Profile Editor with OAuth
- Open the Avatar Picker and select a different avatar
  - Check that the avatar on the demo screen updates acordingly

PR #463, opened 2024-10-04
Add delay to avatar update callback

pinarol:
Continuation of #459 

### Description

Adding a delay to wait until the server has updated the selected avatar before updating the UI. Without the delay the cache busting remains insufficient to capture the new avatar.
With less than 800 ms, we can still see the issue (see [comment](https://github.com/Automattic/Gravatar-SDK-Android/pull/350/files#r1787548797)).
Hopefully, we can remove this delay soon.

### Testing Steps

Use the steps in #459 

Check if the avatar in the demo app gets refreshed


etoledom (approved):
0.8s sounds like a lot 🤦‍♂️ 
But it's not a main user interaction, and possibly in many cases the 3rd party avatar image won't be visible to the user, so I guess is fine as a temporary workaround.

PR #466, opened 2024-10-04
Prevent the session expired error state from changing on tap

### LINKED ISSUE #461: Session expired error state changes on tap to login
Steps:

- Expire your existing session https://wordpress.com/me/security/connected-applications
- Try opening QE with oauth flow
- You should see the session expired error state, so tap on login

Observe: The session expired message changes into Login required
Expected: The session expired message remains same

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/461

### Description

The session expired error state changes immediately after tapping on Login. It changes into Login required state. This looks a bit odd. Please try the testing steps on a different branch to check the problem.

It's happening because we remove the token immediately after tapping on Login. I am trying a different approach. Instead of removing the token i mark it as expired. Let's see how this works and whether if it's worth the extra logic or not.

### Testing Steps

- Disconnect your existing session https://wordpress.com/me/security/connected-applications
- Try opening QE with oauth flow
- You should see the session expired error state, tap login

Observe: The session expired message remains same


etoledom (commented):
I see that the UI won't update after re-authenticating:

CleanShot 2024-10-07 at 11 04 01

pinarol:
I'll check. I must have broken something last minute.

Sources/GravatarUI/SwiftUI/OAuthSession/KeychainToken.swift
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    var toData: Data? {
+        let encoder = JSONEncoder()
+        do {
+            let data = try encoder.encode(self)
+            return data
+        } catch {
+            return nil
+        }
```
etoledom: If we only need to return nil, we can do `try?` (same on the init)
```suggestion
        let encoder = JSONEncoder()
        return try? encoder.encode(self)
```

Sources/GravatarUI/SwiftUI/OAuthSession/KeychainToken.swift
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    init(token: String) {
+        self.token = token
+    }
+
+    var toData: Data? {
```
etoledom: nit: Since it's a field (instead of a method), this could be called `data`.

Sources/GravatarUI/SwiftUI/OAuthSession/KeychainToken.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+struct KeychainToken: Codable {
+    let token: String
+    var isExpired: Bool = false
```
etoledom, line 5: An idea: Just to try keeping immutability, we could have a `func expire() -> KeychainToken` (or `toExpired()`) which creates a new instance with this property set to true.

etoledom (approved):
Looks good and works as intended! 🎉 

Just a few suggestions on code. Nothing blocking 🚢

Sources/GravatarUI/SwiftUI/OAuthSession/KeychainToken.swift
pinarol: We discussed this a bit off PR, since KeychainToken consists of value types for now we'll keep it as is.

PR #467, opened 2024-10-04
Handle payload too large error on upload

### LINKED ISSUE #436: QE: Handle 413 HTTP status for image too large error
When uploading large images(> 10mb) the error response comes back as an HTML document, regardless of the Accept header. It seems this isn’t coming from the PHP app and it is something needs to be fixed at the BE. But unfortunately we can't count on it. So until this is resolved we are told that we could interpret 413 HTTP status code responses (for this endpoint) as being too large of an image and hardcode the error message in the client.

pinarol:
Closes #436 

### Description

For the reasons explained in  #436 we need to handle 413(payloadTooLarge) error on the client side to show the proper message.

### Testing Steps

You need to manipulate the code and set `maxOutputImageSizeInPixels` to a value bigger than 20000.
Again manipulate the code to upload png instead of jpeg.

You can try with `32MBImage.png` from https://drive.google.com/drive/u/0/folders/16bRuYugP99X3grDlh938ffJaomlfFXaT
Try to upload it and you should see the error "The provided image exceeds the maximum size: 10MB"

Regression test:

Try again with `20008x20008.png` from the same folder.
Observe that it gives you the dimension error this time.


etoledom (approved):
Works as described 🎉 

CleanShot 2024-10-07 at 10 25 02@2x

PR #301, opened 2024-07-01
Merge "Release/2.0.1" to trunk

pinarol:
Closes #

### Description

Merge the release branch back to trunk.

### Testing Steps

CI green.

PR #287, opened 2024-06-13
Merge release/2.0.0 into trunk

pinarol:
Closes #

### Description

Merge release/2.0.0 into trunk

### Testing Steps

CI green.

PR #471, opened 2024-10-07
Release/2.0.2 additions back merge

andrewdmontgomery:
Closes #

### Description

This back merges commits from `release/2.0.2-additions` into `trunk`.  Note that this PR only contains commits.  There is no file diff.  But some of these commits contain the `tag` for versions of our release.  So they are missing from `trunk`.

Commits/tags:
- [07f04e9](https://github.com/Automattic/Gravatar-SDK-iOS/commit/07f04e9568dcb28e2c56b988cdbb1aeb28ded940): `2.0.1`
- [4deb714]( https://github.com/Automattic/Gravatar-SDK-iOS/commit/4deb7142cdbb25b6148e76f189025fcbcbf0705c): `2.0.2`
- [337a2b6](https://github.com/Automattic/Gravatar-SDK-iOS/commit/337a2b639ecafc5159eb0d2af49391a95ad1637a): `2.1.0`
- [08c63e5](https://github.com/Automattic/Gravatar-SDK-iOS/commit/08c63e532e957419361ce2dcdcf44f61881d0cd8): `2.1.1`

There were a few conflicts, but they were easy to resolve (just accepting the trunk changes was appropriate).

### Testing Steps

- [ ] CI should be 🟢


andrewdmontgomery:
When this PR merges, do we want to delete the [release/2.0.2-additions](https://github.com/Automattic/Gravatar-SDK-iOS/tree/release/2.0.2-additions) branch?

pinarol:
I guess we can merge this but i wanted to mention, we already merged each release back to trunk:

https://github.com/Automattic/Gravatar-SDK-iOS/pull/394 
https://github.com/Automattic/Gravatar-SDK-iOS/pull/305
https://github.com/Automattic/Gravatar-SDK-iOS/pull/301
https://github.com/Automattic/Gravatar-SDK-iOS/pull/287

i feel like this PR is doing the same thing?

pinarol:
> When this PR merges, do we want to delete the [release/2.0.2-additions](https://github.com/Automattic/Gravatar-SDK-iOS/tree/release/2.0.2-additions) branch?

We can probably delete it, i can't think of a reason to keep it.

pinarol (approved):
LGTM 👍 
I'll merge it to include in the release branch.

andrewdmontgomery:
> We can probably delete it, i can't think of a reason to keep it.

Agreed.  I'll delete it.

andrewdmontgomery:
> I guess we can merge this but i wanted to mention, we already merged each release back to trunk:
> 
> #394 #305 #301 #287
> 
> i feel like this PR is doing the same thing?

@pinarol 

True, these PR's were definitely merged.  But the commits in those PR's **_that were tagged_** did not merge to `trunk` as part of those merges.  They were orphaned.

I'm not sure why, but my best guess is that these PR's were **_squash/merged_**, which would create and merge a new commit, leaving the original commits only in the original branch, which was then deleted, leaving the tagged commits orphaned.

This PR was meant to merge those commits into `trunk`, so that the tagged commits are no longer orphaned.

Before this PR:

```shell
$ git checkout trunk
$ git fetch --tags
$ git tag --merged  # only shows tags available from the current branch
0.1.0
1.0.0
1.0.1
1.0.1-alpha
2.0.0
2.0.0-rc.1
```

But it looks like this PR might have been squash/merged also.  So the commits with these tags still aren't in `trunk`:

```shell
$ git checkout trunk
$ get fetch
$ git log trunk | grep 07f04e9

# no result
```

I'll get another PR up with the same fix.

PR #475, opened 2024-10-08
Remove open-api deprecations

etoledom:
Closes #

### Description

Removing open-api deprecations and its custom generator code.

The only things left on manual up-keeping are the access control changes.

### Testing Steps

- Run the UIKit demo app
- Test the Profile Fetch demo screen
- Run the SwiftUI demo app
- Test the Quick Editor demo app


pinarol (approved):
Looks and works good 👍

PR #479, opened 2024-10-09
Fix RemoteSVGView crash

etoledom:
Closes #

### Description

Fixes a crash when opening the RemoteSVGView demo screen on UIKit demo app.

### Testing Steps

- Run UIKit demo app
- Open the RemoteSVGView demo screen.
  - Check that the icons load as expected.


pinarol (approved):
LGTM 👍

Issue #449, opened 2024-10-02
Remove deprecated types


pinarol:
Now that we had to do some public API changes for concurrency fixes, it can be a good time to remove other deprecated public types as well.

PR #482, opened 2024-10-09
Make HTTPClient internal

pinarol:
Closes part of #449 

### Description

Making `HTTPClient` internal since a custom `HTTPClient` implementation will not work properly anyway. This is mostly because of how we handle and map errors. There's still a way to mock the `URLSession` via `URLSessionProtocol`.

### Testing Steps

Smoke test:

SwiftUI demo app > Profile editor with oauth
UIKit demo app > Profile cards


etoledom (approved):
Looks great!
Smoke tests all good 👍

PR #483, opened 2024-10-09
Add localization progress check and linting of localized strings

andrewdmontgomery:
Closes #

### Description

This adds two checks when downloading localizations:
1. Adds a progress check before downloading
   - If any localizations are less than 100% completed, you are prompted to continue or stop
2. Adds linting after localizations are downloaded:
   - Checks for duplicate keys (which shouldn't happen for us, since we are generating our `.strings` file from source)
   - Compares the localizations against the base language to find potential mismatches for the %s/%d/… parameter placeholders between locales.

### Testing Steps
Both of these functions are provided by the [wpmreleasetoolkit](https://github.com/wordpress-mobile/release-toolkit) fastlane plugin:
- [ios_lint_localizations](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_lint_localizations.rb)
- [check_translation_progress](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_translation_progress.rb)

#### Testing the Progress Check
At the moment, our translations are below 100%.  So you can test the first part easily:
1. Run `make download-translations`
- [ ] OBSERVE that you see a prompt like the following:
```
The translations for the following languages are below the 100% threshold:
 - ar is at 55%.
 - de is at 55%.
 - es is at 55%.
 - fr is at 55%.
 - he is at 55%.
 - id is at 55%.
 - it is at 55%.
 - ja is at 55%.
 - ko is at 55%.
 - nl is at 55%.
 - pt-br is at 55%.
 - ru is at 55%.
 - sv is at 55%.
 - tr is at 55%.
 - zh-cn is at 55%.
 - zh-tw is at 55%.
Do you want to continue? (y/n)
```

#### Testing Localization Linting
This requires creating a linting problem.  Since we don't have one, you will have to create one, and then manually comment out the download portion of the lane:

1. Edit one of the non-English `Localizable.strings` files by creating a string with a number of placeholders that does NOT match the same string in the english `Localizable.strings` file.  At the moment, none of our source strings with placeholders have been localized.  So you will have to add one.  You can use the one below:
_Note: this sample string already has an extra `%@` placeholder (the original has one, this has two).  So you can paste this directly into one of the non-English `Localizable.strings` files._
```
/* A message describing the error and advising the user to login again to resolve the issue */
"AvatarPicker.ContentLoading.Failure.WrongEmailError.subtext" = "It looks like you used the wrong email to log in. Please try again using %@ %@ this time. Thanks!";
```
2. In the `fastlane/lanes/localization.rb` file, in the `:download_localized_strings` lane, comment out two functions (lines 67-76):
```ruby
   SOURCES_TO_LOCALIZE.each do |source|
      next if source.gp_project_url.nil?

      # check_translation_progress(
      #   glotpress_url: source.gp_project_url,
      #   abort_on_violations: false
      # )

      # ios_download_strings_files_from_glotpress(
      #   project_url: source.gp_project_url,
      #   locales: GLOTPRESS_TO_LPROJ_APP_LOCALE_CODES,
      #   download_dir: source.localizations_root
      # )

      ios_lint_localizations(
        input_dir: source.localizations_root
      )

      next if skip_commit

      paths_to_commit << source.localizations_root
    end
```
3. Save those changes
4. Run `make download-strings`
- [ ] OBSERVE that you see a failure with an error like the following:
```
[12:30:29]: Inconsistencies found between 'en' and 'fr':

`AvatarPicker.ContentLoading.Failure.WrongEmailError.subtext` expected placeholders for [String] but found [String,String] instead.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,9 @@ fastlane/README.md
 fastlane/report.xml
 fastlane/test_output
 
+# SwiftGen (part of wpmreleasetoolkit)
+vendor/swiftgen
+
 # Other
 openapi-generator/
 Demo/Demo/Secrets.swift
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -14,7 +14,7 @@ XCODEPROJ_PATH = File.join(PROJECT_ROOT_FOLDER, 'Gravatar-Demo.xcodeproj')
 DEMO_APPS_SOURCES_FOLDER = File.join(PROJECT_ROOT_FOLDER, 'Demo')
 XCCONFIG_PROTOTYPE_BUILD_SWIFTUI = File.join(DEMO_APPS_SOURCES_FOLDER, 'Gravatar-SwiftUI-Demo', 'Gravatar-SwiftUI-Demo.Release.xcconfig')
 XCCONFIG_PROTOTYPE_BUILD_UIKIT = File.join(DEMO_APPS_SOURCES_FOLDER, 'Gravatar-UIKit-Demo', 'Gravatar-UIKit-Demo.Release.xcconfig')
-COMMON_XCARGS = ['-skipPackagePluginValidation'] # Allow SwiftPM plugins (e.g. swiftlint) called from Xcode to be used on CI without prior manual approval
+COMMON_XCARGS = ['-skipPackagePluginValidation'].freeze # Allow SwiftPM plugins (e.g. swiftlint) called from Xcode to be used on CI without prior manual approval
 
 GITHUB_REPO = 'Automattic/Gravatar-SDK-iOS'
 GITHUB_URL = "https://github.com/#{GITHUB_REPO}".freeze
--- a/fastlane/lanes/localization.rb
+++ b/fastlane/lanes/localization.rb
@@ -64,6 +64,11 @@ platform :ios do
     SOURCES_TO_LOCALIZE.each do |source|
       next if source.gp_project_url.nil?
 
+      check_translation_progress(
+        glotpress_url: source.gp_project_url,
+        abort_on_violations: false
+      )
+
       ios_download_strings_files_from_glotpress(
         project_url: source.gp_project_url,
         locales: GLOTPRESS_TO_LPROJ_APP_LOCALE_CODES,
--- a/fastlane/lanes/localization.rb
+++ b/fastlane/lanes/localization.rb
@@ -75,6 +75,10 @@ platform :ios do
         download_dir: source.localizations_root
       )
 
+      ios_lint_localizations(
+        input_dir: source.localizations_root
+      )
+
       next if skip_commit
 
       paths_to_commit << source.localizations_root


andrewdmontgomery:
I'm targeting the `release/3.0.0` branch with this.  But we'll need it in `trunk`, too, since that's the source of truth for localizations.  So once this merges, I'll set up a back merge for that.  We'll want that anyway.

andrewdmontgomery:
@AliSoftware
I considered adding `ios_lint_localizations` to our `generate_strings` lane also, specifically because it checks for duplicate keys (and that would be the best time to check).

But since we are generating our `Localizable.strings` source file from code, it shouldn't be possible for us to have duplicate keys in the `Localizable.strings` file, right?  We could still have two places in code that attempt to use the same key, and we don't want that, of course.  But that won't ever result in duplicate keys.  It will just leave us with a key whose value and comment aren't well defined, with strings from only one of the duplicate instances.

For this reason, I opted to add linting only to the `download_localized_strings` lane.

Good idea?

_Edit: I just saw that you're at a meetup this week.  I'm reasonably confident in this approach.  So I may merge this before you have a chance to comment.  I'm more than happy to update this with any feedback in a follow-up PR.  For now, this process will be manually handled by a developer.  When we eventually automate some of this, it will probably make more sense to extract the progress check into a separate lane (or step), like [wp-ios does](https://github.com/wordpress-mobile/WordPress-iOS/blob/faf0c8b1b6930011746750e621041d6a8e7a9f7d/fastlane/lanes/localization.rb#L523)._

pinarol (approved):
LGTM 👍

AliSoftware:
> But since we are generating our Localizable.strings source file from code, it shouldn't be possible for us to have duplicate keys in the Localizable.strings file, right? 

Correct 👍 

> It will just leave us with a key whose value and comment aren't well defined, with strings from only one of the duplicate instances.

Yes. But `genstrings` can generate warnings and errors when it detects some of those cases iinm. That being said, I don't remember from the top of my head what `genstrings` considers a warning vs an error (e.g. maybe using the same key with the same value _but_ different comment is only a warning, while using the same key with different values is an error…?)
I think [the unit tests we wrote for that action might give you a good idea of the different scenarios though](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/spec/ios_generate_strings_file_from_code_spec.rb#L174-L233)

If you want to _only_ catch **errors** and have those make fail the `generate_strings` lane, all you need is `fail_on_error` (which is `true` by default iinm). But if you want to catch **warnings** _too_ (which is maybe the case we're missing in the situation you describe?), then you might want to adjust the `generate_strings` lane that calls the `ios_generate_strings_file_from_code` action, so that you store the result of that action in a variable and make your lane check if the result is empty (no error nor warning) or not (no error—otherwise `fail_on_error` would have made the action crash on purpose without returning—but still found warnings).

AliSoftware:
PS / Side note while on the topic of localization and I'm thinking about it:

Since https://github.com/wordpress-mobile/release-toolkit/pull/591 was merged and released into `release-toolkit` version `12.1.0`, if you want you can now remove your custom logic about calling `convert_generated_strings!(source: source)` (that you wrote to convert files to UTF16) and instead upgrade to the latest `release-toolkit` (we're at `12.2.0` now) to just use the new `output_encoding: 'UTF8'` option of the `ios_generate_strings_file_from_code` action now, allowing you do do some cleanup of your `Fastfile` 😉

PR #490, opened 2024-10-10
Fix the first time we upload avatar

### LINKED ISSUE #489: The first avatar upload is not observable
- remove all your avatars
- upload a new image
observe: the screen remains same
expected: the upload should show

pinarol:
Closes #489 

### Description

For some reason SwiftUI doesn't update the UI during the avatar upload if the grid is empty. Applying `@Published` property wrapper doesn't help. So I am adding `objectWillChange.send()`.

We should also refresh the page to sync with the BE and get the latest avatar selection.

### Testing Steps

Delete all your avatars
Upload an avatar
Observe that the UI updates as expected


etoledom (approved):
Thank you for the fix! Works as described 🎉

PR #492, opened 2024-10-11
Add support for setting uploaded image as avatar

### LINKED ISSUE #491: Bring back `v1/upload-image` POST
The v3 upload endpoint currently doesn't have the option to immediately set the uploaded avatar for a given email. Jetpack needs this. Let's bring back `"https://api.gravatar.com/v1/upload-image"` until we have this option. We can bring back the old signature for backward compatibility:

> public func upload(_ image: UIImage, email: Email, accessToken: String) async throws -> URLResponse

so we'll have 2 different versions of this method, one for v3, one for v1 until v3 gives us the same functionality.

For reference, v1 was removed here:
https://github.com/Automattic/Gravatar-SDK-iOS/pull/336

andrewdmontgomery:
Closes #491 

### Description

This restores support for uploading images via the `v1` API.

### DIFF
--- a/Demo/Demo/Gravatar-UIKit-Demo/DemoUploadImageViewController.swift
+++ b/Demo/Demo/Gravatar-UIKit-Demo/DemoUploadImageViewController.swift
@@ -33,6 +33,15 @@ class DemoUploadImageViewController: UIViewController {
         textField.isSecureTextEntry = true
         return textField
     }()
+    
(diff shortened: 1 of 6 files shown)


pinarol (approved):
LGTM 👍 I tested it with the UIKit demo app > Image upload.

When i enter email & token and choose v1, the uploaded image gets set as my avatar ✅ 
When i choose v3, it just uploads the image ✅

PR #494, opened 2024-10-11
Fix initialization of `@State` var

pinarol:
Closes #

### Description

This fixes a bug i detected during the Jetpack integration. Because of the way we initialize the `@State` variable, the token was not being set successfully. This article does a good job explaining why: https://www.swiftcraft.io/articles/how-to-initialize-state-inside-the-views-init- 

> From an Apple dev:
> > Although that will compile, @State variables in SwiftUI should not be initialized from data you pass down through the initializer; since the model is maintained outside of the view, there is no guarantee that the value will really be used. The correct thing to do is to set your initial state values inline: @State private var viewModel = SignInViewModel(...)

So I am separating the local `@State` token from the one passed from outside. 
Also removing the `@State` from the `scope` variable as well, it doesn't need to be a `@State`. This is not directly related with the bug, but it's the same wrong usage as well thus needs fixing.

I also rearranged the lines a bit.

### Testing Steps

Go to UIKit Demo app > QE
Enter `email`
Enter `token` <-- important

Tap Logout if the button is present

Open the QE, verify it does NOT go through OAuth and goes directly to QE.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
+++ b/Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
@@ -24,15 +24,18 @@ struct QuickEditor<ImageEditor: ImageEditorView>: View {
     fileprivate typealias Constants = QuickEditorConstants
 
     @Environment(\.oauthSession) private var oauthSession
-    @State var token: String?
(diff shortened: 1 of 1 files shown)


etoledom (approved):
Working as described 🎉

PR #495, opened 2024-10-11
Use `.gravatarBlue` for selected avatar's border

pinarol:
Closes #

### Description

Something I noticed during the Jetpack integration is that we use `.accentColor` for the selection border, and that can vary from app to app: 

We don't want that, we want to use `.gravatarBlue` according to designs.

### Testing Steps

Just check the border in the demo app and verify it's `.gravatarBlue`, it's slightly different than the default accent color.


etoledom (approved):
Good catch!

Issue #497, opened 2024-10-14
Keyboard covers the cropper


pinarol:
- Try uploading a new photo in the Quick Editor
- When the photo picker opens type something in the search bar
- Select an image

Result: Cropper opens and keyboard remains there covering the cropper
Expected: Cropper opens and keyboard is dismissed

https://github.com/user-attachments/assets/01e3ab9f-d745-4367-8e75-02cd61f67ba5

PR #498, opened 2024-10-14
Fix keyboard covering cropper

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/497

### Description

Fixes the issue described in https://github.com/Automattic/Gravatar-SDK-iOS/issues/497
Dismissing the iOS PhotoPicker's keyboard directly doesn't seem possible in SwiftUI so i added a piece of code to force the first responder to resign by sending an action to the shared application.

### Testing Steps

Follow the steps in: https://github.com/Automattic/Gravatar-SDK-iOS/issues/497

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/SystemImagePicker/SystemImagePickerView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/SystemImagePicker/SystemImagePickerView.swift
@@ -89,7 +89,9 @@ private struct ImagePicker<Label, ImageEditor: ImageEditorView>: View where Labe
     }
 
     private func pickerDidSelectImage(_ item: ImagePickerItem) {
-        UIApplication.shared.dismissKeyboard()
+        Task {
+            await UIApplication.shared.dismissKeyboard()
+        }
         imagePickerSelectedItem = item
     }
 }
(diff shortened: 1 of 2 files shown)


etoledom (approved):
Looks good, simple, and works!

I took a fast look to see if we can use SwiftUI API to achieve this same result:

On the cropper sheet, we can call the `.focused()` modifier. This will work as making the new modal the first responder, effectivelly dismissing the keyboard.

Funny enough, this seems a bit more complicated than using UIResponder, since we need to bind it to a `@FocusState private var focusState: Bool`, so it would look something like this:

```swift
var body: some View {
// body definition...
   .sheet {
      // Image Picker body
       .sheet(item: $sourceType, content: { source in
           // Cropper body
        }).focused($focusState) // This will dismiss the keyboard when the cropper is presented
    }).onAppear() {
        focusState = true // This needs to always be true, but it seems it can't be set on init.
    }
}

@FocusState private var focusState: Bool // New var. I'd prefer to pass `true` directly but I didn't find a way.
```

This "SwiftUI way" adds extra complexity to this view which I'd rather avoid. So I'm more inclined to merge the solution proposed on the PR.

pinarol:
Great insight! Thanks for sharing.

Sources/GravatarUI/SwiftUI/AvatarPicker/SystemImagePicker/SystemImagePickerView.swift
```diff
@@ -89,6 +89,9 @@ private struct ImagePicker<Label, ImageEditor: ImageEditorView>: View where Labe
     }
 
     private func pickerDidSelectImage(_ item: ImagePickerItem) {
+        Task {
+            await UIApplication.shared.dismissKeyboard()
+        }
```
etoledom, line 94: I didn't see this commit when I approved 🤔 
This line is giving a warning 
CleanShot 2024-10-14 at 16 37 34@2x

pinarol: Yes i think that happens in XCode 16, but in XCode 15.4 we need to await it otherwise there's a compilation error. We should fix the warning once we officially switch to 16.

PR #499, opened 2024-10-14
Show close button when external token is expired

etoledom:
Closes #

### Description

This PR fixes an issue where, when an external token is expired, the SDK would try to re-authenticate the user.
Now it will prompt the user to close the QE instead. 

https://github.com/user-attachments/assets/72fd7607-8220-4941-be5d-400e3b045bd8

### Testing Steps

- Open UIKit demo app.
- Go to Quick Editor screen.
- Tap on Log out button if it's present.
- Introduce an expired (or wrong) token.
- Open the Quick Editor.
  - Check that the action button says `Close` .
  - Check that the Quick Editor is closed after taping on Close button.


pinarol (approved):
Looks good 🎉

PR #500, opened 2024-10-14
Remove modifiers that directly create `AvatarPickerView`

pinarol:
Closes #

### Description

Removing the modifiers that directly create `AvatarPickerView` instead of `QuickEditor`.

### Testing Steps

Test SwiftUI demo app > Avatar picker view


etoledom (approved):
Looks good! 
Thanks for the clean up 🙏

PR #502, opened 2024-10-14
Remove unused client secret

etoledom:
Closes #

### Description

This PR removes the request of Client Secret on the secrets configuration, which is not needed as we use the `token` response OAuth configuration.

### Testing Steps
- Test the OAuth of the Quick Editor in both SwiftUI and UIKit demos.


pinarol (approved):
Looks good 👍

PR #510, opened 2024-10-16
Fix concurrency warning about editingDidFinish

pinarol:
Closes #

### Description

Screenshot 2024-10-16 at 11 26 53

Swift implicitly assumes that editingDidFinish is an actor-isolated property(due to its use in a View protocol). To address this, we can explicitly mark the property as `@MainActor`, or alternatively, we can make editingDidFinish a `@Sendable` closure. I did the second which means we don't assume it must run on the main actor. 

And then i got a new warning in our ImagePicker:

Screenshot 2024-10-16 at 12 01 19

Even though ImagePicker is a View, the system this time doesn't automatically assume it runs on the main actor. So I added the `@MainActor` explicitly.

And then i got:

Screenshot 2024-10-16 at 12 03 19

Then i put the call inside `Task { ... }`

### Testing Steps

Test image picking & cropping.

### DIFF
--- a/Demo/Demo/Gravatar-SwiftUI-Demo/TestImageCropper.swift
+++ b/Demo/Demo/Gravatar-SwiftUI-Demo/TestImageCropper.swift
@@ -3,9 +3,9 @@ import GravatarUI
 
 struct TestImageCropper: View, ImageEditorView {
     var inputImage: UIImage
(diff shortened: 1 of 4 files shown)


pinarol:
Closing this in favor of https://github.com/Automattic/Gravatar-SDK-iOS/pull/512

PR #511, opened 2024-10-16
Fix concurrency warning for avatarUpdatedHandler

pinarol:
Closes #

### Description

Screenshot 2024-10-16 at 12 15 19

For consistency with the onDismiss handler i solved this by explicitly running the closure on the main actor. 

### Testing Steps

Observe that avatar selection is reflected on the background page in the UIKit demo app.


pinarol:
Closing this since the warning seems like a false positive. It's not there with XCode 16.

PR #514, opened 2024-10-17
Update secrets removing client secret

etoledom:
Closes #

### Description

This PR removes the not-needed `clientSecret` field from the Secrets struct.
Both the local template and the internal secrets repo has been updated.

### Testing Steps

- Install a CI build
- Check that the Quick Editor authentication works as expected.


pinarol (approved):
LGTM 👍

PR #515, opened 2024-10-17
Upload heic image format

etoledom:
Closes #

### Description

Testing uploading HEIC instead of JPEG image format.

- The image size is constantly slightly better while the quality seems unchanged.
- The processing time, in all my tests, where either almost equal (slightly better), and some times considerably better.

In no case I saw neither bigger image size nor increased processing time.
The smaller processing time might be related to the smaller image size.

### Testing Steps

- Upload many images checking both image size and overall request time.


pinarol (approved):
Looks good 🎉

andrewdmontgomery (approved):
Nice

PR #524, opened 2024-10-22
Handle https callback url

etoledom:
Closes #

### Description

This PR is aimed to solve the problem where a oauth callback used has a https scheme.
Since WP.com OAuth only supports `https`, custom schemes can not be used.

### Testing Steps

Pocket Casts PR for testing: https://github.com/Automattic/pocket-casts-ios/pull/2267
Tests on both SDK and PocketCasts:

- Cancel the OAuth flow before completing it.
  - See the `Login required` error screen.
- Full OAuth flow:
  - Open the quick editor and complete the OAuth flow successfully
- OAuth errors:
  - [Revoque](https://wordpress.com/me/security/connected-applications) the access token (it should appear as Pocket Casts for the Pocket Casts app).
  - Open the Quick Editor after it being logged in successfully.
    -  See the `Token expired` error.
  - Log in with a different email.
  - Open the Quick Editor. It should show the previous email on the web view.
  - Accept without changing the email.
    - See the `wrong email` error.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
@@ -82,14 +82,20 @@ public struct OAuthSession: Sendable {
             let newToken = KeychainToken(token: tokenText)
             shared.overrideToken(newToken, for: email)
             await shared.authenticationSession.cancel()
-            NotificationCenter.default.post(name: .authorizationFinished, object: nil)
+            postNotification(.authorizationFinished)
             return true
         } catch OAuthError.couldNotParseAccessCode(email.rawValue) {
             return false // The URL was not a Gravatar callback URL with a token.
         } catch {
-            NotificationCenter.default.post(name: .authorizationError, object: error)
             await shared.authenticationSession.cancel()
+            postNotification(.authorizationError, error: error)
             return true
+        }
+    }
+
+    private static func postNotification(_ name: Notification.Name, error: Error? = nil) {
+        Task { @MainActor in
+            NotificationCenter.default.post(name: name, object: error)
         }
     }
 
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
@@ -1,7 +1,8 @@
 import AuthenticationServices
 
 public struct OAuthSession: Sendable {
-    private static let shared = OAuthSession()
+    static let shared = OAuthSession()
+    private var emailStorage = SessionEmailStorage()
 
     private let storage: SecureStorage
     private let authenticationSession: AuthenticationSession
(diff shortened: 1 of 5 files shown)


Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
```diff
@@ -141,9 +153,16 @@ struct QuickEditor<ImageEditor: ImageEditorView>: View {
...
-            fetchedToken = oauthSession.sessionToken(with: email)?.token
-            isAuthenticating = false
+            onAuthenticationFinished()
+        }
+    }
+
+    func onAuthenticationFinished() {
+        if let fetchedToken = oauthSession.sessionToken(with: email)?.token {
+            self.fetchedToken = fetchedToken
+            oauthError = nil
```
andrewdmontgomery, line 159: When`oauthSession.hasValidSession(with: email)` returns `false` (in `performAuthentication()`), then `oauthSession.sessionToken(with: email)?.token` will return `nil`, and vice versa.

❓ Since we're already setting `oauthEerrorl`  in `performAuthentication()` in cases when `oauthSession.hasValidSesions(with: email)` is `false`, does it makes sense to move this `oauthError = nil`  there also when it's true?

etoledom: I think the problem here, is that when the `https` flow is triggered (which will potentially be all the cases on production), the `performAuthentication()` function won't be canceled (to close the web view), and the flow is finished through the notifications, which will call `onAuthenticationFinished()`, also in the case of success. And in this case we might need `oauthError = nil` inside `onAuthenticationFinished()`.

Maybe what you described will handle all cases properly, but I'd be extra careful testing every success and error case (also a success after an error screen), if there's any change here.

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -54,25 +55,41 @@ public struct OAuthSession: Sendable {
...
+            shared.overrideToken(newToken, for: email)
+            await shared.authenticationSession.cancel()
+            NotificationCenter.default.post(name: .authorizationFinished, object: nil)
+            return true
+        } catch OAuthError.couldNotParseAccessCode(email.rawValue) {
+            return false // The URL was not a Gravatar callback URL with a token.
         } catch {
-            throw OAuthError.from(error: error)
+            NotificationCenter.default.post(name: .authorizationError, object: error)
+            await shared.authenticationSession.cancel()
```
andrewdmontgomery: ❓ Is it ok to send the notification before calling `await ....cancel()`

Could it be problematic for a recipient to receive that notification before `await shared.authenticationSession.cancel()` returns?

andrewdmontgomery (approved):
It works as described now!  Looks good.  I left a couple of questions.  But I wouldn't let them hold this up.

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
etoledom: Not sure if it could be problematic, but it does seem safer to make this change 👍

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -54,25 +55,47 @@ public struct OAuthSession: Sendable {
...
+        do {
+            let tokenText = try shared.tokenResponse(from: callbackURL).token
             guard try await CheckTokenAuthorizationService().isToken(tokenText, authorizedFor: email) else {
                 throw OAuthError.loggedInWithWrongEmail(email: email.rawValue)
             }
             let newToken = KeychainToken(token: tokenText)
-            overrideToken(newToken, for: email)
-            return newToken
+            shared.overrideToken(newToken, for: email)
+            await shared.authenticationSession.cancel()
```
pinarol, line 84: Question: Just wondering why the manual cancel() is necessary? Doesn't the oauth session end itself once it's successful?

etoledom: When the callback URL comes from universal links, the oauth session never returns, so it's never a "success", and the OAuth web view won't dismiss.

As every real scenario will use `https` and universal link, it will always need this manual cancelling, so I think it safer to call it always, though is not necessary for our particular case of the demo app with the custom scheme exclusion. 

Testing the demo app, it didn't seem to be a problem, so I decided to keep the logic simple.

etoledom: Without the canceling, the web view stays like this without dismissing:


So the canceling is basically the simplest way I found to dismiss it.

pinarol: I see, interesting. Thanks!

pinarol (approved):
LGTM 👍

PR #527, opened 2024-10-24
Update image upload request with new options

etoledom:
Closes #

### Description

This PR updates the image upload request with the new image selection options.

- The Quick Editor will automatically select an image when there's no image previously selected.
- The avatar selected callback will be called now also when a selection happens because of an image upload.
- openapi spec updated 👍 

#### Image selection options
- Run the UIKit app
- Go to Image Upload demo screen.
- Add email, token, and choose the 3 different options for image upload.
- Check on web that the image selection hapens as desired.

#### Image selected callback
- Run the UIKit app
- Go to Quick Editor.
- Open the Quick Editor on  Horizontal layout (to see the avatar changes)
  - Select an avatar
    -  See that the avatar on the background changes acordingly
  - Upload a new image.
    - Check that no selection happens. 
  - On web, delete the selected avatar to not have any selected
  - On the demo app, upload a new image
    - Check that the newly uploaded image gets selected on the QE.
    - Check that the avatar on the background changes acordingly.

### DIFF
--- a/Sources/Gravatar/Network/Services/AvatarService.swift
+++ b/Sources/Gravatar/Network/Services/AvatarService.swift
@@ -40,19 +40,6 @@ public struct AvatarService: Sendable {
         return try await imageDownloader.fetchImage(with: gravatarURL, forceRefresh: options.forceRefresh, processingMethod: options.processingMethod)
     }
 
-    /// Uploads an image and sets it as the avatar of the given email at Gravatar.com. Returns the `URLResponse` of the network tasks asynchronously. Throws
-    /// ``ImageUploadError``.
-    /// - Parameters:
-    ///   - image: The image to be uploaded.
-    ///   - email: An`Email` object
-    ///   - accessToken: The authentication token for the user. This is a WordPress.com OAuth2 access token.
-    /// - Returns: An asynchronously-delivered `URLResponse` instance, containing the response of the upload network task.
-    @discardableResult
-    public func upload(_ image: UIImage, email: Email, accessToken: String) async throws -> URLResponse {
-        try await imageUploader.uploadImage(image, accessToken: accessToken, avatarSelection: .selectUploadedImage(for: email), additionalHTTPHeaders: nil)
-            .response
-    }
-
     /// Uploads an image to be used as the user's Gravatar profile image, and returns the `URLResponse` of the network tasks asynchronously. Throws
     /// ``ImageUploadError``.
     /// - Parameters:
@@ -60,8 +47,8 @@ public struct AvatarService: Sendable {
     ///   - accessToken: The authentication token for the user. This is a WordPress.com OAuth2 access token.
     /// - Returns: An asynchronously-delivered `AvatarType` instance, containing data of the newly created avatar.
     @discardableResult
-    public func upload(_ image: UIImage, accessToken: String) async throws -> AvatarType {
-        let avatar: Avatar = try await upload(image, accessToken: accessToken, avatarSelection: .preserveSelection)
+    public func upload(_ image: UIImage, selectionBehavior: AvatarSelection, accessToken: String) async throws -> AvatarType {
+        let avatar: Avatar = try await upload(image, accessToken: accessToken, avatarSelection: selectionBehavior)
         return avatar
     }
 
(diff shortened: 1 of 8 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -138,6 +138,11 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
         .onChange(of: authToken ?? "") { newValue in
             model.update(authToken: newValue)
         }
+        .onChange(of: model.selectedAvatarURL) { _ in
+            if !model.isFirstLoad {
```
pinarol: my suggestion to replace the opt-out mechanism with an opt-in one here:

https://github.com/Automattic/Gravatar-SDK-iOS/pull/528

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -40,28 +40,15 @@ public struct AvatarService: Sendable {
...
     /// Uploads an image to be used as the user's Gravatar profile image, and returns the `URLResponse` of the network tasks asynchronously. Throws
     /// ``ImageUploadError``.
     /// - Parameters:
     ///   - image: The image to be uploaded.
     ///   - accessToken: The authentication token for the user. This is a WordPress.com OAuth2 access token.
     /// - Returns: An asynchronously-delivered `AvatarType` instance, containing data of the newly created avatar.
     @discardableResult
-    public func upload(_ image: UIImage, accessToken: String) async throws -> AvatarType {
-        let avatar: Avatar = try await upload(image, accessToken: accessToken, avatarSelection: .preserveSelection)
+    public func upload(_ image: UIImage, selectionBehavior: AvatarSelection, accessToken: String) async throws -> AvatarType {
```
pinarol, line 51: The docs need updating for the new param `selectionBehavior`.

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -40,28 +40,15 @@ public struct AvatarService: Sendable {
...
+        let avatar: Avatar = try await upload(image, accessToken: accessToken, selectionBehavior: selectionBehavior)
```
pinarol, line 52: seems like we call a method that has the same parameters, so we don't need 2 different methods anymore?

etoledom: This is to be able to keep the `Avatar` type internal and return a `AvatarType` protocol instead. Notice that the return types are different, and we need `Avatar` return internally.

pinarol (approved):
Nice job 👍

Sources/Gravatar/Network/Services/AvatarService.swift
pinarol: I see, gotcha 👍

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
etoledom: Thank you for the suggestion!
Merged into this branch 👍

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -26,6 +26,7 @@ class AvatarPickerViewModel: ObservableObject {
     }
 
     @Published var selectedAvatarURL: URL?
+    @Published var backendSelectedAvatarURL: URL?
```
andrewdmontgomery: Can we make this read only outside the class?  It compiles.  And I think this is the only class that would set it.

```suggestion
    @Published private(set) var backendSelectedAvatarURL: URL?
```

Sources/Gravatar/Network/Services/ImageUploadService.swift
```diff
@@ -56,50 +55,8 @@ struct ImageUploadService: ImageUploader {
...
-        body.append("--\(boundary)\r\n")
-        body.append("Content-Disposition: form-data; name=\"\(UploadParameters.accountKey)\"\r\n\r\n")
-        body.append("\(email)\r\n")
-    }
-    // EOF!
-    body.append("--\(boundary)--\r\n")
-
-    return body
+private func imageUploadBody(with imageData: Data, boundary: String) -> Data {
+    imageUploadBodyV3(with: imageData, boundary: boundary)
```
andrewdmontgomery: Now that we don't need `imageUploadBodyV1(::)`, we could move the logic in `imageUploadBodyV3(::)` back into the main `imageUploadBody::)`.

andrewdmontgomery (approved):
Testing worked as described.  Just one quick question.  And no need to hold this up.

Sources/Gravatar/Network/Services/ImageUploadService.swift
andrewdmontgomery: This is a tiny change, so I just pushed it.

Issue #533, opened 2024-10-25
Add a CI check to force adding labels to PRs


pinarol:
There is [labels_checker](https://github.com/Automattic/dangermattic/blob/trunk/lib/dangermattic/plugins/labels_checker.rb) as a part of Dangermattic. PocketCasts use it [here](https://github.com/Automattic/pocket-casts-android/blob/f51e9cf382e1e9a43bf89d93d61a62b9e8861279/Dangerfile#L29-L33). 

We have an apps-infra request to setup danger configurations first. `appsinfrarequests -> /danger-configurations-for-gravatar-sdk-ios`

PR #534, opened 2024-10-25
[Tooling] Setting up Danger

iangmaia:
### Description

PR setting up [Danger](https://danger.systems) and [Dangermattic](https://github.com/Automattic/dangermattic) checks in the repo.
See pdnsEh-1TU-p2.

### Testing Steps

- Make sure there's a Danger status check posted in the GitHub build
- Make sure that, if applicable, a message from the bot Dangermattic bot is posted in the PR
- Make sure changes in the PR (e.g. removing, adding labels) trigger Danger to run again


Dangerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# `files: []` forces rubocop to scan all files, not just the ones modified in the PR
+# `skip_bundle_exec` prevents RuboCop from running using `bundle exec`, which we don't want on the linter agent
+rubocop.lint(files: [], force_exclusion: true, inline_comment: true, fail_on_inline_comment: true,
+             include_cop_names: true, skip_bundle_exec: true)
+
+manifest_pr_checker.check_gemfile_lock_updated
+
+labels_checker.check(
+  do_not_merge_labels: ['do not merge'],
```
dangermattic, line 13: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/Gravatar-SDK-iOS/blob/2f3a39bafeaf2a0da25bb6e160afdb9e8037000b/Dangerfile#L13"></span>Style/TrailingCommaInArguments: Avoid comma after the last parameter of a method call.</td>
    </tr>
  </tbody>
</table>

Dangerfile
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+github.dismiss_out_of_range_messages
+
+# `files: []` forces rubocop to scan all files, not just the ones modified in the PR
+# `skip_bundle_exec` prevents RuboCop from running using `bundle exec`, which we don't want on the linter agent
```
AliSoftware, line 6: Unrelated to this PR, but now that the new standard is to use the linter agent for all repos that are using `Dangermattic`, I wonder if `skip_bundle_exec: true` shouldn't be the new default for this Dangermattic action 🤔

Dangerfile
```diff
@@ -0,0 +1,14 @@
...
+rubocop.lint(files: [], force_exclusion: true, inline_comment: true, fail_on_inline_comment: true,
+             include_cop_names: true, skip_bundle_exec: true)
+
+manifest_pr_checker.check_gemfile_lock_updated
+
+labels_checker.check(
+  do_not_merge_labels: ['do not merge']
+)
```
AliSoftware: The initial appsinfrarequest (pdnsEh-1TU-p2) was about checking at least one label was set on PRs, so I think we also want to enforce that?
```suggestion
labels_checker.check(
  do_not_merge_labels: ['do not merge'],
  required_labels: [//], # At least one label, regardless of its name
  required_labels_error: "You need to add at least one label to this PR",
)
```

I indeed checked by removing the `CI/CD` and `tooling` labels that were assigned to this PR, and Danger re-ran as expected… but indeed didn't report any error (given the default for `required_labels:` in our `labels_checker` plugin is `[]`, aka, no label required in particular… so that checks out)

---

Though cc @pinarol to provide any additional info in case there were some specific set of labels you wanted to enforce… For example, I don't know if in the near future you plan to name your labels with category prefixes (like some other repos do), e.g. `type: tooling`/`type: feature`/`type: bugfix`/`type: refactor`/`area: concurrency`/`area: ui component`/`area: tech debt`/`area: tooling`/… and then we could adjust this condition to enforce "at least one `type:*` label and at least one `area: *` label", or things like that.

But looking at the **current** labels available on the repo, at least for an initial setup I think "at least one label, whatever that label is" (`[//]`) is a good start and we can iterate on it later.

Dangerfile
iangmaia: This call is actually from `danger-rubocop` itself, but I agree, and this is something I remember thinking / discussing with someone else a while ago. `files: []` can also be a default. Perhaps we can add in Dangermattic a helper or a patch to the Rubocop plugin adding these defaults.

Dangerfile
andrewdmontgomery: > But looking at the current labels available on the repo, at least for an initial setup I think "at least one label, whatever that label is" ([//]) is a good start and we can iterate on it later.

This is still being discussed in an RFC.  But for what it's worth, that's where I think we should start.

The intent of this change is to track metrics related to features and bugs.  But we'll obviously have PR's that don't address features and bugs.  If we enforce specific labels, we'll have to include labels that are outside the scope of this RFC.  Rather than expand the scope, I have suggested that we start simple, as you've described, and see if that's sufficient.

pinarol: Yes, we can start simple 👍

pinarol (approved):
Thank you! 
:shipit:

Dangerfile
```diff
@@ -0,0 +1,16 @@
...
+# `skip_bundle_exec` prevents RuboCop from running using `bundle exec`, which we don't want on the linter agent
+rubocop.lint(files: [], force_exclusion: true, inline_comment: true, fail_on_inline_comment: true,
+             include_cop_names: true, skip_bundle_exec: true)
+
+manifest_pr_checker.check_gemfile_lock_updated
+
+labels_checker.check(
+  do_not_merge_labels: ['do not merge'],
+  required_labels: [//], # At least one label, regardless of its name
+  required_labels_error: "You need to add at least one label to this PR",
```
dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/Gravatar-SDK-iOS/blob/4559cdeb0d9e4c39a43cfd41e78fda25ffc4d865/Dangerfile#L15"></span>Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.</td>
    </tr>
  </tbody>
</table>

dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/Gravatar-SDK-iOS/blob/4559cdeb0d9e4c39a43cfd41e78fda25ffc4d865/Dangerfile#L15"></span>Style/TrailingCommaInArguments: Avoid comma after the last parameter of a method call.</td>
    </tr>
  </tbody>
</table>

Dangerfile
iangmaia: Thanks for the input @andrewdmontgomery & @pinarol! I've added the "at least one label" check @AliSoftware suggested and we can use it as a starting point 🙇

pinarol:
We can close https://github.com/Automattic/Gravatar-SDK-iOS/issues/533 after merging this.

PR #414, opened 2024-09-18
Xcode 16.0

andrewdmontgomery:
Closes #413

### Description

This updates the repo to use Xcode 16.0 for development and CI.  It includes an update to SwiftFormat, which addresses an issue related to Xcode 16.0

### Testing Steps

- CI should be 🟢 
- Smoke test development
   - Build both demo project targets
   - Build the SDK via the Swift Package
- `make test`
- `make generate`

PR #549, opened 2024-10-28
Update to Xcode 16

### LINKED ISSUE #413: Xcode 16.0
Update to Xcode 16.0 for development and CI

etoledom:
Closes #413 

### Description

This PR takes the work from https://github.com/Automattic/Gravatar-SDK-iOS/pull/414, trying to fix the issue with circular corner radius on snapshot tests.

The way to fix this I found is to add this line on the Avatar view layer:
`avatarImageView.layer.cornerCurve = .circular`
And it seems to work.

There are also small differences on how things are rendered, so mostly all snapshot need re-drawing (mostly fonts). 
It's important to check that the differences are not noticeable by the eye.

### Testing Steps

- CI: 🍏 
- Run GravatarUI unit tests locally (on iPhone SE)


Sources/GravatarUI/ProfileView/Avatar/DefaultAvatarProvider.swift
```diff
@@ -65,6 +65,7 @@ class DefaultAvatarProvider: AvatarProviding {
     private func applyCornerRadius() {
         guard !skipStyling else { return }
         avatarImageView.layer.cornerRadius = avatarCornerRadius
+        avatarImageView.layer.cornerCurve = .circular
```
pinarol, line 68: Snapshots look nice as far as I can tell.

pinarol (approved):
LGTM 👍 
(I ran snapshot tests in my local with iPhone SE 3rd gen. iOS 18.0)

.buildkite/shared-pipeline-vars
```diff
@@ -3,6 +3,6 @@
 # This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
 # to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
 
-export IMAGE_ID=$(echo "xcode-$(cat .xcode-version)")
+export IMAGE_ID="xcode-16.0-v7"
```
andrewdmontgomery, line 6: Yep, nice.

Tests/GravatarUITests/AboutMeBuilderTests.swift
```diff
@@ -7,9 +7,10 @@ final class AboutMeBuilderTests: XCTestCase {
     let frameSmall = CGRect(x: 0, y: 0, width: 100, height: 200)
     let palettesToTest: [PaletteType] = [.light, .dark]
 
-    override func setUp() async throws {
-        try await super.setUp()
-        // isRecording = true
+    override func invokeTest() {
+        withSnapshotTesting(record: .failed) {
+            super.invokeTest()
+        }
```
andrewdmontgomery, line 13: I didn't know about this

andrewdmontgomery (approved):
Excellent 🚀

Tests/GravatarUITests/AboutMeBuilderTests.swift
etoledom: This is new after updating the package. 

If the snapshot comparison fails, there will be a new recording right away, but the test will also fail. 

I think this is good to see right away what's de difference, and we can choose to save it (just run the test again it will succeed), or choose to make changes if the diff is unwanted.

etoledom: cc @pinarol

PR #551, opened 2024-10-28
Clean up SwiftLint removal

### LINKED ISSUE #550: Clean up after SwiftLint removal

andrewdmontgomery:
Closes #550 

### Description

This finishes removing SwiftLintPlugin by removing the unused dependency declaration.

It also apparently cleans up some inconsistencies with the Package.resolved files.

### Testing Steps

- [ ] CI should be 🟢


etoledom (approved):
Looking good!

Question:
Should we also remove the custom SwiftLint rule which checks the `SDKLocalizedString` usage? Or are we looking to use it in another way?

andrewdmontgomery:
> Question:
> Should we also remove the custom SwiftLint rule which checks the `SDKLocalizedString` usage? Or are we looking to use it in another way?

I'm looking for another way to use it.  But good catch.  I'll pull it for now.

PR #552, opened 2024-10-29
Support for oauth callback https for iOS 17.4 +

### LINKED ISSUE #554: Support for OAuth Callback https for iOS 17.4 +
Apple released a new API to handle the `https` case which is iOS 17.4+

We can implement this on our SDK

https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/callback

etoledom:
Closes #554 

### Description

This PR implements the new `ASWebAuthenticationSession` API for callback URLs with `https` schemes.

For users, it's a non-breaking change, and if they support only iOS 17.4+, they won't need to implement the app/scene delegate auth callback.

I also made some changes, introducing `SessionStorage` to silence some concurrency warnings (Swift 6 errors). After trying many different paths, this one the only one that worked with localized changes (not spreading to other classes).

If you want to check the warning and experiment with other solutions, roll back the last 3 commits from this branch

### Testing Steps

- Log out from the Quick Editor.
- Open the Quick Editor authorisation.
- Check that `Approbe` and `Deny` buttons on the web view work.
  - You can also check that `_ = await OAuthSession.handleCallback(url)` in `SceneDelegate.swift:55` is not called.
- To test pre iOS17.4, comment the `if #available(iOS 17.4, *)` path and run again.
  - Check that the Auth web view works as expected
  - Check that `_ = await OAuthSession.handleCallback(url)` in `SceneDelegate.swift:55` is called.


etoledom:
I will continue working on this when we have Universal Links setup for the demo apps

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

etoledom:
> I will continue working on this when we have Universal Links setup for the demo apps

Ready for review

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
```diff
@@ -1,36 +1,74 @@
...
+                } else {
+                    .customScheme(callbackURLComponents.scheme ?? "")
                 }
+
+                session = ASWebAuthenticationSession(url: url, callback: callback) { callbackURL, error in
+                    if let error {
+                        continuation.resume(throwing: error)
+                    } else if let callbackURL {
+                        continuation.resume(returning: callbackURL)
+                    }
```
andrewdmontgomery: 💡 Small idea.  I'll leave it to you.

Would it improve readability to declare this completion handler once and then call it from both halves of the `if #available(iOS 1.7.4, *)`?

```swift
let completionHandler: ASWebAuthenticationSession.CompletionHandler = { callbackURL, error in
    if let error {
        continuation.resume(throwing: error)
    } else if let callbackURL {
        continuation.resume(returning: callbackURL)
    }
}
```

And then...

```swift
session = ASWebAuthenticationSession(
    url: url,
    callback: callback,
    completionHandler: completionHandler
)
```

and...

```swift
session = ASWebAuthenticationSession(
    url: url,
    callbackURLScheme: callbackURLComponents.scheme,
    completionHandler: completionHandler
)
```

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
```diff
@@ -1,36 +1,74 @@
...
-            await session?.cancel()
+            guard let session = await sessionStorage.restore() else { return }
+            session.cancel()
         }
     }
 }
+
+// `ASWebAuthenticationSession` is not thread safe. `SessionStorage` helps to silence some warnings (Swift 6 errors),
+// but we are still importing `AuthenticationServices` as `@preconcurrency`.
+// In the other hand, there won't be more than one attempt of oauth at a time, which reduces possible concurrency issues.
```
andrewdmontgomery:
```suggestion
//On the other hand, there won't be more than one attempt of oauth at a time, which reduces possible concurrency issues.
```

andrewdmontgomery (approved):
Works as advertised

Sources/Gravatar/Configuration.swift
```diff
@@ -6,7 +6,11 @@ public actor Configuration {
         package let clientID: String
         package let redirectURI: String
         package var callbackScheme: String {
-            URLComponents(string: redirectURI)?.scheme ?? ""
+            callbackURLComponents?.scheme ?? ""
+        }
+
+        public var callbackURLComponents: URLComponents? {
```
pinarol: This doesn't need to be public I think?

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
```diff
@@ -1,36 +1,74 @@
...
+            Task {
+                await sessionStorage.save(session)
             }
 
             Task { @MainActor in
-                await session?.presentationContextProvider = context
-                await session?.start()
+                guard let session = await sessionStorage.restore() else { return }
+                session.presentationContextProvider = self
+                session.start()
```
pinarol: Why do we have 2 parallel unstructured tasks for something that looks like needs to run in a certain order? `save` must always be happening before `restore`, no? 

There's no guaranteed order between these 2 Tasks, so why not we just put them in the same Task?

```
            Task { @MainActor in
                await sessionStorage.save(session)
                guard let session = await sessionStorage.restore() else { return }
                session.presentationContextProvider = self
                session.start()
            }
```

Actually, this way we don't need the restore part either I think?

```
            Task { @MainActor in
                await sessionStorage.save(session)
                session.presentationContextProvider = self
                session.start()
            }
```

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
```diff
@@ -1,36 +1,74 @@
...
+                        continuation.resume(returning: callbackURL)
+                    }
+                }
+            }
+
+            Task {
+                await sessionStorage.save(session)
             }
 
             Task { @MainActor in
```
pinarol, line 32: Do we really need the `@MainActor` tag here?

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
```diff
@@ -1,36 +1,74 @@
...
+                guard let session = await sessionStorage.restore() else { return }
+                session.presentationContextProvider = self
+                session.start()
             }
         }
     }
 
-    nonisolated
     func cancel() {
         Task { @MainActor in
```
pinarol, line 41: Same, I am wondering why `@MainActor` tag was necessary. Just curious.

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
pinarol: Okay I tried, the `session.start()` call needs it. Nvm.

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
pinarol: Here it doesn't crash if we remove it but let's keep it just in case. I can't be sure what happens in different iOS versions, seems like `ASWebAuthenticationSession` calls might need it.

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
etoledom: Yes. `session.start()` must run on the main thread.

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
etoledom: Actually, on this one I'm not seeing the crash from running on a background thread 🤔 

I see it when `session.starts()` runs in background but not here.
I'll remove `@MainActor` from this one 👍

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
etoledom: Oh this is a good one! 
This was most probably a left over from the many things I tried to solve the compile errors.
We can totally mix them in the same call. It has to be `@MainActor` though.

Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
etoledom: > Here it doesn't crash if we remove it but let's keep it just in case. I can't be sure what happens in different iOS versions, seems like ASWebAuthenticationSession calls might need it.

Just read this comment. Yeah I agree, let's keep it. I was surprised by not seeing the crash myself.

etoledom:
Updated with the review comments 👍

pinarol (approved):
Looks good 👍 🎉 Thanks for taking care of this.

PR #555, opened 2024-10-30
Adding script to modify access control of generated openapi types

### LINKED ISSUE #537: Generate some OpenAPI models as internal
We can maintain a list of public model names in a file. After generating the .yaml we can edit it with a command line script and add a "public: true" property to those models in the yaml. The mustache files can start to respect that flag when putting the "public" modifier.

etoledom:
Closes #537 

### Description

This is a proposal to select the access control of generated openapi types independent of each other.

This is a swift script which will run before SwiftLint. It's a kind of brute-force approach, matching strings on the file to be replaced, but it's simple and works.

There's a big room for improvements, like:
- Passing the types to modify via arguments defined in a separate file.
- Making the script an executable target inside our package.
- Improvements to the algorithm itself.

### Testing Steps

- Run `make generate`
- No changes to the generated files should be present.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

pinarol (approved):
This should give the desired outcome so looks good to me 👍.

I was thinking about pre-processing the yaml to mark the models as public/internal and making the mustache files respect that. But I don't have any intentions to over engineer this. So 👍

etoledom:
> This should give the desired outcome so looks good to me 👍.
> 
> I was thinking about pre-processing the yaml to mark the models as public/internal and making the mustache files respect that. But I don't have any intentions to over engineer this. So 👍

Thank you @pinarol !

I've made the PR ready to review. @andrewdmontgomery - Anything to add?

I think it would be good to merge this before we need to update the spec again. If it's good enough now, we can think of further improvements later on if/when they are needed.

andrewdmontgomery:
I like that this is written in Swift.

Yeah, this is a good start.  Let's merge it.

PR #560, opened 2024-11-13
Refactor to use enum with associated values

andrewdmontgomery:
Closes #

### Description

This refactors #419 to use a `SquaringStrategy` enum with associated values instead of a raw `CGFloat` "threshold" value.  It supports four different squaring strategies:

- `crop`: Use one of four cropping behavior to create a square image:
   - `aspectFill`
   - `aspectFit`
   - `threshold(CGFloat)`: Uses a threshold to decide whether to fill or fit
   - `default`: uses our default threshold to decide whether to fill or fit
- `custom`: Use a custom squaring strategy to create a square image
- `default`: Use the "default" cropping behavior to create a square image (which uses our default cropping threshold)
- `none`: do not attempt to square the image

The "custom" option may be overkill.  It would allow a third party to supply their own squaring logic.  I included just because it requires no extra effort to make it available.

The one reason I can think of where this "custom" option might be useful is if a developer wants to use a different default background color when filling in a square image.

### DIFF
--- a/Sources/Gravatar/UIImage/SquaringStrategy.swift
+++ b/Sources/Gravatar/UIImage/SquaringStrategy.swift
@@ -1,0 +1,49 @@
+import UIKit
+
+public enum SquaringStrategy {
+    case crop(behavior: CroppingBehavior)
+    case custom(strategy: ImageSquarer)
+    case `default`
+    case none
+
+    package var square: (UIImage) -> UIImage {
+        let squarer: ImageSquarer = switch self {
+        case .crop(let behavior):
+            GravatarImageSquarer(threshold: behavior.threshold)
+        case .custom(let strategy):
+            strategy
+        case .default:
+            GravatarImageSquarer(threshold: .defaultThreshold)
+        case .none:
+            NoSquaring()
+        }
+
+        return squarer.square(_:)
+    }
+}
+
+public enum CroppingBehavior {
+    case aspectFill
+    case aspectFit
+    case threshold(CGFloat)
+    case `default`
+
+    var threshold: CGFloat {
+        switch self {
+        case .aspectFill:
+            .aspectFillThreshold
+        case .aspectFit:
+            .aspectFitThreshold
+        case .threshold(let value):
+            value
+        case .default:
+            .defaultThreshold
+        }
+    }
+}
+
+extension CGFloat {
+    public static let defaultThreshold: CGFloat = 0.02
+    public static let aspectFitThreshold: CGFloat = 0.0
+    public static let aspectFillThreshold: CGFloat = 1.0
+}
--- a/Sources/Gravatar/UIImage/SquaringStrategy.swift
+++ b/Sources/Gravatar/UIImage/SquaringStrategy.swift
@@ -1,22 +1,35 @@
 import UIKit
 
+/// A strategy for handline images that are not square.
+///
+/// ## Squareness
+/// `Squareness` is similar to `Aspect Ratio`, except that all values are in the range `0...1`
+/// - A square UIImage (`100 x 100`) has a `squareness` of `1`
+/// - A UIImage with a `size` of `100 x 200` has a `squareness of `0.5`
 public enum SquaringStrategy {
+    /// Sqares the image using an `aspectFill` strategy
     case aspectFill
+    /// Sqares the image using an `aspectFit` strategy
     case aspectFit
-    case edgeRatioDeterminesFitOrFill(edgeRatio: CGFloat)
+    /// Determines the squaring strategy based on how close the image is to square, by comparing the `shortEdge` and `longEdge` of the image.
+    ///
+    /// `Squareness` is similar to `Aspect Ratio`, except that all values are in the range `0...1`
+    /// - A square `UIImage` (`100 x 100`) has a `squareness` of `1`
+    /// - A rectangular `UIImage` with a `size` of `100 x 200` (or `200 x 100`) has a squareness of `0.5`
+    case squarenessDeterminesFitOrFill(squarenessThreshold: CGFloat)
     case `default`
     case none
 
     package func square(_ image: UIImage) -> UIImage {
         switch self {
         case .aspectFit:
-            ImageSquarer(edgeRatioThreshold: .aspectFitThreshold).square(image)
+            ImageSquarer(squarenessThreshold: .aspectFitThreshold).square(image)
         case .aspectFill:
-            ImageSquarer(edgeRatioThreshold: .aspectFillThreshold).square(image)
-        case .edgeRatioDeterminesFitOrFill(let edgeRatio):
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/Gravatar/UIImage/SquaringStrategy.swift
```diff
@@ -0,0 +1,49 @@
+import UIKit
+
+public enum SquaringStrategy {
```
pinarol: We can keep things simple and move fwd with just 1 enum imo. No need for an extra `CroppingBehavior` enum.

```
enum SquaringStrategy {
    case aspectFill
    case aspectFit
    case threshold(value: CGFloat = 0.02)
    case none
}
```

We don't need a public `ImageSquarer` protocol, one can just pass `.none` to disable our squaring and do the squaring on their side before calling the upload method.

We don't also need a `.default` case, we can declare a default param in the upload method signature:

`upload(...., squaringStrategy: SquaringStrategy = .threshold())`

pinarol: And for the "threshold", I suggest a different naming and a different formula here: https://github.com/Automattic/Gravatar-SDK-iOS/pull/419/files#r1840399172

andrewdmontgomery: Ok, simpler.  I tried two different naming schemes:

- `edgeRatioDeterminesFitOrFill(edgeRatio: CGFloat)`
- `squarenessDeterminesFitOrFill(aspectFillMinSquareness: CGFloat)`

The first one basically describes the implementation details, which is ok – we're using "edgeRatio" as a measure of how close to square the image is.  But it obfuscates the goal, which is to make a decision based on how close to square the image is.

I'm hoping that the second one communicates the purpose better.  What do you think?  Does `squareness` make sense?  I have documented the meaning on the `squareness` function.

```swift
public enum SquaringStrategy {
    case aspectFill
    case aspectFit
    case squarenessDeterminesFitOrFill(aspectFillMinSquareness: CGFloat)
    case `default`
    case none

    package func square(_ image: UIImage) -> UIImage {
        switch self {
        case .aspectFit:
            ImageSquarer(aspectFillMinSquareness: .aspectFitThreshold).square(image)
        case .aspectFill:
            ImageSquarer(aspectFillMinSquareness: .aspectFillThreshold).square(image)
        case .squarenessDeterminesFitOrFill(let aspectFillMinSquareness):
            ImageSquarer(aspectFillMinSquareness: aspectFillMinSquareness).square(image)
        case .default:
            ImageSquarer(aspectFillMinSquareness: .defaultThreshold).square(image)
        case .none:
            // TODO: Check for squareness and log non-square images
            image
        }
    }
}

extension CGFloat {
    fileprivate static let defaultThreshold: CGFloat = 0.98
    fileprivate static let aspectFitThreshold: CGFloat = 1.0
    fileprivate static let aspectFillThreshold: CGFloat = 0.0
}
```

As you've pointed out, technically we only _need_ two cases:
- `.squarenessDetereminsFitOrFill()`
- `.none`

The others are derived.  But I think having them makes this all easier to understand and reason about:

- It's easy to reverse the logic of `fit` vs `fill`.  I reversed them multiple times during this refactor.  We can document that, of course.  But when debugging at the call site, being able to see `.aspectFill` instead of `.squarenessThreshold: 0.0` is really helpful.  We could just make those CGFloat `static let` entries public and use those: `squarenessThreshold: .aspectFillThreshold`.  But doesn't read as clearly, and it's harder to discover.
- The `default` case makes the calling site easier to understand.  And enums with associated values that have default values are awkward with autocomplete.

I'm not married to making these extra cases explicit.  If we want to prune down to just the base two cases, I can be convinced.

andrewdmontgomery: If this general approach seems like a good direction, I can merge this into the original branch and we can continue there.

pinarol: Yes `squareness` makes sense! And I also like the whole naming. Much better!

> case squarenessDeterminesFitOrFill(aspectFillMinSquareness: CGFloat)

Sources/Gravatar/UIImage/SquaringStrategy.swift
```diff
@@ -0,0 +1,58 @@
...
+    case aspectFill
+
+    /// Sqares the image using an `aspectFit` strategy
+    case aspectFit
+
+    /// The defaul squaring strategy
+    ///
+    /// This strategy uses a high aspectFillMinSquareness to determine `aspectFit` or `aspectFill`.  Only images that are very close to square will use
+    /// `aspectFill`, which minimizes the amount of image loss that happens during cropping.  All other images will use `aspectFit`.
+    case `default`
```
pinarol: Another option is to define `static let`s so enum cases don't intersect with each other logically. We would just be providing some pre-defined values that can be useful. They would still appear in auto-complete and nothing else in the code needs to change actually.

This feels semantically more accurate to me since now one case doesn't cover other cases. There's actually only 2 squaring strategies `squarenessDeterminesFitOrFill` and `none`. Other ones (default, aspectFill, aspectFit) are just variations of `squarenessDeterminesFitOrFill`. WDYT? 🤔 

```
public enum SquaringStrategy: Sendable {
     public static let `default`: SquaringStrategy = .squarenessDeterminesFitOrFill(aspectFillMinSquareness: 0.98)
     public static let aspectFill: SquaringStrategy = .squarenessDeterminesFitOrFill(aspectFillMinSquareness: 0.0)
     public static let aspectFit: SquaringStrategy = .squarenessDeterminesFitOrFill(aspectFillMinSquareness: 1.0)
   
     case squarenessDeterminesFitOrFill(aspectFillMinSquareness: CGFloat)
     case none
}

```

Sources/Gravatar/UIImage/SquaringStrategy.swift
andrewdmontgomery: Great!  Names are _hard_.  But I thought that read nicely last night.  I was just hoping I agreed with myself in the morning.

Sources/Gravatar/UIImage/SquaringStrategy.swift
andrewdmontgomery: Yes!  Of course.  This is excellent.  Well-encapsulated, easy to use, doesn't abuse cases of an enum.  Much better.  I'll make this change and then merge it into the original branch.  We can continue there.

andrewdmontgomery:
I'm going to merge this into the original branch (#419) now, and we can continue there.

PR #561, opened 2024-11-14
Add a default `ProfileViewDelegate` conformance to `ProfileViewController

### LINKED ISSUE #383: Allow to pass a delegate to `ProfileViewController`
### Current behavior

SDK consumer that uses `ProfileViewController` is not able to pass a `delegate` to apply on click actions on a profile card.

### Expected behavior

It is possible to make profile card interactive when using `ProfileViewController`.

### Workaround

For my case of using `SwiftUI` and `UIViewControllerRepresentable`, @pinarol suggested a solution with copying source code of `ProfileViewController` and `DemoProfileViewsViewController` from this repository and update `ProfileViewController#configuration#didSet` to apply a delegate.

This works fine, but requires SDK consumer to copy boilerplate code, that I'd rather prefer removed from my codebase.
Internal communication: p1725439264859349-slack-C06FRGK6Y5R

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/383

### Description

The original issue was described as:

> SDK consumer that uses ProfileViewController is not able to pass a delegate to apply on click actions on a profile card.

But in fact, it is possible via `ProfileViewConfiguration`:

```
let config: ProfileViewConfiguration = ProfileViewConfiguration.large(palette: preferredPaletteType)
config.delelgate = self
let viewController = ProfileViewController(configuration: config, viewModel: .init(), profileIdentifier: ...)
...
```

So I think there was a confusion there. 

But I still wanted to add a default `ProfileViewDelegate` conformance for convenience.

### Testing Steps

Demo app > Profile View Controller
Type email 
Tap "Show profile bottom sheet"
Tap "View profile"
It should open the in-app Safari
Same for the social media icons

### DIFF
--- a/Sources/GravatarUI/ProfileViewController/ProfileViewController.swift
+++ b/Sources/GravatarUI/ProfileViewController/ProfileViewController.swift
@@ -1,5 +1,6 @@
 import Combine
 import Gravatar
+import SafariServices
 import UIKit
 
 /// A view controller which displays a Profile View.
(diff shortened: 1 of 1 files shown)


etoledom (approved):
Looking good 🙏

PR #570, opened 2024-11-20
Add the initial avatar action menu [...]

### LINKED ISSUE #536: Quick Editor: More options button with empty dialog menu
This is to allow for parallel development. Add the [...] button on the avatar and add at least 1 menu item with no action. 

Image

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/536

### Description

Adds the menu with 2 sample items. At the moment the buttons don't do anything other than printing to the console.

### Testing Steps

The avatars should have the ... menu, and 2 menu items should appear.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
@@ -68,7 +68,6 @@ struct AvatarPickerAvatarView: View {
         }
     }
 
-    @ViewBuilder
     func ellipsisView() -> some View {
         Image("more-horizontal", bundle: Bundle.module).renderingMode(.template)
             .tint(.white)
@@ -77,7 +76,6 @@ struct AvatarPickerAvatarView: View {
             .padding(CGFloat.DS.Padding.half)
     }
 
-    @ViewBuilder
     func actionsMenu() -> some View {
         Menu {
             ForEach(AvatarAction.allCases) { action in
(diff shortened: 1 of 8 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
```diff
@@ -54,12 +55,46 @@ struct AvatarPickerAvatarView: View {
...
+    func ellipsisView() -> some View {
+        Image("more-horizontal", bundle: Bundle.module).renderingMode(.template)
+            .tint(.white)
+            .background(Color(uiColor: UIColor.gravatarBlack.withAlphaComponent(0.4)))
+            .cornerRadius(2)
+            .padding(CGFloat.DS.Padding.half)
+    }
+
+    @ViewBuilder
+    func actionsMenu() -> some View {
```
etoledom: I believe `@ViewBuilder` attribute is not needed here

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
```diff
@@ -54,12 +55,46 @@ struct AvatarPickerAvatarView: View {
...
+                }
             }
         }.onTapGesture {
             onAvatarTap(avatar)
         }
     }
+
+    @ViewBuilder
+    func ellipsisView() -> some View {
+        Image("more-horizontal", bundle: Bundle.module).renderingMode(.template)
```
etoledom, line 72: As an alternative: We can set the rendering mode directly in the asset catalog.
That's good if the icon is always meant to be used this way.

etoledom (approved):
Looking good!

Just a couple small comments with optional changes

🎉

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
pinarol: Right but asset catalogs are more fragile. It is easy to lose that setting if we change the icon. Setting it programmatically is safer.

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
pinarol: Right. Removing.

PR #571, opened 2024-11-21
Optimize avatar thumbnail size

pinarol:
Closes #

### Description

Our avatar grid items are 512px by default(server returns that way), which is bigger than what we need for a thumbnail. 

https://2.gravatar.com/userimage/110207384/45d513e372aff2b7d882025b4b99e036?size=512

Max avatar size in the grid is 100x100pt. So its at most 3x100px. We can use `AvatarGridConstants.maxAvatarWidth` instead.

### Testing Steps

QE Avatars should load without a problem.


etoledom (approved):
Nice! 🙏

PR #572, opened 2024-11-21
QE: Share image

### LINKED ISSUE #566: Share an image
Downloading an avatar image is accessible in the avatar’s contextual menu.

In iOS, a share sheet will be used. The menu item will be “Share” with a system icon described in the [Apple guidelines](https://developer.apple.com/design/human-interface-guidelines/activity-views#Best-practices).The icon in figma can be ignored.

pinarol:
Closes #566 

### Description

Opens a share sheet via the avatar menu. Downloads the image in full size before that. Otherwise the share sheet doesn't share the image. Also adding the URL because then the metadata is captured by the system and automatically displayed in the share sheet's navigation bar.

**Discussion**

I wonder how you feel about the medium detent for the share sheet. I added it to be able to test it, but not sure how I feel about it. Please also test the full size sheet by just commenting out `.presentationDetentsIfAvailable([.medium, .large])` and compare. 

### Testing Steps

Test with both horizontal and vertical layouts.
Try different color schemes.
Test an error case like for example turn off internet.

Demo app > Quick Editor > Avatar ... menu > Share


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -30,8 +31,7 @@ class AvatarPickerViewModel: ObservableObject {
     @Published var selectedAvatarURL: URL?
     @Published private(set) var backendSelectedAvatarURL: URL?
     @Published private(set) var gridResponseStatus: Result<Void, Error>?
-
-    let grid: AvatarGridModel = .init(avatars: [])
+    @Published private(set) var grid: AvatarGridModel = .init(avatars: [])
```
pinarol, line 34: This should have been a `@Published` var from the beginning imo. Otherwise the UI can fail to get updated properly.

andrewdmontgomery:
⛏️  small adjustment

We will want to [append an ellipsis](https://developer.apple.com/design/human-interface-guidelines/menus#Labels) to the "Share" menu item:

`Share...`

> Append an ellipsis to a menu item’s label when the action requires more information before it can complete. The ellipsis character (…) signals that people need to input information or make additional choices, typically within another view.

pinarol:
> We will want to append an ellipsis

Done.

andrewdmontgomery (commented):
In the simulator, the context menu seems to work fine.  But on a physical device, tapping the menu selects the image as an avatar.  Once it is selected, then the menu is tappable

pinarol:
I test in the real device as well. Maybe the tapping area wasn't big enough for you and we need to make it a bit bigger 🤔

pinarol:
The problem seems to be related with iOS 17.x. Putting the Menu in the overlay doesn't seem to work there. I made a fix. Feel free to check again. @andrewdmontgomery and good catch 🎉

Tests/GravatarUITests/AvatarPickerViewModelTests.swift
```diff
@@ -64,6 +68,103 @@ final class AvatarPickerViewModelTests {
         }
     }
 
+    @Test
+    func testFetchOriginalSizeAvatarSuccess() async throws {
+        await model.refresh()
+        guard let avatar = model.grid.avatars.first else {
+            #expect(Bool(false), "No avatar found")
+            return
+        }
```
andrewdmontgomery: This is a great place to use the `#require()` macro.  It is like `#expect()`, except that if it fails, the test returns immediately.  They are meant for cases where it doesn't make sense to continue if a specific test fails.  It's primary use case is unwrapping optionals, where if you get `nil`, it often doesn't make sense to continue.

```suggestion
        let avatar = try #require(model.grid.avatars.first, "No avatar found")
```

Tests/GravatarUITests/AvatarPickerViewModelTests.swift
```diff
@@ -64,6 +68,103 @@ final class AvatarPickerViewModelTests {
...
+    }
+
+    @Test
+    func testFetchOriginalSizeFailsWithURLSessionError() async throws {
+        let model = Self.createModel(imageDownloader: TestImageFetcher(result: .urlSessionError))
+        await model.refresh()
+        guard let avatar = model.grid.avatars.first else {
+            #expect(Bool(false), "No avatar found")
+            return
+        }
```
andrewdmontgomery: Another case for the `#require()` macro

Tests/GravatarUITests/AvatarPickerViewModelTests.swift
```diff
@@ -64,6 +68,103 @@ final class AvatarPickerViewModelTests {
...
+    }
+
+    @Test
+    func testFetchOriginalSizeFailsWithGenericError() async throws {
+        let model = Self.createModel(imageDownloader: TestImageFetcher(result: .fail))
+        await model.refresh()
+        guard let avatar = model.grid.avatars.first else {
+            #expect(Bool(false), "No avatar found")
+            return
+        }
```
andrewdmontgomery: Another case for the `#require()` macro

Tests/GravatarUITests/AvatarPickerViewModelTests.swift
```diff
@@ -64,6 +68,103 @@ final class AvatarPickerViewModelTests {
...
+        let model = Self.createModel(imageDownloader: TestImageFetcher(result: .urlSessionError))
+        await model.refresh()
+        guard let avatar = model.grid.avatars.first else {
+            #expect(Bool(false), "No avatar found")
+            return
+        }
+
+        await confirmation(expectedCount: 2) { confirmation in
+            model.toastManager.$toasts.sink { toasts in
+                #expect(toasts.count <= 1)
```
andrewdmontgomery, line 105: Is this meant to be `<=` (with the `if ==` that follows)?  When I run this test, it always seems to be `1`.

Tests/GravatarUITests/AvatarPickerViewModelTests.swift
```diff
@@ -64,6 +68,103 @@ final class AvatarPickerViewModelTests {
...
+        let model = Self.createModel(imageDownloader: TestImageFetcher(result: .fail))
+        await model.refresh()
+        guard let avatar = model.grid.avatars.first else {
+            #expect(Bool(false), "No avatar found")
+            return
+        }
+
+        await confirmation(expectedCount: 2) { confirmation in
+            model.toastManager.$toasts.sink { toasts in
+                #expect(toasts.count <= 1)
```
andrewdmontgomery, line 136: Same question, here:

Is this meant to be `<=` (with the `if ==` that follows)?  When I run this test, it always seems to be `1`.

andrewdmontgomery (approved):
Looks good to me.  I ran this in iOS 18 in a simulator (iPhone and iPad, in both orientations) and iOS 17.5.1 on an iPhone.  On the physical device, I used network conditioning to simulate a bad network connection and then 100% packet loss.  It behaved as expected:
- bad network connection: it took longer, but it succeeded
- 100% packet loss: timed out before showing the sheet, showed a toast about the timeout

I left a few small suggestions and some questions.  I am AFK this week.

Tests/GravatarUITests/AvatarPickerViewModelTests.swift
pinarol: Great idea! :TIL: 👍

Tests/GravatarUITests/AvatarPickerViewModelTests.swift
pinarol: It is not always `1`. `sink` receives values multiple times, first the initial value then the new values. 

The first time it is called the count is `0`. Not sure why it always seems `1` in your case 🤔.

Screenshot 2024-11-26 at 11 32 52

pinarol: `==` fails:

Screenshot 2024-11-26 at 11 37 11

Tests/GravatarUITests/AvatarPickerViewModelTests.swift
pinarol: [Same](https://github.com/Automattic/Gravatar-SDK-iOS/pull/572#discussion_r1858042332).

Tests/GravatarUITests/AvatarPickerViewModelTests.swift
andrewdmontgomery: In a strange coincidence, I ran into this over my break in a personal project.

PR #574, opened 2024-11-26
Delete avatar

### LINKED ISSUE #544: New internal API: Delete image
Add a service call to delete an image (It can be added to ProfileService but we can split ProfileService if it gets too big)

This depends on the BE task: https://github.a8c.com/Automattic/gravatar/issues/109127

etoledom:
Closes #544 

### Description

Implementing avatar deletion.

CleanShot 2024-11-26 at 11 40 00

### Testing Steps

- Run the demo app.
- Go to `Quick Editor`.
- Open the QE.
- Tap on the action button (elipsis) on any avatar).
- Tap on Delete action
  - A confirmation sheet should appear. 
  - The avatar should be deleted after confirmation.
- For 404 errors, the UX will appear as a success.
- For any other error, a toast will appear with a message and the avatar will animate back into the grid.


Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -76,4 +78,27 @@ public struct AvatarService: Sendable {
...
+            case .URLSessionError(let urlSessionError):
+                throw AvatarDeleteError.responseError(reason: .URLSessionError(error: urlSessionError))
+            case .invalidHTTPStatusCodeError(let response, _):
+                throw AvatarDeleteError.responseError(reason: .invalidHTTPStatusCode(response: response, errorPayload: nil))
+            case .invalidURLResponseError(let urlResponse):
+                throw AvatarDeleteError.responseError(reason: .invalidURLResponse(response: urlResponse))
+            }
+        } catch {
+            throw AvatarDeleteError.responseError(reason: .unexpected(error))
+        }
```
pinarol: Why not just convert the error like this like we do in all other API calls?

```
        do {
            let (_, response) = try await client.data(with: authorizedRequest)
        } catch {
            throw error.apiError()
        }
```

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -326,6 +326,23 @@ class AvatarPickerViewModel: ObservableObject {
...
+        do {
+            try await avatarService.delete(avatarID: avatar.id, accessToken: token)
+        } catch AvatarDeleteError.responseError(let reason) where reason.httpStatusCode == 404 {
+            return // no-op. We delete a not-found avatar from the UI.
+        } catch {
+            withAnimation {
+                grid.insert(avatar, at: deletedIndex)
+            }
+            toastManager.showToast(Localized.avatarDeletionError, type: .error)
+        }
```
pinarol: Can we also handle the URLSession error like we do in other service calls? This way it'll tell me about the connection errors as well.

```
        do {
            try await avatarService.delete(avatarID: avatar.id, accessToken: token)
        } catch APIError.responseError(let reason) where reason.httpStatusCode == 404 {
            return // no-op. We delete a not-found avatar from the UI.
        } catch APIError.responseError(reason: let reason) where reason.urlSessionErrorLocalizedDescription != nil {
            handleError(message: reason.urlSessionErrorLocalizedDescription ?? Localized.avatarDeletionError)
        } catch {
            handleError(message: Localized.avatarDeletionError)
        }
        // Can be an internal function inside delete(...)
        func handleError(message: String) {
            withAnimation {
                grid.insert(avatar, at: deletedIndex)
            }
            toastManager.showToast(message, type: .error)
        }
```

Sources/Gravatar/Network/Errors.swift
```diff
@@ -121,3 +121,7 @@ extension APIError: CustomDebugStringConvertible {
         }
     }
 }
+
+package enum AvatarDeleteError: Error {
```
pinarol: We can use the existing `APIError`, no?

Sources/Gravatar/Network/Services/HTTPClient.swift
```diff
@@ -8,7 +8,7 @@ protocol HTTPClient: Sendable {
     /// Performs a data request using the  information provided,  and delivers the result asynchronously.
     /// - Parameter request: A URL request object that provides request-specific information such as the URL and cache policy.
     /// - Returns: An asynchronously-delivered tuple that contains the URL contents as a Data instance, and a HTTPURLResponse.
-    func fetchData(with request: URLRequest) async throws -> (Data, HTTPURLResponse)
+    func data(with request: URLRequest) async throws -> (Data, HTTPURLResponse)
```
pinarol, line 11: Is there a special reason why you wanted to rename this one? I think both is fine, I am just curious.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -350,6 +367,11 @@ extension AvatarPickerViewModel {
             value: "The provided image exceeds the maximum size: 10MB",
             comment: "Error message to show when the upload fails because the image is too big."
         )
+        static let avatarDeletionError = SDKLocalizedString(
+            "AvatarPickerViewModel.Delete.Error",
+            value: "Error deleting the image.",
```
pinarol: For sake of consistency, can we make this error message more similar to the existing ones? Like:

> Oops, there was an error deleting the image.

Or something similar...

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -116,6 +118,24 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
                 } message: { error in
                     Text(error.errorMessage)
                 }
+                .confirmationDialog(
+                    Localized.deletionConfirmationTitle,
+                    isPresented: $isAvatarDeletionDialogPresented,
```
pinarol: (Optional) We can also do this trick to avoid having a separate boolean to present the dialog:

```
isPresented: Binding(
            get: { avatarToDelete != nil },
            set: { if !$0 { avatarToDelete = nil } }
        )
``` 
This way it is possible to get rid of the var `isAvatarDeletionDialogPresented`

Sources/Gravatar/Network/Services/ImageUploadService.swift
```diff
@@ -102,7 +102,7 @@ extension URLRequest {
 }
 
 extension URL {
-    fileprivate static var imageUploadURL: URL {
+    static var avatarsURL: URL {
```
pinarol: Since this URL is now reused we can maybe move this `extension URL` to its own file like `URL+Endpoints.swift` or something like that.

Sources/Gravatar/Network/Services/HTTPClient.swift
etoledom: "fetch" sound to me specific to `GET` data. Fetching as "obtaining" or "bringing".
So by calling it "data", it could refer to `POST` (or any method that sends/modify data on the backend), where we need to send data.

It just feels a bit wrong to call fetch method where we are not obtaining/bringing data (or is not the main purpose).

Not a strong feeling, more as a proposal. But seems more generic to me this way.

pinarol: ooh right!

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
etoledom: Let's!
I was following the docs, but this is better

pinarol (approved):
Looks good and works like a charm 🎉

PR #576, opened 2024-11-27
Improve share sheet

pinarol:
An improvement on #566 

### Description

I noticed that if we first write the image to a file and then share that file's URL, then there are more relevant options in the share sheet:

Also, updated the share sheet's initial height as same as QE's initial height (in horizontal mode it's not exactly same but it's close). So we see small or no height adjustment done by the system while share sheet is being opened.

Another thing is, I changed the word "download" as "share" in the error message.

Last but not least, I share the jpegData as there's a noticeable time difference between sharing pngData and jpegData. I set the compression quality to 1 (highest), and the shared image's quality seems good to me.

### Testing Steps

(It can be good to compare with the previous version.)
Demo app > QE > avatar [...] menu > share


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -124,10 +124,23 @@ class AvatarPickerViewModel: ObservableObject {
...
+              let imageData = image.jpegData(compressionQuality: 1) else { return nil }
+        let fileURL = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent("image.jpg")
+        do {
+            try imageData.write(to: fileURL)
+            return fileURL
+        } catch {
+            toastManager.showToast(Localized.avatarShareFail, type: .error)
         }
         return nil
     }
```
etoledom: Generally is good to have these functions which explicitaly do two things to have them split, even if they are two small functions like:
```
let image =  fetchFullImage(of: avatar)`
let imageFileURL = image.saveToFile()
```

Maybe in this way we can also show the share sheet even if writing to storage fails 🤔

etoledom (approved):
Working great 🎉 

Just a small option to consider on comments, but not a blocker 👍

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
pinarol: Yeah totally makes sense to have a separate UIImage extension method like `saveToFile()`. I'll add it.

> Maybe in this way we can also show the share sheet even if writing to storage fails

This is doable but I would consider this as an edge case. For sake of keeping our logic simple, I'd say let's just share the file for now.

PR #578, opened 2024-11-28
Fix - "Wrong email" error is NOT shown if the previous email's token is expired

### LINKED ISSUE #579: "Wrong email" error is NOT shown if the existing email's token is expired
This is an edge case but can be good to fix.

1. Open QE with `email_1`, login
2. Wait until `email_1`'s token is expired (well ofc just use .com oauth settings to expire it)
3. Open QE with `email_2`
4. The OAuth flow starts, instead of completing the flow with `email_2`, complete it with `email_1`(it's already the existing session in the browser)
Result: "Session expired" error is shown
Expected: "It looks like you used the wrong email..." error is shown.

(For step 2, you can also expire the token programmatically by manipulating the code and calling `OAuthSession.overrideToken(...)` but  .com oauth settings is easier)

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/579


pinarol:
Fixed with https://github.com/Automattic/Gravatar-SDK-iOS/pull/626

PR #580, opened 2024-11-28
Implementing support for HTTPS scheme in oauth callback url for the demo app

### LINKED ISSUE #543: Find a way to make the OAuth flow work for external contributors on the demo app
It'd be great if we could define a real universal link for the demo app.

etoledom:
Closes #543 

### Description

This PR implements `https` scheme oauth callback url using an Universal Link setup.

In this way, external developers can use this same callback url on their gravatar apps to test oauth on the demo app.
We also gain the ability to run and test the universal link execution of the oath flow on the demo app.

### Testing Steps

- Oauth on the Quick Editor should work as normal.
  - On this path, the dismissal of the webview takes half a second longer than it did before. That's expected.


pinarol (approved):
Great 🎉

Demo/Demo/Gravatar-Demo/SceneDelegate.swift
```diff
@@ -47,6 +48,12 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         // to restore the scene back to its current state.
     }
 
+    func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
+        guard let url = userActivity.webpageURL else { return }
```
pinarol, line 52: Should we check the `webpageURL` to see if it's the one we want as a general practice?

pinarol:
Should we target `release/3.1.0` ?

We can also rebase the release branch after merging this, unless there's something in the `trunk` that we want to exclude.

Demo/Demo/Gravatar-Demo/SceneDelegate.swift
etoledom: As it's the only url callback this app is set to handle, I don't think it's necessary 🤔 
Either way if we pass a wrong URL, the method will return `false`.

Also, if in the future the callback url changes for some reason, at least this part of the code won't need to change

etoledom:
> Should we target `release/3.1.0` ?
> 
> We can also rebase the release branch after merging this, unless there's something in the `trunk` that we want to exclude.

Ok, let's target 3.1.0

PR #583, opened 2024-11-29
Adding tests to avatar deletion flow

etoledom:
Closes #

### Description

Adding unit tests to the Avatar deletion flow.

### Testing Steps

- CI should be 🍏 
- Smoke test Avatar deletion

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarGridModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarGridModel.swift
@@ -66,8 +66,8 @@ class AvatarGridModel: ObservableObject {
         }
     }
 
-    func deleteModel(_ id: String) -> Int {
-        let index = avatars.firstIndex { $0.id == id }!
+    func deleteModel(_ id: String) -> Int? {
+        guard let index = avatars.firstIndex(where: { $0.id == id }) else { return nil }
         avatars.removeAll { $0.id == id }
         if selectedAvatar?.id == id {
             selectedAvatar = nil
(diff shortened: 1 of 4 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarGridModel.swift
```diff
@@ -66,8 +66,8 @@ class AvatarGridModel: ObservableObject {
         }
     }
 
-    func deleteModel(_ id: String) -> Int {
-        let index = avatars.firstIndex { $0.id == id }!
+    func deleteModel(_ id: String) -> Int? {
+        guard let index = avatars.firstIndex(where: { $0.id == id }) else { return nil }
         avatars.removeAll { $0.id == id }
```
etoledom, line 71: Nasty force unwrap went through. 
It was probably this new TAB autocomplete thing and I didn't notice it 🤦 

Deleting an avatar from the grid can fail if it's not found, in this case the deletion index returned is `nil`.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -368,7 +370,7 @@ class AvatarPickerViewModel: ObservableObject {
             try await avatarService.delete(avatarID: avatar.id, accessToken: token)
             return true
         } catch APIError.responseError(let reason) where reason.httpStatusCode == 404 {
-            return false // no-op. We delete a not-found avatar from the UI.
+            return true // no-op. We delete a not-found avatar from the UI.
         } catch APIError.responseError(reason: let reason) where reason.urlSessionErrorLocalizedDescription != nil {
```
etoledom, line 376: Returning `false` on this error could prevent an avatar changed notification. We want the notification as we are effectively removing this avatar from the grid, so we treat it as a success.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarGridModel.swift
pinarol: Oh no, I had overlooked it as well.

pinarol (approved):
Awesome! 🎉

PR #589, opened 2024-12-05
QE: Avatar Rating

andrewdmontgomery:
Closes #546 #548 

### Description

Adds support for viewing and changing the rating of an avatar in the Quick Editor.

### Testing Steps

1. Run the Demo app
2. Choose SwiftUI Demos --> Profile editor with oauth --> Tap `Open Profile Editor with OAuth flow`
3. Tap the action menu `...`
   - OBSERVE:
      - [x] The `Rating` menu appears in the menu
      - [x] The `Rating` menu shows the current rating for the image: `Rating: PG`
      - [x] The `Rating` menu has a icon of a half-filled star ⭐ 
4. Tap the `Rating` menu to show the sub-menu
   - OBSERVE:
      - [x] All rating options are visible, with descriptions (`G (General)`)
      - [x] The selected rating has a checkmark (✔️)
5. Choose a different rating
   - OBSERVE:
      - [x] The rating for the image changes on the device
      - [x] The rating for the image changes on web

Try:
- different layouts
- dark/light
- iPad

PR #590, opened 2024-12-10
Alt text editor

### LINKED ISSUE #535: QuickEditor: Update the ALT text for avatar
Image

etoledom:
Closes #535 

### Description

Adding the option to edit the Alt Text of an avatar.

CleanShot 2024-12-10 at 15 23 47

I've spent a lot of time trying to make the content-height bottom sheet work unsuccessfully.
So I decided to make this first PR with it showing full screen. As the keyboard appears automatically, it's not too bad.

I've also made small modifications to the `.gravatarNavigation` modifier, to share common behaviors.

### Testing Steps

- Open Quick Editor on the demo app.
- Tap the Alt Text option on an avatar's contextual menu.
- Edit the alt text and save.
  - New alt text should be reflected on web.


Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
```diff
@@ -0,0 +1,173 @@
...
+                    }
+                }
+                actionButton
+            }
+            .padding()
+            .avatarPickerBorder(colorScheme: .light)
+            Spacer()
+        }
+        .padding()
+        .gravatarNavigation(
```
pinarol, line 52: The problem about reusing the same modifier for both QE and the alt text is that it breaks the intrinsic height calculation because `GravatarNavigationModifier` uses `InnerHeightPreferenceKey` to emit the navigation bar height. So the navigation bar's height is added to the QE height "twice" when the alt text sheet opens. You can verify this by putting a print statement in `QuickEditorViewController`

```                
.onPreferenceChange(InnerHeightPreferenceKey.self) {
          print("new Height: \($0)")
```

To prevent this, we need to have 2 different preference keys, one for the main QE sheet, another for the alt text sheet. 

I drafted a solution here https://github.com/Automattic/Gravatar-SDK-iOS/pull/593 

And when we want to show the alt text sheet with intrinsic height we should call the function like this for the inner elements of alt text sheet. We don't need to do that in this PR but just saying because this is maybe why it didn't work properly before.

> accumulateIntrinsicHeight(key: AltTextHeightPreferenceKey.self)

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -337,6 +337,28 @@ class AvatarPickerViewModel: ObservableObject {
         await profile
     }
 
+    func update(_ avatar: AvatarImageModel, altText: String) async -> Bool {
+        guard let token = self.authToken else { return false }
+        do {
+            try await avatarService.update(altText, rating: nil, avatarID: avatar.id, accessToken: token)
+            toastManager.showToast(Localized.avatarAltTextSuccess + "\n- \"\(altText)\"")
```
pinarol: `-` combined with quotes looks a bit confusing to me. how about just one or two new lines?

<table>
  <tr>
    <td></td>
    <td></td>
    <td></td>
  </tr>
</table>

pinarol: I think we should also put a `.` at the end of the success message, wdyt? 

> "Image alt text was changed successfully."

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -337,6 +337,28 @@ class AvatarPickerViewModel: ObservableObject {
...
+            withAnimation {
+                grid.replaceModel(withID: avatar.id, with: avatar.updating(altText: altText))
+            }
+            return true
+        } catch APIError.responseError(reason: let reason) where reason.urlSessionErrorLocalizedDescription != nil {
+            handleError(message: reason.urlSessionErrorLocalizedDescription ?? Localized.avatarDeletionError)
+        } catch {
+            handleError(message: Localized.avatarDeletionError)
+        }
```
pinarol: `avatarDeletionError`? It can be good to write some unit tests and check the error message as this area is prone to copy paste errors.

pinarol: there are some similar test cases as reference:

`testFetchOriginalSizeFailsWithGenericError`
`testFetchOriginalSizeFailsWithURLSessionError`

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
```diff
@@ -0,0 +1,173 @@
+import SwiftUI
+
+struct AltTextEditorView: View {
```
pinarol, line 3: The alt text sheet doesn't respect the colorScheme of the QE, seems like we should pass that environment variable when opening it.

> .colorScheme(colorScheme)

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
```diff
@@ -46,14 +47,19 @@ struct AvatarImageModel: Hashable, Identifiable, Sendable {
...
+        self.altText = altText
     }
 
     func settingStatus(to newStatus: State) -> AvatarImageModel {
-        AvatarImageModel(id: id, source: source, state: newStatus, isSelected: isSelected)
+        AvatarImageModel(id: id, source: source, state: newStatus, isSelected: isSelected, altText: altText)
+    }
+
+    func updating(altText: String) -> AvatarImageModel {
+        AvatarImageModel(id: id, source: source, state: state, isSelected: isSelected, altText: altText)
```
pinarol: the default values for the init and these methods combined will likely become source of new bugs if we don't be careful. 

@andrewdmontgomery also added new param here for the rating and it has a default value as well. 

These methods 
`updating(altText ...`, `settingStatus(to...` won't have any conflict or compile error when 2 changes merge but they will actually override the rating for example.

I think we should get rid of these methods. We just need a copy() method. And then we can alter whatever we want in that copy.

pinarol: Doesn't have to be in this PR but we really need to change this next.

pinarol: The more I think, the less reasons I can find to have these methods. They aren't doing any deep copying that can't be done via:

var model = otherModel

Why do we have these methods again?

andrewdmontgomery:
#589 (QE: Avatar Rating) is ready to merge.  It makes a few changes that will create conflicts with this PR.  The noteworthy conflicts are:
- The organization of the `actionsMenu` in the `AvatarPickerAvatarView` is now manually arranged, and no longer loops through `AvatarAction.allCases`.
   - This means `AvatarAction` no longer needs to be `CaseIterable`
- The `AvatarAction` menu adds a case with an associated type (`rating(AvatarRating)`)
   - This means `AvatarAction` can no longer conform to `String`

It also refactored some of the unit test helpers in `AvatarPickerAvatarViewTests`.  Hopefully this will help with tests for this change, too.

Let me know if I can help resolve those conflicts when you are ready.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
andrewdmontgomery: > won't have any conflict or compile error when 2 changes merge but they will actually override the rating for example.

That's a good catch.  That bug would be no fun to find.  Even if we got rid of the default values in the init (which I don't think we would want to do anyway), we'd also have to guarantee that we don't introduce any default values in the future.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
andrewdmontgomery: I just merged #589 .  I refactored the helpers related to these tests, to make it easier to handle new API calls.  You'll probably want to merge that before adding tests.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
etoledom: I like the third screenshot the most. Let's go with that one 👍

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
etoledom: I was working on some test before checking that PR. I'll indeed wait for the merge before continuing.

> It can be good to write some unit tests and check the error message as this area is prone to copy paste errors.

Totally! It was a mistake actually, not done in purpose 😅

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
andrewdmontgomery: +1

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -89,4 +89,25 @@ public struct AvatarService: Sendable {
             throw error.apiError()
         }
     }
+
+    @discardableResult
+    package func update(
+        _ altText: String?,
+        rating: AvatarRating?,
+        avatarID: String,
```
andrewdmontgomery: For consistency with the other functions in `AvatarService`, what do you think about using the `AvatarIdentifier` type (instead of `String`) here?

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
etoledom: These init defaults are only meant to make the previews easier (to avoid setting all the fields all the time). There's no other need of them. I'd vote for removing the default values from the init to be able to keep this struct completely immutable.

> var model = otherModel

Two things I don't like about this approach:

- The structs fields will need to become `var` instead of `let`.
- We will create a mutable copy in the scope which will live together with the previous instance, and this could also create some issues.

Now I understand that the current state is not good, so I'll try to address all the feedback and propose a solution which will have the benefits of both approaches.

Oh, and there's one more thing. Using the request response to update the grid instead of manually updating the avatar model (as it was done on the Ratings), will remove the need of this new update method, so I'll do it this way also.

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
etoledom: I've implemented a small workaround to avoid the AvatarPicker size change when presenting the Alt Text editor for now.
Ideally we can make the intrinsic sizing work properly on a next PR.

etoledom:
@pinarol - @andrewdmontgomery - I believe this is ready for another look.

I've added some extra changes on a [separated PR](https://github.com/Automattic/Gravatar-SDK-iOS/pull/597), since they are a bit more disruptive and in this way we can discuss them without blocking this merge.

pinarol:
> I believe this is ready for another look.

[This thread](https://github.com/Automattic/Gravatar-SDK-iOS/pull/590/files#r1880881777) is missed maybe? I still see `avatarDeletionError`.

etoledom:
> [This thread](https://github.com/Automattic/Gravatar-SDK-iOS/pull/590/files#r1880881777) is missed maybe? I still see `avatarDeletionError`.

My bad. I created the string but didn't change it 🤦‍♂️ 
Fixed now.

pinarol:
> My bad. I created the string but didn't change it 🤦‍♂️
> Fixed now.

Would you mind also adding unit tests since apparently this part is easy to miss. I mentioned some similar unit tests [here](https://github.com/Automattic/Gravatar-SDK-iOS/pull/590#discussion_r1879822925) so it's really just copying them and replacing some lines. Let me know if you face with any difficulty doing that.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -338,19 +341,44 @@ class AvatarPickerViewModel: ObservableObject {
...
+        guard let token = self.authToken else { return false }
+        do {
+            let updatedAvatar = try await avatarService.update(altText: altText, avatarID: .hashID(avatar.id), accessToken: token)
+            toastManager.showToast(Localized.avatarAltTextSuccess + "\n\n \"\(altText)\"")
+            withAnimation {
+                grid.replaceModel(withID: avatar.id, with: .init(with: updatedAvatar))
+            }
+            return true
+        } catch APIError.responseError(reason: let reason) where reason.urlSessionErrorLocalizedDescription != nil {
+            handleError(message: reason.urlSessionErrorLocalizedDescription ?? Localized.avatarDeletionError)
```
pinarol: There's still `avatarDeletionError` here.

etoledom:
> Would you mind also adding unit tests

On it. Though I plan on a next PR for them, to not keep on waiting on this one.

pinarol (approved):
Looks good 👍

PR #591, opened 2024-12-10
Show no selected avatar warning

### LINKED ISSUE #582: No selected avatar banner
From [figma](https://www.figma.com/design/xgXj4jig9m0DYfKjFq6swG/QE-Native-Experience?node-id=44-1094&node-type=canvas&t=RLXzc0XbhPx4aXBH-0):

> I believe it should be dismissible. And we can show it again once they open the Editor or if the normal pattern for triggering the error happens again.

Image

pinarol:
Closes #582 

### Description

Adding a warning for the state where no avatar is selected even though there are avatars.

Designer's note from figma:

> I believe it should be dismissible. And we can show it again once they open the Editor or if the normal pattern for triggering the error happens again.

So I made it so that it is dismiss-able but it will re-appear if the number of avatars changes as `> 0` and the selected avatar changes as `nil`.

 <tr>
    <td></td>
    <td></td>
  </tr>
</table>

I also set a placeholder to the profile view. It was lacking and not looking good. I used the same style we used for the UIKit profile view claim profile state.

### Testing Steps

Demo app > Quick Editor 
Delete your selected avatar and you should see the warning
Close/open the Quick Editor and you should still see the warning
Close the warning by tapping `X`
Close/open the Quick Editor and you should see the warning
Upload a new avatar
The new avatar automatically becomes _the_ avatar and the warning goes away.

What else? Feel free to get creative about testing this one.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
@@ -24,6 +24,7 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
     @State private var avatarToDelete: AvatarImageModel?
     @State private var shareSheetItem: AvatarShareItem?
     @State private var playgroundInputItem: PlaygroundInputItem?
+    @State private var shouldDisplayNoSelectedAvatarWarning: Bool = false
 
     var contentLayoutProvider: AvatarPickerContentLayoutProviding
     var customImageEditor: ImageEditorBlock<ImageEditor>?
@@ -77,6 +78,8 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
         ZStack {
             VStack(spacing: 0) {
                 EmailText(email: model.email)
+                    .accumulateIntrinsicHeight()
+                noSelectedAvatarWarning()
                     .accumulateIntrinsicHeight()
                 profileView()
                     .accumulateIntrinsicHeight()
@@ -168,6 +171,12 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
         .onChange(of: model.backendSelectedAvatarURL) { _ in
             notifyAvatarSelection()
         }
+        .onChange(of: model.selectedAvatarURL) { _ in
+            updateShouldDisplayNoSelectedAvatarWarning()
+        }
+        .onChange(of: model.grid.avatars.count) { _ in
+            updateShouldDisplayNoSelectedAvatarWarning()
+        }
         .sheet(item: $shareSheetItem) { item in
             ShareSheet(items: [item.fileURL])
                 .colorScheme(colorScheme)
@@ -186,6 +195,10 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
                 uploadImage(image)
             }
         ))
+    }
+
+    private func updateShouldDisplayNoSelectedAvatarWarning() {
+        shouldDisplayNoSelectedAvatarWarning = model.selectedAvatarURL == nil && model.grid.avatars.count > 0
     }
 
     private func header() -> some View {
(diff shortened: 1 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -188,6 +197,10 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
         ))
     }
 
+    private func updateShouldDisplayNoSelectedAvatarWarning() {
+        shouldDisplayNoSelectedAvatarWarning = model.selectedAvatarURL == nil && model.grid.avatars.count > 0
+    }
```
etoledom, line 202: Maybe we can use `withAnimation` here also?

I gave it a fast try and I'm not convinced of the kind of animation it gives, but not animation at all doesn't feel great either 🤔 

There's specially an odd effect on the profile avatar:

CleanShot 2024-12-13 at 13 08 19

Not sure if we can be more specific about how to animate the elements. But either way I think a small smooth transition feels better than instant change.

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -142,4 +142,13 @@ extension View {
...
     }
+
+    @ViewBuilder
+    func `if`(_ condition: Bool, transform: (Self) -> some View) -> some View {
+        if condition {
+            transform(self)
+        } else {
+            self
+        }
+    }
```
etoledom, line 153: Hah nice one!

etoledom (approved):
Looking good!

I have this feedback about the animation I mentioned in code, but it's not a blocker.

Appart from that, there are a couple of new deprecation warnings, from the deprecation of `AvatarView` init.

One on: `AvatarPickerAvatarView`.
Other one on: `AvatarView` preview.

It would be great to remove those before merging. Other than that, 🚢 🎉

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
pinarol: The warning intertwines with the profile card during animation and looks broken. Especially when disappearing.

UIKit:


SwiftUI:


The non-animated version doesn't have that problem. Plus, in the UIKit version, the sheet's overall height changes animatedly and feels kinda sufficient I think.

pinarol:
> Appart from that, there are a couple of new deprecation warnings, from the deprecation of AvatarView init.

> One on: AvatarPickerAvatarView.

I'll address this one.

> Other one on: AvatarView preview.

We should keep this one to be able to test the deprecated init. Even though it is deprecated we still need to make sure it works until we remove it.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
etoledom: All those jumps are not good. 
I agree that the transition animation looks broken also. 

With UIKit UIStackViews we have good control over animations and customization. Maybe we can make a good customization of animations here also, but I don't know how is done on SwfitUI and sure will take long.

Let's move forward as it is then. But apart from this, there are many instance of instant change that I'd love to add a little animation. I've failed every time I've tried 😓  It's a whole topic where we can improve.

pinarol: Trying to animate the disappearance of a child view together with the bottom sheet height change is tricky even if we try to design it off-code via a design tool. Maybe we should design the animation first outside of code and then see how we can apply it in the code.

PR #595, opened 2024-12-11
Merge Ratings into AtlText

andrewdmontgomery:
### Description

Now that #589 (QE: Avatar Rating) PR just merged into `trunk`, this merges `trunk` back into #590 (AltText Editor).

- merges `trunk` (including the new Avatar Rating capability)
- Adds `.altText` into the new Menu structure
- Renames `setRating(_ rating: AvatarRating, for avatar: AvatarImageModel) async` --> `update(_ avatar: AvatarImageModel, rating: AvatarRating) async -> Bool`
- Makes the `update()` functions `@discardableResult`
   - `AvatarService.update(::::)`
   - `AvatarPickerViewModel.update(_: altText)`
   - `AvatarPickerViewModel.update(_: rating)`
- Removes the `ProfileService.setRating(:::)` function in favor of the `AvatarService.update(::::)` function in the AltText PR
- Updates related unit tests

### Testing Steps

- Make sure the AltText changes are still intact
- Make sure the AltText and Rating menu items work as expected

### DIFF
--- a/Sources/Gravatar/Network/Services/AvatarService.swift
+++ b/Sources/Gravatar/Network/Services/AvatarService.swift
@@ -90,7 +90,12 @@ public struct AvatarService: Sendable {
         }
     }
 
-    package func update(_ altText: String?, rating: AvatarRating?, avatarID: String, accessToken: String) async throws {
+    package func update(
+        _ altText: String?,
+        rating: AvatarRating?,
+        avatarID: String,
+        accessToken: String
+    ) async throws -> Avatar {
         var request = URLRequest(url: .avatarsURL.appendingPathComponent(avatarID))
         request.httpMethod = "PATCH"
         let updateBody = UpdateAvatarRequest(rating: rating, altText: altText)
@@ -98,7 +103,8 @@ public struct AvatarService: Sendable {
 
         let authorizedRequest = request.settingAuthorizationHeaderField(with: accessToken)
         do {
-            _ = try await client.data(with: authorizedRequest)
+            let (data, _) = try await client.data(with: authorizedRequest)
+            return try data.decode()
         } catch {
             throw error.apiError()
         }
--- a/Sources/Gravatar/Network/Services/AvatarService.swift
+++ b/Sources/Gravatar/Network/Services/AvatarService.swift
@@ -90,6 +90,7 @@ public struct AvatarService: Sendable {
         }
     }
 
+    @discardableResult
     package func update(
         _ altText: String?,
         rating: AvatarRating?,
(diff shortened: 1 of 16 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -90,15 +90,22 @@ public struct AvatarService: Sendable {
         }
     }
 
-    package func update(_ altText: String?, rating: AvatarRating?, avatarID: String, accessToken: String) async throws {
+    @discardableResult
+    package func update(
+        _ altText: String?,
```
andrewdmontgomery, line 95:
```suggestion
        altText: String?,
```

Now that we have multiple parameters that we can "update", I think the calling site might read better if we name the first parameter.  What do you think?

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -90,15 +90,22 @@ public struct AvatarService: Sendable {
...
+        rating: AvatarRating?,
+        avatarID: String,
+        accessToken: String
+    ) async throws -> Avatar {
```
andrewdmontgomery, line 99: We don't currently use this return data.  But since the API defines a return type for this, I thought we probably should return it, and make it @discardableRersult.

That does mean that if we can't decode it (which I would expect to be quite rare since it would be a backend error that breaks the API), this will throw an error.

What do you think?

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -337,6 +338,7 @@ class AvatarPickerViewModel: ObservableObject {
         await profile
     }
 
+    @discardableResult
```
andrewdmontgomery, line 341: Reasonable?  Or do we want to check the result?

Sources/Gravatar/Network/Services/AvatarService.swift
andrewdmontgomery: Eventually, if (when?) we make this part of the API public, we can follow the pattern of:

```swift
public update(::::) --> AvatarType
package update(::::) --> Avatar
```

etoledom (approved):
Looks super good!
Thank you for helping with the merge 🙏

Sources/Gravatar/Network/Services/AvatarService.swift
etoledom: Sounds good!  👍

Sources/Gravatar/Network/Services/AvatarService.swift
etoledom: Let's go with this change 👍

etoledom (approved):

PR #596, opened 2024-12-12
Update the cropper to have both circle and square guidelines

### LINKED ISSUE #594: Make the cropper use square + circle layer
We can do something similar to Android and put a circle hole in the middle with a square frame around it

pinarol:
Closes #594 

### Description

Updates the image cropper to have both circle and square guidelines.

### Testing Steps

Demo app > Image cropper (in debug mode)
OR,
you can try uploading a new image via Quick Editor.


etoledom (approved):
Looks good and works perfectly as described 🎉

PR #597, opened 2024-12-12
New `updating` method for avatar model changes

etoledom:
Closes #

### Description

This is an alternative to the current way of updating a field in the avatar model.

- Removing the defaults from `AvatarImageModel` init. (this might be enough already to avoid possible issues)
- Creating a helper static method for previews and unit tests (to make initialization shorter). Not a must but some help.
  - I'd like to make this one accessible only by the preview, but I didn't find a way to do this.
- Creating a new way to create a new model updating just one or a handful of fields. 
  - I like this one since we don't need to make the struct or their fields mutable. But it's not a must.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
@@ -57,9 +57,9 @@ struct AvatarImageModel: Hashable, Identifiable, Sendable {
         self.altText = altText
     }
 
-    func updating(_ callback: (inout Builder) -> Void) -> AvatarImageModel {
+    func updating<Value>(_ keyPath: WritableKeyPath<AvatarImageModel.Builder, Value>, to newValue: Value) -> AvatarImageModel {
         var builder = Builder(self)
-        callback(&builder)
+        builder[keyPath: keyPath] = newValue
         return builder.build()
     }
 }
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
@@ -57,9 +57,9 @@ struct AvatarImageModel: Hashable, Identifiable, Sendable {
         self.altText = altText
     }
 
-    func updating<Value>(_ keyPath: WritableKeyPath<AvatarImageModel.Builder, Value>, to newValue: Value) -> AvatarImageModel {
+    func updating(_ callback: (inout Builder) -> Void) -> AvatarImageModel {
         var builder = Builder(self)
-        builder[keyPath: keyPath] = newValue
+        callback(&builder)
         return builder.build()
     }
 }
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
@@ -48,7 +48,7 @@ struct AvatarImageModel: Hashable, Identifiable, Sendable {
         return image
     }
 
-    init(id: String, source: Source, state: State = .loaded, isSelected: Bool = false, rating: AvatarRating = .g, altText: String = "") {
+    init(id: String, source: Source, state: State, isSelected: Bool, rating: AvatarRating, altText: String) {
         self.id = id
         self.source = source
         self.state = state
@@ -59,5 +59,11 @@ struct AvatarImageModel: Hashable, Identifiable, Sendable {
 
     func settingStatus(to newStatus: State) -> AvatarImageModel {
         AvatarImageModel(id: id, source: source, state: newStatus, isSelected: isSelected, rating: rating, altText: altText)
+    }
+}
+
+extension AvatarImageModel {
+    static func preview_init(id: String, source: Source, state: State = .loaded, isSelected: Bool = false, rating: AvatarRating = .g) -> Self {
+        AvatarImageModel(id: id, source: source, state: state, isSelected: isSelected, rating: rating, altText: "")
     }
 }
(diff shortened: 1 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

pinarol (approved):
👍

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -207,7 +207,7 @@ class AvatarPickerViewModel: ObservableObject {
 
         let localID = UUID().uuidString
 
-        let localImageModel = AvatarImageModel(id: localID, source: .local(image: image), state: .loading, rating: .g)
+        let localImageModel = AvatarImageModel(id: localID, source: .local(image: image), state: .loading, isSelected: false, rating: .g, altText: "")
```
andrewdmontgomery, line 210: This feels like a much better place to specify the default values.  This is the moment where we actually implement the decision that the default rating (for example) for a newly uploaded image is `.g`.

andrewdmontgomery (approved):
Nice 🚀

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarImageModel.swift
```diff
@@ -57,7 +57,40 @@ struct AvatarImageModel: Hashable, Identifiable, Sendable {
         self.altText = altText
     }
 
-    func settingStatus(to newStatus: State) -> AvatarImageModel {
-        AvatarImageModel(id: id, source: source, state: newStatus, isSelected: isSelected, rating: rating, altText: altText)
+    func updating(_ callback: (inout Builder) -> Void) -> AvatarImageModel {
+        var builder = Builder(self)
+        callback(&builder)
+        return builder.build()
+    }
```
andrewdmontgomery, line 64: The callback syntax at the call site reads a little awkwardly (although it isn't confusing):

`let toggledModel = imageModel.updating { $0.state = state }`

What about using a KeyPath?
```swift
    func updating<Value>(_ keyPath: WritableKeyPath<AvatarImageModel.Builder, Value>, to newValue: Value) -> AvatarImageModel {
        var builder = Builder(self)
        builder[keyPath: keyPath] = newValue
        return builder.build()
    }
```

`let test = imageModel.updating(\.state, to: state)`

etoledom: Oh I like it!

pinarol: When we want to update multiple values at once this is better though:

```
let toggledModel = imageModel.updating { 
   $0.state = state 
   $0.something = something 
   $0.another = another 
}
```

the keypath version would require us to create 3 different instances for this.

pinarol: we can maybe pass a dictionary or something to improve the keypath version so we can update multiple properties at once.

etoledom: We can chain the keypath updates. It doesn't look bad to me 🤔 

```swift
let toggledModel = imageModel.updating(\.state, to: state)
    .updating(\.altText, to: altText)
    .updating(\.rating, to: rating)
```

pinarol: but each `updating()` call creates a new instance from scratch unnecessarily

etoledom: > but each updating() call creates a new instance from scratch unnecessarily

True!
Though structs are very cheap and fast to instantiate.
I like this syntax way more, but I have no problem rolling it back or changing it.

How would it look with a dictionary? Ideally the syntax stays easy to use.

@andrewdmontgomery - any thoughts?

pinarol: I am not saying we have to revert this, just saying we need a way to also update multiple properties at once.

Would it be too bad to have both functions for example?

```
imageModel.updating(\.state, to: state)
```

and 

```
imageModel.updating(
  [\.state: state, 
   \.altText: altText,
   \.rating: rating]
)
```

pinarol: Compared to just having "var" type properties in the image model and letting compiler to do its thing, we are doing explicit initializations for both the struct itself and the Builder to update a property. At this point we can at least try to keep these extra inits minimal I think.

andrewdmontgomery: Good points.  Creating new instances of the `AvatarImageModel` is probably pretty efficient... as long as the model doesn't grow large.  (and I don't know what "large" means).

I thought maybe a variadic function that takes a list of `(KeyPath/Value)` tuples... but `Value` needs to be a single type, which obviously won't work.

I'm thinking we either:
- stick with the callback approach, and live with the need to setup the assignments properly (not hard)
- use the keypath approach and live with the (likely negligible) performance hit of instantiating multiple transient structs

pinarol: > but Value needs to be a single type

oh right, I see...

pinarol: then I would stick with the callback approach

andrewdmontgomery: That works for me.

> Would it be too bad to have both functions for example?

I was wondering that, too.

Issue #539, opened 2024-10-28
Add unit tests for SwiftUI views


pinarol:
Let's check how we can add unit tests for SwiftUI Views. 
 
- AvatarPickerView 
- QuickEditor 
- AvatarGrid 
- Probably some other small Views

We can try snapshotting with the existing snapshot library, like:  

```
func testMyViewSnapshot() {
        // Create the SwiftUI view
        let view = MyView()
        
        let hostingController = UIHostingController(rootView: view)
        hostingController.view.frame = CGRect(x: 0, y: 0, width: 200, height: 100)
        
        assertSnapshot(matching: hostingController, as: .image)
    }
```

pinarol:
Closing this. Let's open more specific issues that state which component.

PR #600, opened 2024-12-13
Add a sample snapshot test for a SwiftUI View

pinarol:
Closes part of https://github.com/Automattic/Gravatar-SDK-iOS/issues/539

### Description

Adds a snapshot test for `AvatarPickerProfileView` to demonstrate how we can do snapshot testing for SwiftUI Views. 

Uses `Snapshotting.image(...)` strategy from the existing snapshot library. It is a snapshot strategy for comparing SwiftUI Views based on pixel equality. There are different `layout` types we can pass, but `.sizeThatFits` didn't work well for this view not sure why. So I just used `.device` and put some red borders around the content so the bounds are also visible and comparable.

### Testing Steps

CI green.


etoledom (commented):
Just a question:
Is there a way to snapshot only de view without the whole bound of the device screen?

pinarol:
Can you check again? @etoledom  I think I found a way to make `.sizeThatFits` work by applying a custom `.frame(...)`

etoledom (approved):
Looks great 👍

PR #603, opened 2024-12-16
Release 3.1.0

pinarol:
Closes #

### Description

Release 3.1.0.

### Testing Steps

Smoke test the new features in the avatar context menu.

### DIFF
--- a/Sources/Gravatar/Resources/SDKInfo.plist
+++ b/Sources/Gravatar/Resources/SDKInfo.plist
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>CFBundleShortVersionString</key>
-	<string>3.1.0-rc.1</string>
+	<string>3.1.0</string>
 </dict>
 </plist>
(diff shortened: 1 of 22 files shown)


pinarol:
After the PR is approved, I'll push the tag and then resolve the conflict and merge.

andrewdmontgomery (approved):
🚀

andrewdmontgomery:
Testing under bad network connections looks good.  The one behavior that I think is awkward is the delete:
1. Load Picker
2. Use Network Conditioner to set `100% packet loss`
3. Delete an avatar
4. Wait for the timeout period

The avatar re-appears and the toast about the error pops up.  That's all good.  It's just that it takes a full minute to error.  It's a weird experience.  I'm not sure the deleting experience is improved with a 60 second retry window.

In the future, we should consider shortening the retry window for deletes substantially.  Delete requests are tiny.  If a bad network connection can't be retried successfully immediately, I think the likelihood that retrying for a full minute will improve the situation.  Instead, I think instant (or near-instant) feedback of the failure is more useful.

PR #608, opened 2024-12-19
Alt text UX changes

etoledom:
Closes #

### Description

Implementing requested changes to the Alt Text editor experience:

- Switch the Save button to a loading indicator while the request is running.
- Keep the alt text screen open until the request returns.
- Close the alt text screen on success (without toast)
- Show an error toast on failure
- Keyboard return button is `Done` button and closes the keyboard.
- No new lines allowed on the Text view.

Not done on this PR:
- Avoid the text from overlapping when text size is big.

CleanShot 2024-12-19 at 11 51 49

CleanShot 2024-12-19 at 11 53 41

### Testing Steps

- Run the Quick Editor screen from the demo app.
- Edit the Alt Text from an avatar.
  - Check that the changes mentioned on description work as expected

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
@@ -23,9 +23,37 @@ struct AltTextEditorView: View {
     let onCancel: () -> Void
 
     var body: some View {
-        VStack {
-            if let email {
-                EmailText(email: email)
+        GeometryReader { geometry in
+            ZStack {
+                ScrollView {
+                    VStack {
+                        if let email {
+                            EmailText(email: email)
+                        }
+                        VStack(alignment: .leading) {
+                            HStack {
+                                titleText
+                                Spacer()
+                                altTextHelpButton
+                            }
+                            ZStack(alignment: .bottomTrailing) {
+                                HStack(alignment: .top) {
+                                    imageView
+                                    altTextField
+                                }
+                                if shouldShowCharCount {
+                                    characterCountText
+                                }
+                            }
+                            Spacer()
+                            actionButton
+                        }
+                        .padding()
+                        .avatarPickerBorder(colorScheme: .light)
+                    }
+                    .padding(.bottom)
+                    .padding(.horizontal)
+                }
             }
             VStack(alignment: .leading) {
                 HStack {
@@ -48,7 +76,6 @@ struct AltTextEditorView: View {
             .avatarPickerBorder(colorScheme: .light)
             Spacer()
         }
-        .padding()
         .gravatarNavigation(
             doneButtonTitle: Localized.cancelButtonTitle,
             actionButtonDisabled: false,
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
@@ -14,6 +14,8 @@ struct AltTextEditorView: View {
     @State var charCount: Int = 0
     @State var safariURL: URL? = nil
     @State var isLoading: Bool = false
+    @ObservedObject var toastManager: ToastManager = .init()
+
 
     @FocusState var focused: Bool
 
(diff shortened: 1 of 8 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
```diff
@@ -2,60 +2,60 @@ import SwiftUI
...
 
     @FocusState var focused: Bool
 
-    let onSave: (AvatarImageModel) -> Void
+    let onSave: (AvatarImageModel) async -> Void
     let onCancel: () -> Void
 
     var body: some View {
         // Scroll view helps detaching the height of the child view from the height of the parent view.
         // This avoids a UI problem while scrolling down the sheet whith the keyboard being present.
```
andrewdmontgomery: Just noticed this small typo (from a previous PR):

```suggestion
        // This avoids a UI problem while scrolling down the sheet with the keyboard being present.
```

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
```diff
@@ -2,60 +2,60 @@ import SwiftUI
...
         // GeometryReader also has the same effect. For now we want the content to scroll when the content grows.
         ScrollView {
-            VStack {
-                if let email {
+            ZStack {
+                VStack {
```
andrewdmontgomery, line 26: ❓ Is this VStack embedded in a ZStack necessary?  I'm guessing it is, but I couldn't immediately see why, so I thought I'd ask.

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -99,24 +99,26 @@ extension View {
...
+                altTextEditor(with: selectedModel).presentationDetents([.height(AltTextEditorView.Constants.sheetHeight)])
             }
         } else {
             return modifier(
-                ModalPresentationModifier(
-                    isPresented: Binding(
-                        get: { model.wrappedValue != nil },
-                        set: { if !$0 { model.wrappedValue = nil }}
-                    ),
+                ModalItemPresentationModifier(
```
andrewdmontgomery, line 119: Building the Demo for iOS 15 doesn't work.  So I can't test this part.  I opened an issue to resolve this: #609

andrewdmontgomery (approved):
This works as described.  I tested with:
- simulated 401 and 403 errors
- simulated poor network conditions
- simulated 100% packet loss

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
etoledom: This ZStack is there for the toast.
In this way it can show over the rest of the content.

True that it's not easy to see since the toast is added at the very end 🤔

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
```diff
@@ -2,60 +2,60 @@ import SwiftUI
...
+                            titleText
+                            Spacer()
+                            altTextHelpButton
                         }
-                        if shouldShowCharCount {
+                        ZStack(alignment: .bottomTrailing) {
+                            HStack(alignment: .top) {
+                                imageView
+                                altTextField
+                            }
```
pinarol, line 38: The character counter intersects with the content. We should place it under the text input not on top of it.

pinarol: In the shared designs the counter font looks smaller. I think we should match this.

image

pinarol: Seems like the counter also should be `.footnote`

Screenshot 2024-12-20 at 12 03 14

Screenshot 2024-12-20 at 12 02 19

Sources/GravatarUI/SwiftUI/View+Additions.swift
pinarol: Good one! Yeah I've been wanting to make that update for a while.
In the previous PR I tested it by manipulating the code. :(

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
```diff
@@ -2,60 +2,60 @@ import SwiftUI
...
-    }
+    let email: Email
 
     @Environment(\.colorScheme) var colorScheme
 
     @State var altText: String = ""
     @State var charCount: Int = 0
     @State var safariURL: URL? = nil
+    @State var isLoading: Bool = false
+    @ObservedObject var toastManager: ToastManager = .init()
```
pinarol: This should not be initialized here(also, you won't get a compile error when you remove this initialization).

We expect that the `ToastManager` is passed here from else where and toasts are added else where. Initializing here makes ToastManager optional since it kinda means "it is OK if the toasts doesn't show".

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
etoledom: > The character counter intersects with the content. We should place it under the text input not on top of it.

I mentioned on the description:

> Not done on this PR:
Avoid the text from overlapping when text size is big.

There's a very easy way to do this on UITextView, which is not supported on SwiftUI, so I plan another PR to tackle this using UITextView in SwiftUI.

etoledom: Counter text size updated 👍

pinarol: > There's a very easy way to do this on UITextView, which is not supported on SwiftUI, so I plan another PR to tackle this using UITextView in SwiftUI.

Can't we just put the counter under the text input? I am not sure if it is worth the complexity of embedding a UIKit component in a SwiftUI View just for this 🤔 But if you think it's easy then go ahead.

pinarol: And even if you wish to use UITextView in the near future, it is best to prevent this overlap in this PR so we don't have a broken experience until that piece of work is done. It is just a matter of putting the counter under the input instead of on top of it.

Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
andrewdmontgomery: Ah, then it's just my rusty SwiftUI. Thanks!

pinarol (approved):
Cool 👍 

I think this is a nice foundation to make more accessibility changes in the future.

PR #610, opened 2024-12-19
Set Swift 6 as the minimum build tools version

andrewdmontgomery:
Closes #510 

### Description

- Updates the Package.swift to require Swift 6 as the minimum toolchain
- Sets each module to use the Swift 5 language
- Updates `.experimentalFeature("StrictConcurrency")` (with the Swift 5 toolchain) to `.enableUpcomingFeature("StrictConcurrency")` (with the Swift 6 toolchain).
- Updates the Demo project so that it explicitly uses the Swift 5 language with SWIFT_STRICT_CONCURRENCY set to `complete`

### Testing Steps

- TBD


pinarol (approved):
Looks good! 🎉

andrewdmontgomery:
hmm... CocoaPods is problematic here.

CocoaPods does not make any distinction between "toolchain" and "language mode".  We can only set SWIFT_VERSION, which is setting the "language mode".

If we leave it at 5.10, then clients with 5.10 will be able to integrate it, and they will get compile errors.

If we change it to 6.0, clients with 5.10 won't be eligible (which is good).  But it applies the full Swift 6 language checks, and we get build errors when we try to use the pod (or in our case, when we lint it during CI).

With CocoaPods, we can't do what we had planned to do.  There is no way to require clients to use the Swift 6.0 toolchain but still use the Swift 5 Language Mode.

We can either:
- Leave the Podspecs on Swift 5.10 (which allows developers to use it in Swift 5.10 even though it doesn't compile)
- Fix all the Swift 6 bugs and stop depending on Swift 5 Language Mode
- Stop supporting CocoaPods entirely

The reasons to stop supporting CocoaPods continue to grow.

andrewdmontgomery:
I am going to revert the Podspec update so CI passes

PR #611, opened 2024-12-20
Avoiding overlapping of alt text with the counter text

etoledom:
Closes #

### Description

This is an attempt to avoid the overlapping of the alt-text string with the counter string.

These screenshot has been taken on an iPhone SE, which offers the smallest width for content.

---
When the dynamic type is normal, a bit bigger, or any smaller; there's no chance for the content to overlap with the counter, so the layout remains the same.

CleanShot 2024-12-20 at 17 17 00@2x
---
For larger dynamic type sizes which are not accessibility sizes, overlapping is super rare, happening when the alt text is long (using close to all characters), and with the last word wrapping right at the end of the last line. 

In this case we move the counter floating down, using most of the space between the text view and the action button. No other part of the layout is affected.

CleanShot 2024-12-20 at 16 55 01@2x
---

For accessibility sizes, overlapping is quite unavoidable, so we move the counter label completely out from the text view, placed in a layout in between the text view and the button. We also increase the text view height to improve readability of the text. This is not the best ui-wise, but is a simple way to avoid the overlapping while keeping the feature usable.

CleanShot 2024-12-20 at 16 55 34@2x

### Testing Steps

- Run the demo app.
- Open the Quick Editor example page.
- Open the alt text editor from any of the avatars.
  - Check that the UI looks as in the screenshots.
- Change dynamic type sizes to bigger ones.
  - Check that the counter won't overlap the text, and they are both always readable

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AltTextEditorView.swift
@@ -5,6 +5,7 @@ struct AltTextEditorView: View {
     let email: Email
 
     @Environment(\.colorScheme) var colorScheme
+    @Environment(\.dynamicTypeSize) var dynamicTypeSize
 
     @State var altText: String = ""
     @State var charCount: Int = 0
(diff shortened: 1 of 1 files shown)


pinarol (approved):
LGTM 👍

PR #614, opened 2024-12-26
Quick Editor: Update the intro screen

### LINKED ISSUE #592: Quick Editor intro screen
1. replace the current initial state(To modify your Gravatar profile, you need to log in first.) with this one:

Image

2. instead of automatically opening the oauth flow, we let the user to open it via “Log in” button when there's no access token.

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/592

gravatarp2 -> native-quick-editor-intro-screen-i2

### Description

Update the QuickEditor Intro screen.

Non-existing user:

<table>

  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

Existing user:

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

### Testing Steps

**Test 1**

Demo app > Quick Editor 
Logout (if needed)
Open the QE
Observe: The new intro screen ☝️ (OAuth should not start immediately)
Observe: There are new fields in the profile card (pronouns, pronounciation)
Close and re-open the QE
Observe: Again, OAuth should not start immediately
Tap: Continue
Observe: OAuth starts
Cancel the OAuth flow
Close and re-open the QE
Observe: OAuth starts automatically this time

**Test 2**

Test the logged in with wrong email error (login with a different email inside OAuth and check if you see the wrong email error)

**Test 3**

Pass the OAuth successfully
Change your selected avatar
Verify that the profile view reflects the change


Sources/Gravatar/BundleInfo.swift
```diff
@@ -1,11 +1,11 @@
 import Foundation
 
-enum BundleInfo {
-    public static var sdkVersion: String? {
+package enum BundleInfo {
+    package static var sdkVersion: String? {
```
pinarol, line 4: This was not supposed to be `public` but somehow overlooked before. 
This is not a breaking change because `BundleInfo` was internal so the variables in it were not publicly reachable.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerProfileViewWrapper.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+import SwiftUI
+
+struct AvatarPickerProfileViewWrapper: View {
```
pinarol, line 4: Extracted out from the `AvatarPickerView`

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -10,13 +10,8 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
     @Environment(\.verticalSizeClass) var verticalSizeClass
     @Environment(\.horizontalSizeClass) var horizontalSizeClass
 
-    // Declare "@StateObject"s as private to prevent setting them from a
-    // memberwise initializer, which can conflict with the storage
-    // management that SwiftUI provides.
-    // https://developer.apple.com/documentation/swiftui/stateobject
-    @StateObject private var model: AvatarPickerViewModel
+    @ObservedObject var model: AvatarPickerViewModel
```
pinarol, line 13: `model` is now owned by the `QuickEditor` and observed here.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -426,6 +413,13 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
     }
 
     func notifyAvatarSelection() {
+        // Trigger the inner avatar refresh
+        model.forceRefreshAvatar = true
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.01) {
+            // Reset the flag with a small delay otherwise the system ignores the value change.
+            model.forceRefreshAvatar = false
```
pinarol, line 409: We need to set `forceRefreshAvatar` to `true` to trigger a force refresh on the avatar when the avatar selection changes. And when we are done, we need to also revert it back to `false`. This made me think we can trigger a network call based on a random taskID or something like that. But needs some changes on the `CachedAsyncImage` and `AvatarView` components and I won't tackle that in this PR.

etoledom: Maybe we can move this to the model.

The model exposes a function `refreshAvatar()` which handles the logic of setting the property.
The taskID sounds like a good idea also, but I agree any of this can be thought about and worked on a separated PR.

etoledom (approved):
Tested all the cases and it's working as expected.

Thank you! 🙏

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
pinarol: All the `notifyAvatarSelection()` calls can probably be moved to the model and unit tested. 🤔 

I thought about the taskID as a replacement for setting the flag to `true` and then to `false` consequently but I haven't thought much deeply about it. I might come up with something totally different when I dive into it...

Anyway yeah, belongs to a separate PR.

PR #615, opened 2024-12-27
Extra profile refresh for the new users

### LINKED ISSUE #612: Profile needs refreshing after the first avatar fetch on a new account
I noticed this when testing the Jetpack app
When a new user is created on the Jetpack app they don't have a Gravatar profile yet
The Gravatar profile is created first time the user fetches their avatars, even if the list is empty
But the profile view remains blank in this case
Let's refresh the profile view so the profile view doesn't remain blank

pinarol:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/612

### Description

Fixes the issue described in https://github.com/Automattic/Gravatar-SDK-iOS/issues/612
This is to avoid the initial blank state of the profile view that a new user sees. When a user creates an account via `WordPress.com` they don't have a Gravatar profile yet so the Quick Editor shows blank fields initially. The Gravatar profile is created at the backend during the first avatars fetching, but we never refresh the client so the profile view remains blank and shows grey placeholders. This PR fixes that issue. 

The issue is not too bad so I was on the fence of fixing/leaving it. Doing this only because I believe that the solution is not risky, it only adds an extra profile fetch call. But let me know if you think otherwise. 

### Testing Steps

Easiest way to test is using the Jetpack app

Check out WordPress-iOS `trunk`
Point the Gravatar dependency to this branch' latest commit
Run the Jetpack app
You can use a temp inbox like [www.emailondeck.com](http://www.emailondeck.com/) to create an email
Logout and create a new user
Go to your inbox, verify the email (open the email verification link in incognito mode to isolate from your existing .com session otherwise it may not work)
Go to Me > My Profile > Add profile photo
Observe: The profile view does NOT show grey placeholder fields, instead, it shows the user's username and "View profile ->" button.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
@@ -38,6 +38,7 @@ class AvatarPickerViewModel: ObservableObject {
     @Published var shouldDisplayNoSelectedAvatarWarning: Bool = false
     @ObservedObject var toastManager: ToastManager = .init()
     private var cancellables = Set<AnyCancellable>()
+    private(set) var compensatingFetchProfileTask: Task<Void, Never>? // for unit testing
 
     init(
         email: Email,
@@ -114,7 +115,7 @@ class AvatarPickerViewModel: ObservableObject {
 
         $profileResult
             .combineLatest($gridResponseStatus)
-            .filter { profileResult, gridResponseStatus in
+            .map { profileResult, gridResponseStatus in
                 let isProfileStatus404 = switch profileResult {
                 case .failure(APIError.responseError(let .invalidHTTPStatusCode(response, _))) where response.statusCode == HTTPStatus.notFound.rawValue:
                     true
@@ -128,13 +129,15 @@ class AvatarPickerViewModel: ObservableObject {
                 default:
                     false
                 }
-
                 return isProfileStatus404 && isGridResponseSuccess
             }
-            .sink { [weak self] _, _ in
-                Task {
-                    // Profile does not exists but `/v3/me/avatars` is success. This means it's a new account. Backend creates a profile
-                    // during the first GET `/v3/me/avatars` of a new user. So we refresh the profile to fetch it.
+            .filter { $0 == true }
+            .removeDuplicates()
+            .sink { [weak self] _ in
+                self?.compensatingFetchProfileTask = Task {
+                    // Profile does not exists but `/v3/me/avatars` is success. This means it's a new account. Backend creates a new
+                    // profile during the first GET `/v3/me/avatars` of a new user. So we refresh the profile to fetch it.
+                    // Happens only when the token is passed externally.
                     await self?.fetchProfile()
                 }
             }
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
@@ -274,6 +274,14 @@ class AvatarPickerViewModel: ObservableObject {
                 grid.selectAvatar(withID: avatar.id)
                 self.selectedAvatarURL = URL(string: avatar.url)
                 self.backendSelectedAvatarURL = URL(string: avatar.url)
+                switch profileResult {
+                case .failure:
+                    // Serverside creates a new profile on the first successful
+                    // avatar upload so we refresh the profile.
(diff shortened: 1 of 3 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -115,6 +99,52 @@ class AvatarPickerViewModel: ObservableObject {
...
+                }
+                return isProfileStatus404 && isGridResponseSuccess
+            }
+            .filter { $0 == true }
+            .removeDuplicates()
+            .sink { [weak self] _ in
+                self?.compensatingFetchProfileTask = Task {
+                    // Profile does not exists but `/v3/me/avatars` is success. This means it's a new account. Backend creates a new
+                    // profile during the first GET `/v3/me/avatars` of a new user. So we refresh the profile to fetch it.
+                    // Happens only when the token is passed externally.
```
andrewdmontgomery: Small typo:
```suggestion
                    // Profile does not exist but `/v3/me/avatars` is success. This means it's a new account. Backend creates a new
                    // profile during the first GET `/v3/me/avatars` of a new user. So we refresh the profile to fetch it.
                    // Happens only when the token is passed externally.
```

andrewdmontgomery:
This looks like a reasonable workaround to me.

At first blush, at least, this seems like something the backend should handle better.  It seems strange that the backend lazily creates the profile when `/v3/me/avatars` is fetched – but not when the actual profile is fetched.  I wonder why?  It seems like a backend error if the backend returns a 404 for a profile of a valid Gravatar account, regardless of how recently the account was created.

pinarol:
I thought it was because the `/profiles/profile_id` is an unauthenticated endpoint, but no, we add the access token to the request so there should not be an authorization issue. Let me check this internally, maybe we won't need this change at all.

pinarol:
Discussed with @jom, the `/v3/profiles/...` doesn't have an authenticated version atm. We add the token but it's ignored.

etoledom (commented):
I have reproduced the issue and tested this solution.

I think a small fix for this backend issue is good to have. In my experience, though it's not a blocker for the user, it might appear odd for some, but probably many won't really pay attention to it.

Still it seems like a malfunction by our SDK so I'd vote for merging this fix.

This is how the issue looks like:

https://github.com/user-attachments/assets/579e9a40-9e8b-46b2-ba69-9b93aa985ee7



I also tested setting an image to see if the whole Profile card updates (magically). If it were the case, this fix wouldn't be needed IMO, but it does not refresh, only the avatar.

https://github.com/user-attachments/assets/1f421b99-5e40-4574-9400-bcf4c1111df7

Finally, while testing this branch, the Profile card loads as expected 🎉 

https://github.com/user-attachments/assets/361d9316-deb0-4578-b002-42642ba26c6b

I saw first a small blink where the card shows the placeholder labels:


To then finally load the user info:



But most importantly, what I noticed is that the avatar from the Profile won't change after selecting an avatar (shown in the last video).

I have two thoughts about this:

- If the Profile Card loads the placeholder text in the labels instead of keeping the gray loading state, maybe is much less obvious that it is an error, and we can keep it this way.

- The avatar selection no showing on the Profile card is a blocker to merge this PR. But I'd vote for a simpler solution if possible which I mentioned in the previous point.

etoledom:
After restarting Jetpack, the problem persisted, but I continued changing and testing and it eventually started showing the change on the card:



Not sure if the problem is on the SDK, backend, or even Jetpack cache.

pinarol:
> But most importantly, what I noticed is that the avatar from the Profile won't change after selecting an avatar (shown in the last video).

I noticed the same issue and I think fixed it by increasing the delay here in the release branch. https://github.com/Automattic/Gravatar-SDK-iOS/blob/d82936aca08ee1ea98306fd88cd509fea82980cd/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift#L408 I'll rebase this on it. Not ideal though, I should make an enhancement on the AvatarView so we don't need this workaround anymore.

pinarol:
I made this PR https://github.com/Automattic/Gravatar-SDK-iOS/pull/644 to remove the workaround I mentioned here https://github.com/Automattic/Gravatar-SDK-iOS/pull/615#issuecomment-2602486453

etoledom (approved):
The problem mentioned before seems to be resolved:

https://github.com/user-attachments/assets/69847efb-1bff-4abf-8c45-5f09451d3d7f

The user profile is loaded on the first opening of the quick editor, and the avatar image updates as expected 🎉

PR #623, opened 2025-01-07
Fix "Multiple images are selected when the same image is uploaded"

### LINKED ISSUE #618: Multiple images are selected when the same image is uploaded
I've come across with this nasty bug that causes multiple images to become selected:

Steps:

Open the QE
Upload the "same" image multiple times
Select the image
Result: All the uploaded images are selected at the same time, the loading indicator is also on the wrong one.
Expected: Only the image that's tapped on should be selected.

pinarol:
Closes #618

### Description

Backend generates avatar ids based on the image file's hash. When we upload the same image multiple times it has the same id (not always though, I can sometimes upload the same image twice because a different hash is generated somehow). So we should remove the other duplicate one. This also matches behavior with android and web.

### Testing Steps

Upload the same image multiple times
Select it
Only 1 image should be selected

### DIFF
--- a/Sources/Gravatar/Resources/SDKInfo.plist
+++ b/Sources/Gravatar/Resources/SDKInfo.plist
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>CFBundleShortVersionString</key>
-	<string>3.1.0</string>
+	<string>3.1.1</string>
 </dict>
 </plist>
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarGridModel.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarGridModel.swift
@@ -25,6 +25,16 @@ class AvatarGridModel: ObservableObject {
     func replaceModel(withID id: String, with model: AvatarImageModel) {
         guard let index = index(of: id) else { return }
         avatars[index] = model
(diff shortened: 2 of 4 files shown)


pinarol:
I was tempted to use `OrderedSet<Element>` for the grid items to ensure that each element appears only once(based on its id), but it comes with a [separate package](https://github.com/apple/swift-collections) and I am not sure if this is worth it.

etoledom (approved):
This solves the problem, and as a fast patch of an ugly bug, I think we can go with this.

Some things I'd improve in a next release:
- A small animation when deleting the repeated image.
- A note saying something like "This avatar has already been uploaded before".
- Maybe instead of deleting instantly, we can keep it with an error/warning icon, with the delete option.

Probably we can revisit this with a designer later on.

But for now let's fix the main problem 👍 
Thank you for taking care of this 🙏

pinarol:
> Maybe instead of deleting instantly, we can keep it with an error/warning icon, with the delete option.

I would find this confusing as a user because I didn't even interact with the other image. 😬

Showing a toast can be an option but there's an inconsistency problem unfortunately, and the toast can make it more visible. It sometimes allows me to upload duplicate images(it must be generating a different hash somehow, but it's indifferentiable by eye) so a user might find it buggy when we allow duplicate images sometimes but not the other times... There's no good explanation for this from a user's point of view. So not pulling the attention here might be better. 🤷‍♀️

PR #626, opened 2025-01-07
Initialize fetchedToken with the keychain item

### LINKED ISSUE #625: OAuth doesn't start after a fresh install combined with token expiration
Have the Quick Editor logged in
From the WP.com security panel, revoke the demo app token
Delete the app and build-run again
Go to the Quick Editor and log in with the same email than before
It will show the intro message, tap continue
Result: Switches to the token expired state
Expected: Starts the OAuth flow

pinarol:
Closes #625 #579

### Description

Removing the app doesn’t clear the keychain entry for that email so oauthSession.hasValidSession(with: email) returns true which causes it to skip the oauth flow.

The source of truth about tokens is the Keychain but SwiftUI uses `@State` variables to shape and refresh the UI. `fetchedToken` needs to be initialized from the keychain so the UI can adapt accordingly.

### Testing Steps

Refer to #625 and #579 for testing steps.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
+++ b/Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
@@ -71,7 +71,11 @@ struct QuickEditor<ImageEditor: ImageEditorView>: View {
             } else {
                 noticeView()
             }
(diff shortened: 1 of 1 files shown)


etoledom (approved):
Great find!
I tested both tickets and they seem fixed by this PR.

🎉

PR #629, opened 2025-01-09
Set demo app min iOS deployment target to 15.0

### LINKED ISSUE #609: The Demo app minimum iOS target should match our Swift Package minimum supported platforms.
Our Demo app minimum target is iOS 17.  But our SDK supports iOS 15.  We should update the Demo app target to match the SDK, so we can build the Demo in all supported OS versions.

Right now, the demo app cannot be built in iOS 15.

pinarol:
Closes #609

### Description

Sets demo app iOS deployment target to 15.0 and updates the Demo app's code accordingly. SDK code doesn't change.

### Testing Steps

Smoke test demo app QE in both SwiftUI and UIKit.

I checked in iOS 15.5 and 16.4 as well. It takes a bit long to download older iOS environments though so if you don't already have those, you can do that kind of testing later imo.


etoledom (approved):
Thank you! 🙏

PR #631, opened 2025-01-10
Update podspecs to Swift 6.0

andrewdmontgomery:
Closes #

### Description

This updates the podspecs so that they specify Swift 6.0 as a minimum supported version.  This mirrors the changes in #610  and should have been included in that PR.

### Testing Steps

- [ ] CI should be 🟢

### DIFF
--- a/version.rb
+++ b/version.rb
@@ -3,6 +3,6 @@
 module Gravatar
   VERSION = '3.1.1'
   SWIFT_VERSIONS = [
-    '5.10'
+    '6.0'
   ].freeze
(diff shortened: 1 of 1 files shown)


andrewdmontgomery:
CocoaPods doesn't have a way to handle the configuration we need it to handle:
- Only available in the Swift 6 toolchain, but using the Swift 5 Language Mode

CocoaPods can only set `SWIFT_VERSION` build setting, which controls the "Swift Language Mode", not the build tools.

PR #622, opened 2025-01-06
Reintroduce SwiftLint

andrewdmontgomery:
Closes #

PR #634, opened 2025-01-14
Moving SwiftLint and SwiftFormat to Docker

### LINKED ISSUE #636: Remove SwiftFormat as a dependency of the Package

andrewdmontgomery:
Closes #636 
Closes #622 

### Description

This removes SwiftFormat as a dependency of the Package.swift.  It also re-introduces SwiftLint linting.

It accomplisheses these objects using the same technique.  In both cases, we now use Docker images to run these:
- The Makefile tasks have been updated to use Docker
- CI has been updated to use Docker
   - SwiftLint actually uses a custom `linter` agent which bundles SwiftLint using Docker

### Updating the "Required" Status Checks after merging
Back when SwiftLint ran at build time, we didn't need a dedicated `SwiftLint` step, since we were already building the package.

Now that it needs to be called explicitly, we have a new required step in our CI builds.  This will mean updating our "required" checks:

1. When this PR is ready to merge, we will update the "required" checks:
   - Remove the old `buildkite/gravatar-sdk-ios/lint` required check
   - Add the new `SwiftLint` and `SwiftFormat linting` checks as required
2. After this PR has merged to `trunk`, all branches with PR's targeting `trunk` will need to merge this change in.  Until they do, they won't be able to run these updated "required" checks, and they won't be mergable. 

### Making GitHub Checks more readable
I've taken this opportunity to use `notify` blocks in these new steps for SwiftLint and SwiftFormat:

```yaml
    steps:
      - label: ":swift: SwiftLint"
        command: swiftlint
        notify:
          - github_commit_status:
              context: "SwiftLint"
        agents:
          queue: "linter"
```

This produces a much cleaner, more readable name for these checks in the GitHub Checks section of a PR:
Screenshot 2025-01-15 at 4 28 00 PM

In a future PR, we can use the same technique to replace the rest of the `buildkite/gravatar-sdk-ios/*` checks with more readable names.

### Limitations
In the past, SwiftLint ran at build time.  This produced linting errors in line with Xcode.  It may be possible to re-introduce this capability in the future.  But it will be tricky.  The Xcode build phase script would need a guarantee that it could find the `docker` binary, etc.

### pre-commit hook
In a separate PR, we can explore a pre-commit hook that runs both SwiftLint and SwiftFormat.

### Testing Steps
You will need Docker installed and running on your Mac for these steps:

- [ ] CI won't be gree until we update the required checks (see above)
- [ ] CI should pass the following new checks:
   - SwiftLint
   - SwiftFormat Linting
- Test `make swiftlint`
   1. Add an `NSLocalizedString()` in one of the `Gravatar` or `GravatarUI` targets
   2. Run `make swiftlint`
      - [x] OBSERVE that the first time you run this, docker downloads the required image
      - [x] OBSERVE that subsequent invocations don't download the image again
      - [x] OBSERVE that SwiftLint throws an error for you `NSLocalizedString()`
   3. Remove your change and run `make swiftlint` again
      - [x] OBSERVE that SwiftLint runs cleanly
- Test `make swiftformat`
   - Make a formatting change to a source file in `Gravatar` or `GravatarUI`.  For example, find an `if/else` block, and move the brackets:
```diff
if something {
    // do something
-} else {
+}
+else {
    // do something else
}
```
   - Run `make swiftformat`
   - [x] OBSERVE that your formatting change is fixed
- Test `make swiftformat-lint`
   - Repeat the steps above, only use the command `make swiftformat-lint`
   - [x] OBSERVE that the command throws an error due to your formatting change:
```
ImageUploadService.swift:24:1: error: (elseOnSameLine) Place else, catch or while keyword in accordance with current style (same or next line).
```

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -21,18 +21,18 @@ steps:
     agents:
       queue: "linter"
 
-  # - label: ":swift: SwiftFormat Linting"
-  #   plugins:
-  #     - docker#v5.12.0:
-  #         image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
-  #         command: ["--lint", "Sources"]
-  #         workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
-  #     - docker#v5.12.0:
-  #         image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
-  #         command: ["--lint", "Tests"]
-  #         workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
-  #   agents:
-  #     queue: "default"
+  - label: ":swift: SwiftFormat Linting"
+    plugins:
+      - docker#v5.12.0:
+          image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
+          command: ["--lint", "Sources"]
+          workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
+      - docker#v5.12.0:
+          image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
+          command: ["--lint", "Tests"]
+          workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
+    agents:
+      queue: "default"
 
   - label: ☢️ Danger - PR Check
     command: danger
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -18,8 +18,6 @@ steps:
     plugins:
       - docker#v5.12.0:
           image: "ghcr.io/realm/swiftlint:$SWIFTLINT_VERSION"
-          volumes:
-            - "${BUILDKITE_BUILD_CHECKOUT_PATH}:${BUILDKITE_BUILD_CHECKOUT_PATH}"
           workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
     agents:
       queue: "default"
@@ -29,14 +27,10 @@ steps:
       - docker#v5.12.0:
           image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
           command: ["--lint", "Sources"]
-          volumes:
-            - "${BUILDKITE_BUILD_CHECKOUT_PATH}:${BUILDKITE_BUILD_CHECKOUT_PATH}"
           workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
       - docker#v5.12.0:
           image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
           command: ["--lint", "Tests"]
-          volumes:
-            - "${BUILDKITE_BUILD_CHECKOUT_PATH}:${BUILDKITE_BUILD_CHECKOUT_PATH}"
           workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
     agents:
       queue: "default"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -33,6 +33,9 @@ steps:
               image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
               command: ["--lint", "Tests"]
               workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
+        notify:
+          - github_commit_status:
+              context: "SwiftFormat Linting"
         agents:
           queue: "default"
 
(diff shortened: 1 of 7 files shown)


.swiftlint.yml
AliSoftware, line 1: @andrewdmontgomery you will need to explicitly add a `swiftlint_version: …` key somewhere in this `.swiftlint.yml` config file, for CI to find and know which version of SwiftLint's docker image it should use.

.buildkite/pipeline.yml
```diff
@@ -7,26 +7,47 @@ agents:
...
+              workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
+          - docker#v5.12.0:
+              image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
+              command: ["--lint", "Tests"]
+              workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
+        notify:
+          - github_commit_status:
+              context: "SwiftFormat Linting"
+        agents:
+          queue: "default"
```
AliSoftware, line 40: @iangmaia 💡  would it be worth it to make the linter agent support `swiftformat` at some point too? 🤔

iangmaia: Indeed. I don't recall other of our repos using it but perhaps more people would be interested 👍

.swiftlint.yml
andrewdmontgomery: Nice, thanks!

.buildkite/pipeline.yml
andrewdmontgomery: We like it quite a bit.  That said, [swiftlang/swift-format](https://github.com/swiftlang/swift-format) is embedded within Xcode now.  I haven't looked at swapping to that yet.  But it is intriguing.

andrewdmontgomery: I moved the source of truth for the SwiftFormat version into its `.swiftformat` config, and used the same technique for reading it from the Makefile and pipeline.

andrewdmontgomery:
@pinarol @etoledom 
This is targeting `trunk`.  Do we want to merge this into `release/3.2.0`?

It doesn't really impact the SDK.  So it doesn't need to merge into a "minor" release, necessarily.  But the removal of a dependency in the `Package.swift` is a nice improvement.

AliSoftware:
> In a future PR, we can use the same technique to replace the rest of the buildkite/gravatar-sdk-ios/* checks with more readable names.

Let us (AppsInfra) know when you get there, since after you've added explicit `notify` to all the steps of your pipeline we could then disable the implicit check names that Buildkite auto-generates (to avoid having each of the steps reported twice, implicit + explicit)

andrewdmontgomery:
Very nice. Will do.

.buildkite/pipeline.yml
```diff
@@ -7,26 +7,47 @@ agents:
...
+      - label: ":swift: SwiftFormat Linting"
+        plugins:
+          - docker#v5.12.0:
+              image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
+              command: ["--lint", "Sources"]
+              workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
+          - docker#v5.12.0:
+              image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
+              command: ["--lint", "Tests"]
+              workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
```
AliSoftware, line 35: TIL you can have 2 `docker` plugins entries on the same step (and Buildkite will just execute them one after the other in that single step / CI job) — as opposed to have 2 separate steps each invoking one docker plugin instance each. Nice 👍

AliSoftware (commented):
Just looking at the code diff in GitHub from my iPad, changes look good to me 👍

I'll just comment (instead of approve) though, to let someone else actually run the test steps described in the PR description and properly go through the different test cases on their Mac before approving the PR proper 🙂

pinarol:
> @pinarol @etoledom
> This is targeting trunk. Do we want to merge this into release/3.2.0?

> It doesn't really impact the SDK. So it doesn't need to merge into a "minor" release, necessarily. But the removal of a dependency in the Package.swift is a nice improvement.

Seems like a safe change, yeah why not?

pinarol (approved):
Works great 🎉 Thank you for this!

.buildkite/pipeline.yml
andrewdmontgomery: That was me 🤞 and hoping it worked 😂

andrewdmontgomery:
> > @pinarol @etoledom
> > This is targeting trunk. Do we want to merge this into release/3.2.0?
> 
> > It doesn't really impact the SDK. So it doesn't need to merge into a "minor" release, necessarily. But the removal of a dependency in the Package.swift is a nice improvement.
> 
> Seems like a safe change, yeah why not?

Nice, agreed.  I'll retarget the release branch.  Once I merge it into `release`, I'll need to setup a back merge from `release --> trunk`.

Then, before _that_ PR can merge, I'll need to reconfigure the "required" checks (per the description)

PR #639, opened 2025-01-16
Add docker check

andrewdmontgomery:
Closes #

### Description

This fixes a few minor bugs with the new docker implementations of SwiftLint and SwiftFormat:
- Ensures that all `docker run` commands use the `--rm` flag, which removes the container when the run stops
- Checks for a functional docker installation before using docker, throws a descriptive error if there is a problem with the Docker installation

#### Check that all docker containers are removed when they stop
1. Ensure that Docker is running locally
2. Run `docker ps -aq | wc -l`
   - This will return the number of docker containers (running or stopped)
3. Run `make lint` multiple times
4. Run `docker ps -aq | wc -l` again
   - [ ] OBSERVE: the number of containers has not changed

#### Docker is installed but not running
1. Stop docker on your workstation
6. Run `docker info` to ensure that docker is not running
   - `ERROR: Cannot connect to the Docker daemon...`
7. Run `make lint`
   - [ ] OBSERVE: lint generates the following error:
```shell
$ make lint
Error: Docker is installed but not running or accessible by the current user
make[1]: *** [check-docker] Error 1
make: *** [lint] Error 2
```

#### Docker is not in PATH (simulating not installed)
1. Stop docker on your workstation
2.  Run `docker info` to ensure that docker is not running
   - `ERROR: Cannot connect to the Docker daemon...`
3. Find the `docker` binary and temporarily move it out of your path
```shell
# Find the path to the docker binary
$ which docker
/usr/local/bin/docker

# Rename the docker binary (may require sudo)
$ sudo mv /usr/local/bin/docker /usr/local/bin/docker-original
```

8. Run `make init`
   - [ ] OBSERVE: lint generates a different error:
```shell
$ make lint
Error: Docker is not installed or not in PATH
make[1]: *** [check-docker] Error 1
make: *** [lint] Error 2
```

9. Restore the docker binary
```shell
$ sudo mv /usr/local/bin/docker-original /usr/local/bin/docker
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,24 +68,24 @@ fetch-code-signing: bundle-install
 setup-secrets: bundle-install
 	bundle exec fastlane run configure_apply
 
-swiftformat: # Automatically find and fixes lint issues
+swiftformat: check-docker # Automatically find and fixes lint issues
 	@docker run --rm -v $(shell pwd):$(shell pwd) -w $(shell pwd) ghcr.io/nicklockwood/swiftformat:$(SWIFTFORMAT_VERSION) Sources
 	@docker run --rm -v $(shell pwd):$(shell pwd) -w $(shell pwd) ghcr.io/nicklockwood/swiftformat:$(SWIFTFORMAT_VERSION) Tests
 
-swiftformat-lint:
+swiftformat-lint: check-docker
 	@docker run --rm -v $(shell pwd):$(shell pwd) -w $(shell pwd) ghcr.io/nicklockwood/swiftformat:$(SWIFTFORMAT_VERSION) --lint Sources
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -115,11 +115,7 @@ update-example-snapshots:
 	cd ./Sources/GravatarUI/GravatarUI.docc/Resources/ProfileExamples && \
 	for filePath in *; do name=$${filePath%.*}; mv $$filePath $${name//-dark/~dark}@2x$${filePath#$$name}; done
 
-install-and-generate: $(OPENAPI_GENERATOR_CLONE_DIR) # Clones and setup the openapi-generator.
-	"$(OPENAPI_GENERATOR_CLONE_DIR)"/run-in-docker.sh mvn package
-	make generate
```
andrewdmontgomery, line 120: This should have been removed when I refactored the way we generate OpenAPI models.  Since we're using a docker container now, there is nothing to install.

pinarol (approved):
👍

PR #641, opened 2025-01-20
Remove unused lane that duplicated existing one

mokagio:
### Description

I landed here while regenerating the Enterprise provisioning profiles after creating a new distribution certificate. Internal reference pdnsEh-1ZG-p2

I was looking at the code and noticed a lane that looked unused (no `git grep` matches`) and that had the same implementation of another one, which is instead in use in the codebase.

### Testing Steps

N.A. This only removes a duplicated lane.

Having said that, a successful prototype build will show that the code that was touched is still working.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -120,15 +120,9 @@ platform :ios do
     #
     # Which suggest the authenticated session from the first call was not carried over to the second.
     configure_code_signing(
-      bundle_ids: [
-        bundle_id_prototype_build!
-      ],
+      bundle_ids: [bundle_id_prototype_build!],
       readonly: readonly
     )
-  end
-
-  lane :configure_code_signing_prototype_build do |readonly: true|
-    configure_code_signing(bundle_ids: [bundle_id_prototype_build!], readonly: readonly)
   end
 end
 


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/Fastfile
```diff
@@ -120,16 +120,10 @@ platform :ios do
...
-        bundle_id_prototype_build!
-      ],
+      bundle_ids: [bundle_id_prototype_build!],
       readonly: readonly
     )
   end
-
-  lane :configure_code_signing_prototype_build do |readonly: true|
-    configure_code_signing(bundle_ids: [bundle_id_prototype_build!], readonly: readonly)
-  end
```
mokagio, line 132: This was exactly the same call as `:configure_code_signing` above.

andrewdmontgomery: Nice find.   I'll just give it a 👍 here, since someone from your team should review (as you requested).

iangmaia (approved):
👍

PR #644, opened 2025-01-22
Add one time force refresh ability for CachedAsyncImage and AvatarView

pinarol:
Closes Described here https://github.com/Automattic/Gravatar-SDK-iOS/pull/615#pullrequestreview-2562244320 (see "it does not refresh" part)

### Description

We need to force refresh the avatar just once when the avatar selection changes. Previously I did that by setting forceRefresh to `true` and then `false` with a little delay but sometimes the underlying system can ignore the `true` value somehow. So I added a new binding just for *one time* refreshing purpose, which is actually the real need here. This way we can set the flag back to `false` after the call is really done.

### Testing Steps

- Switch between avatars many times or delete the selected avatar
- Make sure the avatar in the QE's profile card gets updated

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerProfileView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerProfileView.swift
@@ -106,7 +106,7 @@ struct AvatarPickerProfileView: View {
                     .colorScheme(colorScheme)
                     .background(Color(UIColor.systemBackground))
             },
-            forceRefresh: $forceRefreshAvatar,
+            oneTimeForceRefresh: $forceRefreshAvatar,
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

andrewdmontgomery (approved):
This approach looks better.  Nice.  Testing seems to work great, too.  I ran into a bug, but it's happening on the `release/3.2` branch also.  And to some degree, it also happens on the `release/3.1.1` commit.  I'll create an issue.

PR #647, opened 2025-01-23
Adds `rating: .x` to the avatar URL in the QE's profile card and some demo screens

### LINKED ISSUE #646: GravatarUI elemets that show avatars only load `.g` rated avatars
When a Gravatar account uses an avatar with a rating above `G`, our GravatarUI view that shows avatars load the default avatar instead.  This is because the default AvatarService query doesn't specify a rating, and the default is `G`.

pinarol:
Closes #646

### Description

Adds `rating: .x` to the avatar URL in the QE's profile card and some demo screens. This is to successfully demonstrate avatar changes otherwise the avatar endpoint returns 404 when the rating changes to something lower than "general".

### Testing Steps

Change ratings of avatars
Select those avatars
Observe: The avatar in the QE's profile card and the demo app reflects the change properly

### DIFF
--- a/Demo/Demo/Gravatar-Demo/DemoQuickEditorViewController.swift
+++ b/Demo/Demo/Gravatar-Demo/DemoQuickEditorViewController.swift
@@ -197,7 +197,7 @@ final class DemoQuickEditorViewController: UIViewController {
             token: token
         )
         presenter.present(in: self, onAvatarUpdated: { [weak self] in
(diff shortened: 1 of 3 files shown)


etoledom (approved):
I was able to reproduce the issue, and this solution solves it 👍 

CleanShot 2025-01-23 at 13 27 04

PR #652, opened 2025-01-24
Fix inconsistencies with android

etoledom:
Closes #650
Closes #649 

### Description

On this PR we we fixe some inconsistencies with the design and Android implementation:

### Testing Steps

- Run the demo app -> Quick Editor demo screen
- Open the quick editor with a new email (or log-out if needed)
  - When the intro screen show, check that the email is not showing.
- Log in with an account.
- Go back and try to login again with a different email (without login out previously) 
  - When the wrong email error screen shows, check that the email is visible.

PR #653, opened 2025-01-24
Leaving only location on QE profile card details

### LINKED ISSUE #651: Show only location in the QE's profile card
We should only show the location in the profile card, and the placeholder text should be just "Location"

Image

etoledom:
Closes #651 

### Description

Removing extra info from the QE profile card, leaving only the Location.

### Testing Steps

- Run the demo app
- Open the Quick Editor demo
- Open the quick editor.
  - Check that the profile card shows only the location for both the placeholder (no account for given email), and for the actual profile info card (with an existing account).

Check the updated snapshots for refetence.


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerProfileView.swift
```diff
@@ -11,14 +11,9 @@ struct AvatarPickerProfileView: View {
...
         var profileURL: URL?
-        var pronunciation: String
-        var pronouns: String
 
         var profileDetails: String? {
-            let joinedFields = [pronunciation, pronouns, location]
-                .filter { !$0.isEmpty }
-                .joined(separator: "・")
-            return joinedFields.isEmpty ? nil : joinedFields
+            location
```
pinarol: We need to return `nil` if the field is empty text. Otherwise a space occurs there. Seems like we don't have a snapshot test cover this unfortunately, would be good to add that also.

andrewdmontgomery: Looking around for other instances of this, wondering if this is something that would be worth addressing more consistently, and I found an extension for this in PersonalInfoModel (in GravatarUI).  We could move it into its own source file (for clarity) and reuse it here:

```swift
extension String {
    fileprivate func nilIfEmpty() -> String? {
        isEmpty ? nil : self
    }
}
```

etoledom: Good catch!

I used `nilIfEmpty() `, and added an extra unit test to catch this issue next time on tests.

etoledom:
Ready for another look 👀

pinarol (approved):
Nice work! 🎉 :shipit:

pinarol:
Wait, @etoledom you should target release/3.2.0 branch.

pinarol:
Same for https://github.com/Automattic/Gravatar-SDK-iOS/pull/652 :(

pinarol:
I'll merge this to `trunk` as well and update the release branch from trunk.

PR #655, opened 2025-01-24
Update localizations

andrewdmontgomery:
Closes #

### Description

Update localizations for release 3.2.0


pinarol (approved):
This looks good but we have 1 more localized string we added last week, so let's keep the issue open after merging this.

PR #659, opened 2025-02-07
Add App Clip to Demo project

mokagio:
Internal ref pdnsEh-21E-p2

### Description

- Adds an App Clip to the Demo project
- Wires up code signing for the Debug configuration — More work required for the Release build, given it can only be tested on the App Store account and we don't have it set up yet
- Updates the Prototype Build automation to strip the App Clip. The Prototype Build is distributed via the Enterprise account but Enterprise apps are not allowed App Clips.

### Testing Steps

Checkout this branch and run the debug build on device.

Here's a screenshot from mine:

IMG_6F3CC09BE35F-1

### Where to go from here

For @andrewdmontgomery @pinarol & team to

- Implement the actual App Clip. What's here is just the Xcode template
- Let me / `@Automattic/apps-infrastructure` know whether to set up TestFlight distribution to test the App Clip without having to deploy the app on device. That will require a bunch of Fastlane updates, so I didn't go down that path in the context of this PR.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/Fastfile
```diff
@@ -213,3 +217,32 @@ def read_from_xcconfig!(key:, xcconfig:)
...
+    app.build_phases.delete_if { |b| b.display_name == 'Embed App Clips' }
+  end
+
+  UI.message('Saving the project file....')
+  project.save
+
+  UI.success('Dependency successfuly removed.')
+end
+# rubocop:enable Metrics/AbcSize
+# rubocop:enable Metrics/CyclomaticComplexity
```
mokagio, line 248: This is duplicated from Pocket Casts. We might want to centralize at the release toolkit level, see https://github.com/wordpress-mobile/release-toolkit/issues/629

pinarol:
Thank you very much! This is very useful for us. I won't merge it for now as this effort might move to a separate repo.

pinarol:
@mokagio The app clip target works fine but this time, I can't get the old demo app target to build in this branch. I get the error: `The CFBundleVersion of an App Clip ('1') must match that of its containing parent app ('0').`

pinarol:
Also, we need to test OAuth in the app clip target so we'll need this config. Does this require updating the provisioning profile also? @mokagio 

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>com.apple.developer.associated-domains</key>
	<array>
		<string>appclips:gravatar.com</string>
		<string>applinks:gravatar.com</string>
		<string>webcredentials:gravatar.com</string>
	</array>
	<key>com.apple.developer.parent-application-identifiers</key>
	<array>
		<string>$(AppIdentifierPrefix)com.automattic.gravatar-sdk-demo-uikit</string>
	</array>
</dict>
</plist>

```

andrewdmontgomery:
> @mokagio The app clip target works fine but this time, I can't get the old demo app target to build in this branch. I get the error: `The CFBundleVersion of an App Clip ('1') must match that of its containing parent app ('0').`

I fixed this. [[5e5a614](https://github.com/Automattic/Gravatar-SDK-iOS/pull/659/commits/5e5a61439b4e48868b5e8433e1508ed39841d098)]

fastlane/Fastfile
andrewdmontgomery: Oh right, the original development profile is for an Enterprise app, which doesn't support App Clips.  Thanks for this.

andrewdmontgomery:
> Does this require updating the provisioning profile also?

Ah, yes, I believe it does.  Apologies, I didn't consider that when I opened the request.  Thanks for catching it.

fastlane/Fastfile
AliSoftware: ~~@mokagio given that's the second repo/app for which we need to do that trick (removing AppClip for Enterprise / Prototype builds to compile), worth P2-ing about it for future reference?~~

NVM, I missed your comment above suggesting to centralize this at the `release-toolkit` level, which would indeed be an even better idea 👍

andrewdmontgomery:
This happens on a device, when I attempt to install/run the Demo App after having installed/ran the App Clip:

image

Edit: I had more than one previous version of the "Gravatar Demo" apps installed, from back when we had two demo targets (SwiftUI and UIKit).  Removed.  Problem solved.

mokagio:
> Also, we need to test OAuth in the app clip target so we'll need this config. Does this require updating the provisioning profile also? @mokagio

@pinarol @andrewdmontgomery 

51b5a1ac addresses this.

The app clip already had the associated domains capability because of the `appclips:gravatar.com` domain, so no update at the developer portal level was needed.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">This PR is tagged with <code>do not merge</code> label(s).</td>
    </tr>
  </tbody>
</table>

</p>

mokagio:
Hey @andrewdmontgomery @pinarol 👋 

Wanted to check in on the state of this. Pinar mentioned "this effort might move to a separate repo," did you make a decision on that?

Not trying to put pressure on you, just touching base. As far as I'm concerned, this PR can remain open indefinitely. No issue on my end.

pinarol:
@mokagio This PR has been super useful, but I think we shouldn't merge it. We can keep the branch a while longer, but I think we can close the PR. WDYT @andrewdmontgomery ?

mokagio:
Sounds good to me @pinarol

PR #662, opened 2025-02-07
Simplify SwiftFormat Invocations

andrewdmontgomery:
### Description

This just consolidates the two individual invocations of SwiftFormat (one for each path we check) into a single invocation.  When I originally refactored this, I couldn't get that to work, hence the two invocations.

The key here was moving `--lint` after the list of paths:

```shell
# Works
[docker command] --lint Sources

# Fails with error about arguments
[docker command] --lint Sources Tests

# Works
[docker command] Sources Tests --lint
```

### Testing Steps

- [ ] CI should be 🟢 
- [ ] `make swiftformat-lint` should REPORT issues in `Sources/` and `Tests`
- [ ] `make swiftformat` should FIX issues in `Sources/` and `Tests/`

The following code will trigger SwiftFormat:
```swift
if {
    true
}
else {
    false
}
```

Add that code to two source files:
- Any source file in `Sources/`
- Any source file in `Tests/`

1. Run `make swiftformat-lint`
   - [ ] OBSERVE that it reports both files with errors
2. Run `make swiftformat`
   - [ ] OBSERVE that SwiftFormat fixed that code block in both source files:

```swift
if {
    true
} else {
    false
}
```

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -29,11 +29,7 @@ steps:
         plugins:
           - docker#v5.12.0:
               image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
-              command: ["--lint", "Sources"]
-              workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
-          - docker#v5.12.0:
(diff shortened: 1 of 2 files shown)


etoledom (approved):
Following the testing steps, everything worked as expected 🎉

PR #665, opened 2025-02-11
Make ImageCache init private

### LINKED ISSUE #388: Make "ImageCache" init private
Let's make "ImageCache" init private to avoid such issues: https://github.com/Automattic/Gravatar-SDK-iOS/pull/387

andrewdmontgomery:
Closes #388 

### Description

This PR makes the ImageCache init private. Since `TestImageCache` was using an ImageCache as its storage, I refactored it to use a dictionary.

And since it now has its own caching logic, I thought it would be good to unit-test it.

### Testing Steps

- [ ] Code review
- [ ] CI should be 🟢


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false"><code>Package.swift</code> was changed without updating its corresponding <code>Package.resolved</code>. Please resolve the Swift packages as appropriate to your project setup (e.g. in Xcode or by running <code>swift package resolve</code>).</td>
    </tr>
  </tbody>
</table>

</p>

Sources/Gravatar/Cache/ImageCaching.swift
```diff
@@ -19,6 +19,9 @@ public protocol ImageCaching: Sendable {
     /// `.inProgress(task)`  is used by the image downloader to check if there's already an ongoing download task for the same image. If yes, the image
     /// downloader  awaits that ask instead of starting a new one.
     func getEntry(with key: String) -> CacheEntry?
+
+    /// Clears all entries from the cache
+    func clear()
```
andrewdmontgomery, line 24: This isn't strictly necessary for this PR.  My original attempt at this was to just use the `ImageCache.shared` instance in the `TestImageCache` class.  In that case, I wented to clear the cache between test runs.

But that required setting a bunch of other `private` types to `package`.  So I refactored `TestImageClass` so that it doesn't use `ImageCache.shared`.  Testing no longer requires clearing the cache.

I can remove this code.  But I wondered if we wanted to the ability to clear the cache in the Demo app.  It's an in-memory cache.  But I wonder if it would be helpful while debugging.

Sources/TestHelpers/TestImageCache.swift
```diff
@@ -2,50 +2,60 @@ import Gravatar
 import UIKit
 
 package final class TestImageCache: ImageCaching, @unchecked Sendable {
-    let imageCache = ImageCache()
+    private var cache: [String: CacheEntry] = [:]
```
andrewdmontgomery, line 5: By using a Dictionary instead of the `ImageCache.shared` instance, we avoid having to make a bunch of private types `package`:

- `private class NSCacheForImage`
- `private final class CacheEntryObject`
- `NSCache` extension `fileprivate subscript(_ key): String) -> CacheEntry?`

Sources/TestHelpers/TestImageCache.swift
```diff
@@ -2,50 +2,60 @@ import Gravatar
...
 
-    package typealias CacheMessage = (operation: CacheMessageType, key: String)
+    typealias CacheMessage = (operation: CacheMessageType, key: String?)
```
andrewdmontgomery, line 7: This compiles with `internal` access.

More importantly, I made `key` optional because there is no `key` used when clearing the cache.

Sources/Gravatar/Cache/ImageCaching.swift
pinarol: AFAIU this is just added for the unit tests. We want to test the default implementation there which is the `ImageCache` struct. So I think we don't need to add a new method to the protocol just for unit testing purpose. 🤔 

We can cast the shared instance to `ImageCache` and add `clear()` only to `ImageCache`. 

Like:

```
 let cache = try XCTUnwrap(ImageCache.shared as? ImageCache)
 cache.clear()
```

andrewdmontgomery: Ah shoot, I had meant to add a comment about this.

Yes, true, we only need it for testing in this PR.  But I wondered if we wanted to make it clearable, generally.  I know it's an in-memory cache.  But are there times when we (or developers) might legitimately want to clear the cache without quitting the app?

I was thinking it would be handy during development.

pinarol: Cool why not...

pinarol: Oh wait, I talked too soon. Adding a new protocol method is a breaking change. So it would require a major release and we wouldn't want that just for this.

> during development

During development is easy, `ImageCache` is a public struct, one can typecast the shared instance to it and call `clear()` that way.

andrewdmontgomery: > Adding a new protocol method is a breaking change. So it would require a major release and we wouldn't want that just for this.

That's a good point.  Since it would be adding functionality (not breaking existing functionality), it would require a `MINOR` release.

I just removed that.  But then it occurred to me, that the main purpose of this PR is to remove functionality.  That breaks existing functionality.  That requires a `MAJOR` release.

I think this PR should not target `trunk`.  It should target our next major release.

pinarol: Ooh good point. I don't know why but I assumed the `init()` was was just internal. Yeah we can freeze this PR for now.

pinarol: > That's a good point. Since it would be adding functionality (not breaking existing functionality), it would require a MINOR release.

Having said that, adding things to a "protocol" is not a new functionality I think, It's "demanding" new functionality. Like for example, Jetpack has its own image cache that implements this protocol. After this change it would generate a build error because Jetpack misses the implementation for this new method.

PR #676, opened 2025-03-21
Demo Project: Convert Groups to Buildable Folder references

andrewdmontgomery:
### Description

This makes two changes:
- Converts the `Gravatar-Demo` group/folder in the project to a "Buildable Folder reference".  This simplifies and reduces commit churn of the project file.
- Moves configuration files and resources into dedicated folders, to give priority to the source files

### Buildable Folder References in Xcode 16

[Xcode 16 introduced](https://href.li/?https://developer.apple.com/documentation/xcode-release-notes/xcode-16-release-notes#Project-Management) "Buildable Folder references":

> - Minimize project file changes, and avoid version control conflicts with buildable folder references.
> Convert an existing group to a buildable folder with the “Convert to Folder” context menu item in the Project Navigator. Buildable folders only record the folder path into the project file without enumerating the contained files. This minimizes diffs to the project when files are added and removed, and avoids source control conflicts with your team.
To use a folder as an opaque copiable resource, the default behavior before Xcode 16, uncheck the “Build Folder Contents” option in the File Inspector. (123729918)

With a Buildable Folder reference:
- Individual files in the folder are no longer tracked in the project file
- The folder reference can be added to a target
   - Any files added to that folder will then automatically be added to the target
      - This works well for source files and any other files that should be added to a target
      - Files that should NOT be added to a target (such as an Info.plist, which is processed separately) can be excluded from being added to the target on a per-file bases, using the File inspector

Note that any file added to this folder is automatically added to the target, even if the file is added from outside of Xcode.

### Testing Steps

- [ ] Build and run the demo project
- [ ] Use Xcode to add a new source file to the `Gravatar-Demo` folder
   - [ ] Observe that it is automatically added to the `Compile sources` Build Phase for the `Gravatar-Demo` target
   - [ ] Observe that the `Demo/Gravatar-Demo.xcodeproj/project.pbxproj` does NOT change
- [ ] In the Finder, add a new source file to the `Gravatar-Demo` folder (you can duplicate one)
   - [ ] Observe that it is automatically added to the `Compile sources` Build Phase for the `Gravatar-Demo` target
   - [ ] Observe that the `Demo/Gravatar-Demo.xcodeproj/project.pbxproj` does NOT change


andrewdmontgomery:
This change caused us to stumble into a bug with fastlane.  Fastlane will be fixing this.  In the interim, we have a simple workaround in place (#681).

Once Fastlane has a fix in place, we can update Fastlane, and revert that PR.  In fact, we should revert _most_ of that PR, but not all of it, since one commit actually fixes a typo (which would be a nice fix to keep).  See that PR for details.

pinarol (approved):
LGTM 🎉

Issue #682, opened 2025-03-27
Bug - Quick Editor WebView Holds Onto Old Email After Sign Out


haightt:
Our app uses the Quick Editor to manage our users' avatars. However, I notice an issue when a user signs out and a new one signs in. 

When the Quick Editor opens, it shows the profile of the newly signed-in user.

Then, after selecting Continue and authorizing the WebView to sign in, I am presented with the previous user's email.

![Image](https://github.com/user-attachments/assets/62ab8dbd-2953-488a-bc91-04d91797cc53)

Whereas when I try to sign in initially with the first user:

![Image](https://github.com/user-attachments/assets/6ebe1310-c800-475c-998b-e04f966dcd5d)

haightt:
I can reproduce this in the demo application.

Steps to reproduce:
1. In the demo app, tap `Quick Editor`
2. Enter an email in the text field
3. Tap `Show Quick Editor`
4. Go through the authentication process
5. Tap `Logout`
6. Enter a different email in the text field
7. Tap `Show Quick Editor`
8. The previous email will display in the authorization WebView

https://github.com/user-attachments/assets/b5f375f2-502d-40bb-8fe6-4a8f0f6926a7

haightt:
I have a PR that fixes this here: https://github.com/Automattic/Gravatar-SDK-iOS/pull/683

pinarol:
Thanks for the issue. 👋  I dropped a response [here](https://github.com/Automattic/Gravatar-SDK-iOS/pull/683#issuecomment-2782911272).

PR #683, opened 2025-03-27
Set the ASWebAuthenticationSession to prefer an ephemeral authorization session

haightt:
Closes #[682](https://github.com/Automattic/Gravatar-SDK-iOS/issues/682)

### Description

This change sets the ASWebAuthenticationSession to prefer an ephemeral authorization session. This prevents the WebView from retaining the user's information if they sign out, their session token is deleted, and a new user signs in to the application.

### Testing Steps
1. In the demo app, tap Quick Editor
2. Enter an email in the text field
3. Tap Show Quick Editor
4. Go through the authentication process
5. Tap Logout
6. Enter a different email in the text field
7. Tap Show Quick Editor
8. The correct email should now display in the WebView

https://github.com/user-attachments/assets/061fc686-2e5f-42b0-9165-84b22f0195e9

### DIFF
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OldAuthenticationSession.swift
@@ -32,6 +32,7 @@ final class OldAuthenticationSession: NSObject, Sendable {
             Task { @MainActor in
                 await sessionStorage.save(session)
                 session.presentationContextProvider = self
+                session.prefersEphemeralWebBrowserSession = true
                 session.start()
             }
         }


pinarol:
Thanks for the PR @haightt 🎉 We will work on a solution to resolve your issue. But we think we can help you better if we learn more about your use case. 

At the moment we expect users to use the "Login with a different account" link to go back to the authentication step and the user should see the other email there. Do you think that's not suitable for your particular use case? We'd like to learn more about how you use the Gravatar Quick Editor inside your app if you don't mind sharing more information. What kind of app are you building and what is the user flow that uses the Gravatar Quick Editor? Does your use case require you to switch between different emails often? 

Screenshot 2025-04-07 at 11 01 27

Gravatar access token has 2 weeks lifetime, using `prefersEphemeralWebBrowserSession = true` would require users to login from scratch by entering their passcode every 2 weeks. Thus, we want to avoid making the OAuth flow more difficult. When `prefersEphemeralWebBrowserSession = false`, the browser remembers the authentication state so the user only goes through the authorization step by just tapping "Approve". 

But depending on your use case, we can make `prefersEphemeralWebBrowserSession` configurable via a new flag. Does that sound good?

And we are happy to keep communication channels open, so don't hesitate to reach out if you have other feedbacks.

Also, if you'd like to chat with us on slack, let us know your email via this contact form https://support.gravatar.com/contact/ so we can invite you.

pinarol:
Closing this one as we addressed the request in this PR: https://github.com/Automattic/Gravatar-SDK-iOS/pull/701

PR #678, opened 2025-03-26
Represent profile views in SwiftUI

pinarol:
Closes #191

### Description

Adds a public type `ProfileViewRepresentable` to provide SwiftUI support for profile views. Not sure it's the best name though. Open to suggestions.

### Testing Steps

`Demo app > SwiftUI Demos > Profile view`. And try out different styles. Tap the actions etc.

`Demo app > SwiftUI Demos > Profile editor with oauth`. Change your avatar and observe the avatar in the profile card changes in the demo page.

PR #685, opened 2025-03-31
Profile cards SwiftUI wrapper - small demo views changes

etoledom:
### Description

Small changes over https://github.com/Automattic/Gravatar-SDK-iOS/pull/678

The idea is to use the default padding, and making the Profile Card bounds visible by changing the screen background color.


pinarol (approved):
LGTM 🎉

PR #689, opened 2025-04-09
QE: Add the new authorized profile endpoint to fetch the profile

### LINKED ISSUE #672: Use the new authorized profile endpoint in the QE
`/me/profile` check the [specs](https://github.a8c.com/Automattic/gravatar/blob/f1575dd835f7a786454623a1b5ddc5be004faf40/docs/gravatar-public-openapi.json#L880) for further info.

With this new endpoint, when a new user in the Jetpack apps opens QE for the first time, the profile card will be properly loaded because the identity will be created. So we can get rid of the extra profile refresh we did [here](https://github.com/Automattic/Gravatar-SDK-iOS/pull/615).

pinarol:
Closes #672 

### Description

With this new endpoint, when a new user in the Jetpack apps opens QE for the first time, the profile card will be properly loaded because the identity will be created. So I am also reverting [this change](https://github.com/Automattic/Gravatar-SDK-iOS/pull/615).

### Testing Steps

Same test steps with https://github.com/Automattic/Gravatar-SDK-iOS/pull/615
Note: [www.emailondeck.com](http://www.emailondeck.com/) doesn't work anymore but you can create a new email by adding `+` at your Gmail address before `@` like `myexistingaccount+@gmail.com`.

### DIFF
--- a/Sources/Gravatar/Network/Services/ProfileService.swift
+++ b/Sources/Gravatar/Network/Services/ProfileService.swift
@@ -2,6 +2,7 @@ import Foundation
 
 private let baseURL = URL(string: "https://api.gravatar.com/v3/profiles/")!
 private let avatarsBaseURLComponents = URLComponents(string: "https://api.gravatar.com/v3/me/avatars")!
+private let meProfileURL = URL(string: "https://api.gravatar.com/v3/me/profile")!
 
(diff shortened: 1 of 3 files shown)


etoledom (approved):
Looks good and works as described 🎉

PR #690, opened 2025-04-09
Improving VoiceOver on the Avatars grid.

### LINKED ISSUE #674: Accessibility: Avatar selection
With VoiceOver turned on:

When selecting an avatar, screen reader will only read "image".

Expected:

When selecting an avatar, screen reader should read something like:
"Avatar image, - alt description -, button. Tap to select. 
"Selected, Avatar image. - Alt description - "

- When changing the label after selecting the image, VoiceOver should read it again, now with the "selected" word added to it. If this doesn't occur. There should be an implementation to make this happen.

https://github.com/user-attachments/assets/1c8a85e8-7e37-4df8-9f53-0aff2226626f

etoledom:
Closes #674

### Description

This PR improves VoiceOver on the Avatar grid. When the avatar is selected by VoiceOver, it will read the alt text, and say if it's the selected image. It will also read the toast if an error occurs.

https://github.com/user-attachments/assets/0bc87761-5a9e-4584-8032-50a9385e8830

### Testing Steps

- Run the demo app
- Display the Quick Editor
- Enable Voice Over
- Navigate around the avatars grid.
- Check that:
  - Selecting an avatar with screen reader will read the alt text.
  - Selecting with screen reader a selected avatar, will read "selected" first.
  - The options button will read "Avatar Options".
  - Double tapping to run the action on a non-selected avatar will make it selected, and Voice Over will read "selected".
  - Selecting an avatar without connectivity, Voice Over will read the toast's error text.


pinarol:
The rest works as described but I couldn't manage to observe this:

> Selecting an avatar without connectivity, Voice Over will read the toast's error text.

https://github.com/user-attachments/assets/8b1f589c-4761-4ab3-ae3b-0cd5dae2b636

etoledom:
> The rest works as described but I couldn't manage to observe this:
> 
> > Selecting an avatar without connectivity, Voice Over will read the toast's error text.
> 

@pinarol - This is the no-connection error, right? 
Though it shouldn't influence the kind of error.

I've been trying it out and I can't reproduce the problem:


https://github.com/user-attachments/assets/a2eee65d-73a4-4f6b-9aa2-df656d35a8a7

pinarol:
Yes that's the connection error. Maybe it's about the length of the `alt text`, and by the time it finishes reading the alt text, the error message starts to disappear so it doesn't read it? To repro the issue, it can be helpful to try with a longer `alt text` (mine was `Pınar's avatar` and it fails to read it quickly in English because of the letter "ı" I think).

I did a quick search and it seems like `UIAccessibility.post(notification: .announcement, argument: <the error message >)` is a better way to read out toast messages due to their disappearing nature? WDYT?

```
class ToastManager: ObservableObject {
    @Published var toasts: [ToastItem] = []

    func showToast(_ message: String, type: ToastType = .info, stackingBehavior: ToastStackingBehavior = .avoidStackingWithSameMessage) {
        UIAccessibility.post(notification: .announcement, argument: message) // <<<<<< announce here
        let toast = ToastItem(message: message, type: type, stackingBehavior: stackingBehavior)
        dismissExistingIfNeeded(upcomingToast: toast)
        toasts.append(toast)
        DispatchQueue.main.asyncAfter(deadline: .now() + calculateToastDuration(for: message)) {
            self.removeToast(toast.id)
        }
    }

```

etoledom:
> Maybe it's about the length of the alt text

I did try with many different alt text lengths, and it all worked out good for me. The alt text reading is interrupted by the focus change.

> I did a quick search and it seems like UIAccessibility.post(notification: .announcement, argument: <the error message >) is a better way to read out toast messages due to their disappearing nature? WDYT?

This indeed looks like a much better fit, as it also won't move the focus from the avatar after a toast error.

I've implemented this on a new commit.
@pinarol - Please check if this also solve the problem you are experiencing 🙏

pinarol:
> @pinarol - Please check if this also solve the problem you are experiencing 🙏

Yes it does solve. 🎉 
Why don't we enable it for all toast types though? I think this is useful for any kind of toast.

etoledom:
> Why don't we enable it for all toast types though? 

When it was changing focus, it was way too distracting to do so after every avatar selection.
Now testing this with the post notification, it feels much more natural, so I agree on let voice over read all toast messages 👍 

Updated sent @pinarol

pinarol (approved):
Looks and works great 🎉 🚢

PR #693, opened 2025-04-10
Propagate color scheme to QE sheet

### LINKED ISSUE #692: QE Color scheme preference doesn't propagate to the QuickEditor in SwiftUI
**Describe the bug**

Demo app > SwiftUI Demos > Profile editor with oauth
Set Color Scheme to Dark (or to whatever is **not** the system's color scheme)
Tap "Open Profile editor ..."
Result: The QE sheet opens in light mode

**Expected behavior**

Result: The QE sheet opens in dark mode

pinarol:
Closes #692

### Description

Seems like the preferredColorScheme doesn't propagate into the sheet. So we need to apply it to the View **inside** the sheet. We also need to pass it using `.environment(\.colorScheme, ...)` instead.

### Testing Steps

Follow the steps in #692

### DIFF
--- a/Demo/Demo/Gravatar-Demo/SwiftUI/DemoProfileEditorView.swift
+++ b/Demo/Demo/Gravatar-Demo/SwiftUI/DemoProfileEditorView.swift
@@ -49,7 +49,7 @@ struct DemoProfileEditorView: View {
                                 onDismiss: {
                                     updateHasSession(with: email)
                                 }
-                        )
+                            ).environment(\.colorScheme, ColorScheme(selectedScheme) ?? .light)
                     }
                     else {
                         view
@@ -64,6 +64,7 @@ struct DemoProfileEditorView: View {
                                     updateHasSession(with: email)
                                 }
                             )
+                            .environment(\.colorScheme, ColorScheme(selectedScheme) ?? .light)
                     }
                 }
 
@@ -85,7 +86,7 @@ struct DemoProfileEditorView: View {
             updateHasSession(with: newValue)
             requestProfile()
         }
-        .environment(\.colorScheme, ColorScheme(selectedScheme) ?? .light)
+        
     }
 
     func requestProfile() {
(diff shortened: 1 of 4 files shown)


etoledom (approved):
Looking good and works as described 🎉 

One thought. I would spect that by choosing a color scheme, just the QE would be affected and not the demo screen too. Something like in the UIKit demo:



From what I see, this is implementation detail by the demo app, and this behavior can also be achieved, right?

I think it shouldn't be necessary to set the parent environment to set the QE user interface style.

pinarol:
> One thought. I would spect that by choosing a color scheme, just the QE would be affected and not the demo screen too. Something like in the UIKit demo:

It's easy to achieve that in UIKit because we pass the interfaceStyle as an explicit parameter to `QuickEditorPresenter`. But in SwiftUI, we use the environment variable that's set in the parent level. So we would probably need to set constant colors to all the rest of the items one by one in the page, not sure if there's an easier way or worth the maintenance effort in the long run. 🤔

pinarol:
> I think it shouldn't be necessary to set the parent environment to set the QE user interface style.

We rely on the environment variable for the colorScheme from the beginning and the environment variable is set on the parent level. Alternatively we can pass the colorScheme via `gravatarQuickEditorSheet(...)` modifier but it means we need a new parameter there. WDYT? @etoledom

etoledom:
> We rely on the environment variable for the colorScheme from the beginning and the environment variable is set on the parent level. Alternatively we can pass the colorScheme via gravatarQuickEditorSheet(...) modifier but it means we need a new parameter there. WDYT? @etoledom

Uhm interesting. I'm not sure what's the best approach.

I did a small test, overriding the environment on the modifier which presents the sheet directly, instead than on the whole body.

CleanShot 2025-04-11 at 14 25 09@2x

This results on the same behavior than the UIKit editor. 
My concern was only to have the technical possibility of separating the color scheme shown on the sheet from the one set on the screen which presents the sheet.

Do you think this is good enough? It does make it for me :)

pinarol:
> I did a small test, overriding the environment on the modifier which presents the sheet directly, instead than on the whole body.

I swear I tried this before and it didn't work, but apparently I overlooked something and somehow failed to actually try. Oh yes if it works, that's quite good to me too.

pinarol:
This works well 🎉 Pushed the change.

PR #695, opened 2025-04-11
Custom image editor injection from UIKit QE presenter

### LINKED ISSUE #686: QE: Allow injecting custom ImageEditor via UIKit QuickEditorPresenter as well
At the moment we only allow injecting a custom ImageEditor via our SwiftUI modifiers `.gravatarQuickEditorSheet(...)`.

Let's allow it via QuickEditorPresenter as well.

etoledom:
Closes #686

### Description

This PR adds support to inject a custom image editor controller when using the UIKit Quick Editor presenter.

Ideally on the user side, everything is UIKit related, and they don't need to worry about bridging with SwiftUI, so I have implemented the bridge internally, using a `CustomImageEditorControllerRepresentable` internal class.

This new config parameter is optional, so it shouldn't introduce any breaking change.

A new option has been introduced in the UIKit Quick Editor demo screen, to use the default image editor, or a simple custom image editor, similar to the one used in the SwiftUI demo.

### Testing Steps

- Run the demo app
- Go to `Quick Editor` demo screen
- Switch to `Custom image editor` toggle
- Tap on `Show quick editor`
- Upload an image
  - Check that the dummy custom image editor appears instead of the default one.


Sources/GravatarUI/QuickEditor/QuickEditorViewController.swift
```diff
@@ -187,3 +211,39 @@ public struct QuickEditorPresenter {
         parent.present(quickEditor, animated: animated, completion: completion)
     }
 }
+
+@MainActor
+public protocol CustomImageEditorController: UIViewController {
+    var inputImage: UIImage { get }
+    var editingDidFinish: @Sendable (UIImage) -> Void { get }
+
+    init(
```
pinarol: I think we shouldn't have the `init(...)` inside the protocol. We don't care how they create their `UIViewController`, the important thing is that they give us the `inputImage` and `editingDidFinish` and the protocol conformance will ensure that.

Sources/GravatarUI/QuickEditor/QuickEditorViewController.swift
```diff
@@ -1,7 +1,11 @@
 import SwiftUI
 import UIKit
 
+public typealias CustomImageEditorControllerFactory = (UIImage, @escaping @Sendable (UIImage) -> Void) -> CustomImageEditorController
```
pinarol: The `Factory` suffix brings to mind the `Factory Pattern` from `OOP`. Would it make sense to use a different name, like `CustomImageEditorControllerProvider` instead?

pinarol (commented):
Looks good, just small comments. Thanks.

Sources/GravatarUI/QuickEditor/QuickEditorViewController.swift
etoledom: 💯

Sources/GravatarUI/QuickEditor/QuickEditorViewController.swift
etoledom: Right, that's much better 👍 
Thank you!

pinarol (approved):
LGTM 🎉

PR #698, opened 2025-04-15
Show connectivity/session errors when selecting an avatar

### LINKED ISSUE #696: Show URLSession's error message when avatar selection fails due to lack of internet connection
We show the URLSession's localized error message when the upload fails due to lack of connectivity, let's do the same for avatar selection.

pinarol:
Closes #696 

### Description

I noticed that we always show a generic error message if avatar selection fails. I think more context is better. I suggest showing the localized error message from the `URLSession` which gives us connectivity related messages as well.

### Testing Steps

Open Quick Editor
Disconnect from the internet
Try selecting a different avatar
You should see a connectivity related error in the toast


etoledom (approved):
Working as described 🎉 

It also recognised "Connection time out" and "Couldn't connect to the server" errors.

PR #701, opened 2025-04-17
QE: Add new public method to enable ephemeral web browser session

### LINKED ISSUE #682: Bug - Quick Editor WebView Holds Onto Old Email After Sign Out

pinarol:
Closes #682

### Description

Allows enabling ephemeral web browser session for QE OAuth flow. 

### Testing Steps

Test both SwiftUI and UIKit demo screens for QuickEditor
Use the new toggle to enable/disable ephemeral web browser session
Go through OAuth flow
When enabled, your session is not kept so you need to go through login from scratch
When disabled, you just need to go through the authorization step and click "Approve"

### DIFF
--- a/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
+++ b/Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
@@ -2,7 +2,7 @@ import AuthenticationServices
 
 public struct OAuthSession: Sendable {
     static let shared = OAuthSession()
-    private var emailStorage = SessionEmailStorage()
+    private var sessionData = SessionData()
 
     private let storage: SecureStorage
     private let authenticationSession: AuthenticationSession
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -247,15 +287,24 @@ extension [URLQueryItem] {
...
+    func authenticate(using url: URL, prefersEphemeralWebBrowserSession: Bool, callbackURLComponents: URLComponents) async throws -> URL
     func cancel() async
 }
 
 extension OldAuthenticationSession: AuthenticationSession {}
 
 // Stores the email used for the current OAuth flow
-private actor SessionEmailStorage {
-    var current: Email?
+private actor SessionData {
```
pinarol, line 297: Renamed this since it has additional data now. And removed `Storage` from the name since it brings to mind a rather persistent store.

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
```diff
@@ -17,6 +17,42 @@ public struct OAuthSession: Sendable {
...
+    /// Given that Gravatar access tokens expire after 2 weeks, this effectively means logging in every 2 weeks.
+    ///
+    /// When set to `false`, the user is still redirected through the OAuth flow every 2 weeks, but their session remains
+    /// active in the browser. As a result, they only need to authorize the app by tapping “Approve” without re-entering credentials.
+    ///
+    /// See also:  https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/prefersephemeralwebbrowsersession
+    public static func setPrefersEphemeralWebBrowserSession(_ value: Bool) async {
+        await shared.setPrefersEphemeralWebBrowserSession(value)
+    }
+
```
pinarol, line 55: Unfortunately we both need the static and non-static versions of the same methods, one for SwiftUI other for UIKit. I really thought of refactoring the structure here since it's causing duplications and feels difficult to understand but didn't want to do that as part of this effort.

etoledom: What about making the `shared` instance public?

etoledom (approved):
Working as described 👍

Sources/GravatarUI/SwiftUI/OAuthSession/OAuthSession.swift
pinarol: That would definitely help to a degree. We would at least get rid of the `static` version of the same image. Let me try how it would look like.

pinarol: I am making a new PR for this: https://github.com/Automattic/Gravatar-SDK-iOS/pull/716

PR #702, opened 2025-04-17
Improve Voice Over on the avatar image.

### LINKED ISSUE #700: Accessibility: Support system generated image descriptions
I'd like the system to read auto-generated image description after the alt text when `VoiceOver Recognition > Image descriptions` are on in the device. There are apps that do this like Tumblr, and it's really cool. I think we can support this for avatars in the QE and in profile cards.

etoledom:
Closes #700

### Description

- Allow reading of image description.
- Include the avatar options menu in the accessibility avatar element.

By sweeping up and down, the user can choose the action to be executed by double tap to be either selecting the avatar, or opening the options menu.

The image description comes at the very end of Voice Over reading.

https://github.com/user-attachments/assets/518a9adb-a247-4c71-9d01-2421b904d82a

### Testing Steps

- Run the demo app
- Display the Quick Editor
- Enable Voice Over
- Navigate around the avatars grid.
- Check that:
  - Selecting an avatar with screen reader will read the alt text.
  - Voice Over will mention "Actions Available"
  - "Select this avatar" action will be read only on non-selected avatar (this text can probably be improved)
  - The last part of the Voice Over description of the element, will be a description of the image.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
+++ b/Sources/GravatarUI/SwiftUI/AvatarPicker/Views/AvatarPickerAvatarView.swift
@@ -64,17 +64,20 @@ struct AvatarPickerAvatarView: View {
             .onTapGesture {
                 onAvatarTap(avatar)
             }
-            .accessibilityRepresentation {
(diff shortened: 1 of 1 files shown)


pinarol (approved):
Great 🎉

PR #717, opened 2025-04-24
Make API calls public

### LINKED ISSUE #699: Make API calls public
Let's make the API calls that are documented in the backend specs publicly available via our SDK similar to Android. 
Let's look for ways to keep auto-generated response models internal when convenient but if they are as big as `Profile` then it is fine to avoid duplication there and open the generated one directly.

pinarol:
Closes #699

### Description

Opens following functionality as public:
- Set as avatar
- Update alt text
- Update rating
- Delete avatar
- Fetch own profile

Introduces a new protocol as `AvatarDetails` to hide the generated type `Avatar`.

Reduces the access modifier of `Avatar` , `Avatar.Rating`, `AvatarRating`, `UpdateAvatarRequest` as `internal` to avoid using them from `GravatarUI` directly. This way, the code gets simpler and we make sure we use the public APIs so they get tested.

Replaces the `AvatarIdentifier` usage with a new String alias `ImageID`. This is because `AvatarIdentifier` represents the public avatar ID which is different from the ID we get from `/v3/me/avatars`. This PR uses `ImageID` everywhere related to `/v3/me/avatars`.

### Testing Steps

Test these functionalities:

- Set as avatar
- Update alt text
- Update rating
- Delete avatar


Sources/TestHelpers/TestImageProcessor.swift
```diff
@@ -2,15 +2,13 @@ import Gravatar
...
-        self.identifier = identifier
+    let image: UIImage?
+    package init(image: UIImage? = nil) {
+        self.image = image
     }
 
     package func process(_: Data) -> UIImage? {
-        let image = UIImage()
-        image.accessibilityIdentifier = identifier
-        return image
```
pinarol, line 13: This fixes a warning:

> Main actor-isolated property 'accessibilityIdentifier' can not be mutated from a nonisolated context; this is an error in the Swift 6 language mode

Sources/Gravatar/Network/Services/AvatarService.swift
```diff
@@ -47,40 +47,52 @@ public struct AvatarService: Sendable {
...
             throw ImageUploadError.responseError(reason: .unexpected(error))
         }
     }
 
-    package func delete(avatarID: String, accessToken: String) async throws {
-        var request = URLRequest(url: .avatarsURL.appendingPathComponent(avatarID))
+    /// Deletes the avatar.
+    /// - Parameters:
+    ///   - imageID: Image ID of the avatar to be deleted. This is the ID of each avatar that returns us via `/v3/me/avatars`. See: `fetchAvatars(...)` from
+    /// ``ProfileService``.
```
etoledom:
```suggestion
    ///   - imageID: Image ID of the avatar to be deleted. This is the ID of each avatar returned from the `/v3/me/avatars` endpoint. 
    /// See: ``ProfileService/fetchAvatars(profileID:token:)``.
```

etoledom: You can make a link directly to a member of type with a `/` (i.e: ` ``SomeStruct/aMethod()`` `)

etoledom (approved):
Looks good!
Tested all new public requests, everyting works as expected.

Just a small comment on docc comments.

Sources/Gravatar/Network/Services/AvatarService.swift
etoledom: This also repeats in a few other methods

pinarol: Great tip. All updated.

PR #718, opened 2025-04-24
About editor scope 

### LINKED ISSUE #706: New About QE scope
New scope for About editing.

This task includes a corresponding temporary empty content shown in the QE,
and a scope selector in the Demo app, both UIKit and SwiftUI.

etoledom:
Closes #706

### Description

This PR implements the About info editor scope for the Quick Editor.

Bigest public changes:

- New `gravatarQuickEditorSheet()` method on `View` extension.
  - The `scope` parameter is replaced by `scopeOption` parameter which takes a `QuickEditorScopeOption` instance
  - The `avatarUpdatedHandler` is replaced by a generic `updatedHandler` which specifies the kind of update via a `QuickEditorUpdateType` parameter.
  - Pre iOS16 version of this new method.
  - The old `gravatarQuickEditorSheet()` gets deprecated.

- `QuickEditorScopeOption` is astruct with will replace the `QuickEditorScope` enum (which got deprecated) to avoid breaking changes on any new scope addition.
- Corresponding changes applied to UIKit's `QuickEditorPresenter`.

CleanShot 2025-04-25 at 20 03 30

### Testing Steps

- Test both UIKit and SwiftUI demo apps.
- About editor scope should present a blank screen.
- OAuth should work as expected regardless of the scope.
- Check the code looking for unwanted breaking changes.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
```diff
@@ -1,11 +1,13 @@
 import Gravatar
 import SwiftUI
 
-@available(iOS, deprecated: 16.0, renamed: "QuickEditorScope")
+@available(iOS, deprecated: 16.0, renamed: "QuickEditorScopeOption", message: "This will become internal in a next mayor release.")
 public enum QuickEditorScopeType: Sendable {
     case avatarPicker
+    case aboutInfoEditor
```
pinarol: Was this new case forgotten to be removed? Because this is a breaking change?

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
```diff
@@ -1,11 +1,13 @@
 import Gravatar
 import SwiftUI
 
-@available(iOS, deprecated: 16.0, renamed: "QuickEditorScope")
+@available(iOS, deprecated: 16.0, renamed: "QuickEditorScopeOption", message: "This will become internal in a next mayor release.")
```
pinarol: mayor -> major

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
```diff
@@ -17,6 +19,21 @@ public enum QuickEditorScope: Sendable {
...
+        case aboutInfoUpdate
+    }
+
+    private let rawValue: QEUpdateType
+
+    /// Indicates that the update was triggered by a change to the user's avatar.
+    public static let avatarUpdate = Self(rawValue: .avatarUpdate)
+    /// Indicates that the update was triggered by a change to the user's about section information.
+    public static let aboutInfoUpdate = Self(rawValue: .aboutInfoUpdate)
+}
```
pinarol, line 34: Later, we can also add a kind of dictionary similar to `userInfo` to give more info about the event. Not necessary in this PR.

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
```diff
@@ -36,29 +53,42 @@ struct QuickEditor<ImageEditor: ImageEditorView>: View {
...
+        self._model = StateObject(wrappedValue: AvatarPickerViewModel(email: email, authToken: token))
+    }
+
+    init(
+        email: Email,
+        scope: QuickEditorScopeOption,
+        token: String? = nil,
+        isPresented: Binding<Bool>,
+        updatedHandler: ((QuickEditorUpdateType) -> Void)? = nil
+    ) {
```
pinarol: Seems like we don't need this new `init(...)`, no? The `customImageEditor` param in the upper `init()` is already optional.

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -34,6 +34,7 @@ extension View {
     ///   - onDismiss: (Optional) A callback to execute when the sheet is dismissed.
     /// - Returns: A modifier to display the QuickEditor sheet.
     @available(iOS, deprecated: 16.0, message: "Use the new method that takes in `QuickEditorScope`.")
+    @available(*, deprecated, renamed: "gravatarquickEditorSheet(isPresented:email:authToken:scopeOption:customImageEditor:updatedHandler:onDismiss:)")
```
pinarol: Small typo: gravatar**Q**uickEditorSheet

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -34,6 +34,7 @@ extension View {
     ///   - onDismiss: (Optional) A callback to execute when the sheet is dismissed.
     /// - Returns: A modifier to display the QuickEditor sheet.
     @available(iOS, deprecated: 16.0, message: "Use the new method that takes in `QuickEditorScope`.")
```
pinarol: Needs to be updated as:
> "Use the new method that takes in `QuickEditorScopeOption`

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -42,15 +43,51 @@ extension View {
         customImageEditor: ImageEditorBlock<some ImageEditorView>? = nil as NoCustomEditorBlock?,
         avatarUpdatedHandler: (() -> Void)? = nil,
```
pinarol: We have a `QuickEditorScopeType.aboutInfoEditor` but this callback only covers `avatarUpdatedHandler`. Was this intended?

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -42,15 +43,51 @@ extension View {
...
+    /// A modifier to display the QuickEditor sheet. QuickEditor can be used to select and upload a new avatar.
+    /// - Parameters:
+    ///   - isPresented: A Binding boolean to manage showing/hiding the sheet.
+    ///   - email: Email for the Gravatar account.
+    ///   - authToken: (Optional) Gravatar OAuth token. If not passed, Gravatar OAuth flow will start to gather the token internally.
+    ///   Pass this only if your app already has a Gravatar OAuth token.
+    ///   - scope: Scope for the QuickEditor.
+    ///   - customImageEditor: (Optional) A custom image editor to show the user right after an image is picked for
+    ///   cropping and other sorts of image editing operations.
+    ///   - avatarUpdatedHandler: (Optional) A callback to execute when a different avatar is selected.
```
pinarol: updatedHandler

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -42,15 +43,51 @@ extension View {
...
+            isPresented: isPresented,
+            customImageEditor: customImageEditor,
+            updatedHandler: { _ in
+                avatarUpdatedHandler?()
+            }
+        )
+        return modifier(ModalPresentationModifier(isPresented: isPresented, onDismiss: onDismiss, modalView: editor))
+    }
+
+    /// A modifier to display the QuickEditor sheet. QuickEditor can be used to select and upload a new avatar.
```
pinarol: > QuickEditor can be used to select and upload a new avatar.

Needs update to mention about info editing?

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -42,15 +43,51 @@ extension View {
...
+    /// - Parameters:
+    ///   - isPresented: A Binding boolean to manage showing/hiding the sheet.
+    ///   - email: Email for the Gravatar account.
+    ///   - authToken: (Optional) Gravatar OAuth token. If not passed, Gravatar OAuth flow will start to gather the token internally.
+    ///   Pass this only if your app already has a Gravatar OAuth token.
+    ///   - scope: Scope for the QuickEditor.
```
pinarol: scopeOption

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -42,15 +43,51 @@ extension View {
...
+    ///   - customImageEditor: (Optional) A custom image editor to show the user right after an image is picked for
+    ///   cropping and other sorts of image editing operations.
+    ///   - avatarUpdatedHandler: (Optional) A callback to execute when a different avatar is selected.
+    ///   - onDismiss: (Optional) A callback to execute when the sheet is dismissed.
+    /// - Returns: A modifier to display the QuickEditor sheet.
+    @available(iOS, deprecated: 16.0, message: "Use the new method that takes in `QuickEditorScope`.")
```
pinarol: I am confused about this one. It is deprecated for iOS 16, but it lets you pass `QuickEditorScopeOption` which includes presentation modes that are not supported in iOS 15. The whole point of having a `QuickEditorScopeType` was to prevent offering unsupported presentation types pre iOS 16.

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -97,6 +136,54 @@ extension View {
         }
     }
 
+    /// A modifier to display the QuickEditor sheet. QuickEditor can be used to select and upload a new avatar.
```
pinarol: Same, needs update to mention about info editing.

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -97,6 +136,54 @@ extension View {
...
+    /// - Parameters:
+    ///   - isPresented: A `Binding<Bool>` to control the presentation of the sheet.
+    ///   - email: The email address associated with the Gravatar account.
+    ///   - authToken: *(Optional)* A Gravatar OAuth token. If not provided, the QuickEditor will initiate
+    ///   the Gravatar OAuth flow to obtain a token. Provide this only if your app already has a token.
+    ///   - scopeOption: The scope option for the QuickEditor. See: ``QuickEditorScopeOption``.
+    ///   - customImageEditor: *(Optional)* A custom image editor provider to use after the user picks an image in the Avatar Picker.
+    ///   - avatarUpdatedHandler: *(Optional)* A closure called when the user selects a new avatar.
```
pinarol: updatedHandler

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -42,15 +43,51 @@ extension View {
...
+    ///   - onDismiss: (Optional) A callback to execute when the sheet is dismissed.
+    /// - Returns: A modifier to display the QuickEditor sheet.
+    @available(iOS, deprecated: 16.0, message: "Use the new method that takes in `QuickEditorScope`.")
+    public func gravatarQuickEditorSheet(
+        isPresented: Binding<Bool>,
+        email: String,
+        authToken: String? = nil,
+        scopeOption scope: QuickEditorScopeOption,
+        customImageEditor: ImageEditorBlock<some ImageEditorView>? = nil as NoCustomEditorBlock?,
+        updatedHandler: (() -> Void)? = nil,
```
pinarol: Should it not pass `QuickEditorUpdateType`?

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -97,6 +136,54 @@ extension View {
...
+    /// - Returns: A view modifier that presents the QuickEditor sheet.
+    @available(iOS 16.0, *)
+    @ViewBuilder
+    public func gravatarQuickEditorSheet(
+        isPresented: Binding<Bool>,
+        email: String,
+        authToken: String? = nil,
+        scopeOption scope: QuickEditorScopeOption,
+        customImageEditor: ImageEditorBlock<some ImageEditorView>? = nil as NoCustomEditorBlock?,
+        updatedHandler: ((QuickEditorUpdateType) -> Void)? = nil,
```
pinarol: Not a big deal but I think we should use `updateHandler` instead of `updatedHandler`. Follows Swift naming conventions better.

Sources/GravatarUI/QuickEditor/QuickEditorViewController.swift
```diff
@@ -218,7 +244,37 @@ public struct QuickEditorPresenter {
...
+    ///   - animated: Whether the presentation should be animated. Defaults to `true`.
+    ///   - completion: An optional closure called after the presentation finishes.
+    ///   - onUpdate: An optional closure triggered when the user profile info is updated.
+    ///   - onDismiss: An optional closure triggered when the editor is dismissed.
+    @MainActor
+    public func present(
+        in parent: UIViewController,
+        animated: Bool = true,
+        completion: (() -> Void)? = nil,
+        onUpdate: ((QuickEditorUpdateType) -> Void)? = nil,
```
pinarol, line 269: We can use the same naming for this parameter between UIKit and SwiftUI, let's use `updateHandler` here as well imo.

Demo/Demo/Gravatar-Demo/DemoQuickEditorViewController.swift
```diff
@@ -64,12 +64,51 @@ final class DemoQuickEditorViewController: UIViewController {
...
+    }
+    private var selectedVerticalContentPresentationStyleRepresentation: VerticalContentPresentationStyleRepresentation = .expandableMedium {
+        didSet {
+            aboutPresentationStyleButton.setTitle(
+                "Vertical Presentation Style: \(selectedVerticalContentPresentationStyleRepresentation.rawValue)",
+                for: .normal
+            )
+        }
+    }
+
```
pinarol, line 110: I don't see the presentation style option in the UI when I select `Scope: About Editor`.

Sources/GravatarUI/SwiftUI/ModalPresentationModifier.swift
```diff
@@ -21,6 +21,36 @@ struct ModalPresentationModifier<ModalView: View>: ViewModifier {
     }
 }
 
+@available(iOS 16.0, *)
+struct ModalVerticalPresentationWithStylesModifier<ModalView: View>: ViewModifier {
```
pinarol: It doesn't look like we need a new modifier. AvatarPickerModalPresentationModifier contains some quite useful bug fixes so I think we should try to reuse it by making it a bit more generic. We can brainstorm on how to do that if you like.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerContentLayout.swift
```diff
@@ -70,6 +73,21 @@ public enum AvatarPickerContentLayout: AvatarPickerContentLayoutProviding, Equat
     }
 }
 
+extension VerticalContentPresentationStyle {
+    @available(iOS 16.0, *)
+    var detents: [PresentationDetent] {
```
pinarol: Let's try to use `QEDetent` because before that we had some issues about keeping parity between detents of UIKit and SwiftUI.

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
etoledom: Exactly

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift
etoledom: True, it came when trying to avoid `QuickEditor` itself from being generic on `ImageEditorView`. It's not needed on the current implementation 👍

Sources/GravatarUI/SwiftUI/View+Additions.swift
etoledom: We [removed](https://github.com/Automattic/Gravatar-SDK-iOS/pull/718#discussion_r2063381781) `QuickEditorScopeType.aboutInfoEditor` to avoid a breaking change (I think it's the right call).

So, users wanting to use the "about info editor" will need to use the new init, as we don't want to modify this old one to avoid a build breaking change.

Sources/GravatarUI/SwiftUI/View+Additions.swift
etoledom: Right, I had this in mind but then I didn't come back to it.

What we can do is to mark the `public static func avatarPicker(config:)` as iOS16+ , and have another static function `public static func avatarPicker()` without the config, which will be pre iOS16 and we marked as deprecated for new iOS versions.

Similar to the usage of `QuickEditorScopeType` but we won't need new type declarations.

I'll try this out after fixing the typos and other small things

Sources/GravatarUI/SwiftUI/View+Additions.swift
etoledom: I'll try [this](https://github.com/Automattic/Gravatar-SDK-iOS/pull/718#discussion_r2066003755), it should give both avatar picker and about info options without the confusion of the layout options for the picker 👍

Sources/GravatarUI/SwiftUI/View+Additions.swift
etoledom: Totally 👍

Sources/GravatarUI/SwiftUI/ModalPresentationModifier.swift
etoledom: We could do something like 
`AvatarPickerContentLayout.vertical(presentationStyle: scope.aboutEditorConfig.presentationStyle)` for the About editor case (when we don't need horizontal layouts)

And probably renaming `AvatarPickerContentLayout` to `QuickEditorContentLayout` or `ModalContentLayout` for example 🤔

pinarol: Yeah I agree with the need to rename it to sth like `QuickEditorContentLayout`.

I think we can probably hold `QuickEditorScopeOption` in `AvatarPickerModalPresentationModifier` instead of `AvatarPickerContentLayout`. Then we would just need to adapt `QEDetent.detents(for:intrinsicHeight:verticalSizeClass)` method to also accept QuickEditorScopeOption(by replacing the existing `AvatarPickerContentLayout` param) and update the inner implementation of `QEDetent.detents(for:intrinsicHeight:verticalSizeClass)`. Then we won't need to do anything else to create utilities to generate detents for SwiftUI, UIKit separately because the existing `.map()` functions already do that. In theory this would work, haven't tried it though. But I remember it's important to take `verticalSizeClass` into consideration so updating `QEDetent.detents(for:intrinsicHeight:verticalSizeClass)` will help with it.

Demo/Demo/Gravatar-Demo/DemoQuickEditorViewController.swift
etoledom: Fixed

etoledom:
Out of the review comments, one big change became clear:

The selection of a layout not supported by pre-iOS 16 versions is made on the scope level, where the configuration (of such layout) is made.

So now we have 2 `avatarPicker` scopes, one for pre-iOS 16 without layout configuration, and one for iOS 16+, with layout configuration.

```swift
public struct QuickEditorScopeOption {
// (...)

    @available(iOS 16.0, *)
    public static func avatarPicker(
        _ config: AvatarPickerConfiguration
    ) -> Self {}

    @available(iOS, deprecated: 16.0, renamed: "avatarPicker(_:)", message: "Use `avatarPicker(_:)` with the new configuration instead.")
    public static func avatarPicker() -> Self {}

    public static func aboutEditor(
        _ config: AboutEditorConfiguration = .init()
    ) -> Self {}
}
```

This means that now we only expose one view modifier to show the quick editor, for all iOS versions:

```swift
    public func gravatarQuickEditorSheet(
        isPresented: Binding<Bool>,
        email: String,
        authToken: String? = nil,
        scopeOption scope: QuickEditorScopeOption,
        customImageEditor: ImageEditorBlock<some ImageEditorView>? = nil as NoCustomEditorBlock?,
        updateHandler: ((QuickEditorUpdateType) -> Void)? = nil,
        onDismiss: (() -> Void)? = nil
    ) -> some View { }
```

This is noticeable in the demo app, where the dependency on iOS version is moved to the selection of the scope, instead to the presentation of the QuickEditor. I think this is easier to use and reason with.

Sources/GravatarUI/SwiftUI/ModalPresentationModifier.swift
etoledom: I pushed changes for almost every other comment expect for this one.
We can give it a spin to find the best solution.

What you mention looks promising, but I don't understand the first part:

> I think we can probably hold QuickEditorScopeOption in AvatarPickerModalPresentationModifier instead of AvatarPickerContentLayout

Since `AvatarPickerContentLayout` doesn't hold a `QuickEditorScopeOption` instance, so I'm a bit lost there.

What I did on the last set of pushes is [this](https://github.com/Automattic/Gravatar-SDK-iOS/pull/718#discussion_r2066139657) which was a simple change and it works, but I see that we can work it out further 👍 

Maybe it's worth a separate PR?

pinarol:
The problem is that what's supported depends on whether we use SwiftUI or UIKit. :( The pre-iOS 16 restriction was there because SwiftUI doesn't support any type of detents pre-iOS 16. But the same thing isn't true for UIKit. Pre-iOS 16 UIKit supports the `.medium` detent for example and the option to prioritize scrolling, it doesn't support the custom fraction though so we fallback to `.medium` there. Doing that would also limit the capability of `QuickEditorPresenter` because we allowed passing `AvatarPickerConfiguration` so far pre-iOS 16 as well. So with this change, we are taking back a functionality from the UIKit side by forcing the usage of `avatarPicker()` there. So I feel like handling the pre/post iOS difference at the `QuickEditorScopeOption` is not a great foundation to build on. It could be more ideal if we could define these limitations for SwiftUI and UIKit APIs separately. :( 

----

Another thing, I see only 1 method for the about editor:

```
    public static func aboutEditor(
        _ config: AboutEditorConfiguration = .init()
    ) -> Self {}
```

Any reason for not having pre/post iOS 16 variations similar to `.avatarPicker(...)` ?

Sources/GravatarUI/SwiftUI/ModalPresentationModifier.swift
pinarol: I am just saying go to `QuickEditorModalPresentationModifier` and find/replace `AvatarPickerContentLayout` with `QuickEditorScopeOption`. And then update `QEDetent.detents(for:intrinsicHeight:verticalSizeClass) ` to accept `QuickEditorScopeOption` and adjust its implementation a bit to support `aboutInfoEditor` scope option as well. The rest should just compile. If still confusing, I can make a separate PR.

etoledom:
> ... Doing that would also limit the capability of `QuickEditorPresenter` because we allowed passing `AvatarPickerConfiguration` so far pre-iOS 16 as well. 

@pinarol - as we discussed internally, this was handled by bringing back a second view modifier to present the avatar picker.

I believe this is ready for another look 👍

Sources/GravatarUI/SwiftUI/View+Additions.swift
```diff
@@ -21,7 +21,7 @@ extension View {
             .padding(.vertical, borderWidth) // to prevent borders from getting clipped
     }
 
-    /// A modifier to display the QuickEditor sheet. QuickEditor can be used to select and upload a new avatar.
+    /// A modifier to display the QuickEditor sheet. The QuickEditor can be used to modify the information and avatar images of your Gravatar profile.
```
pinarol: Nit: This one feels a bit more natural to me:

> QuickEditor allows updating the Gravatar profile information and avatar.

etoledom: Super 👍

pinarol (approved):
I started making a new PR for this one. https://github.com/Automattic/Gravatar-SDK-iOS/pull/718#discussion_r2066629224 But i think we don't need to keep this one from getting merged.

🚢

PR #720, opened 2025-04-25
Add accessibility label to avatar & Generate strings

pinarol:
Closes #

### Description

Add accessibility label to avatar and generate strings.

### Testing Steps

VoiceOver on
Open QE 
Tap on the avatar in the profile card
It should readout "Your avatar"


etoledom (approved):
Works as described 🎉 

Thank you for this 🙏

PR #723, opened 2025-04-30
About editor: Reuse QE top views

### LINKED ISSUE #707: Make the the top views of the QE be reused by the scopes
The top part of the current Avatar picker will be reused by different scopes. We want to find a way to avoid code duplication.

Possibly extracting the profile card and email field out of the Avatar picker, to be presented by it's parent. 
Every different scope content would be shown above these views, aiding the switching between scopes

etoledom:
Closes #707

### Description

This PR reuses the email, profile card, and navigation items to be shared across different scopes:

The previously empty `About editor` scope now contains the shared screen, same as the avatar picker:

### Testing Steps

- Check that both scopes display the profile card, user email, and navigation items.
- Test all buttons and interactions in said UI elements.
- Check auth and no-internet error screens.


pinarol (approved):
Looks good 👍

PR #727, opened 2025-04-30
Fallback to the public profile endpoint if there's no token

### LINKED ISSUE #725: Profile card doesn't load properly on wrong email error
**Describe the bug**
When we have the wrong email error, the the profile card shows the avatar, but no name or location (testing in trunk) I believe it should either show everything or nothing.

**Screenshots**

Image

**Expected behavior**
Profile card should load.

pinarol:
Closes #725

### Description

Recently we introduced the authenticated profile fetching endpoint but it only works when there's an auth token. We should fallback to the old endpoint when there isn't.

### Testing Steps

Go to QuickEditor
Login with email A
Use the logout button in the demo app
Change the email in the demo app
Start QuickEditor but use email A for the OAuth flow
This will generate the wrong email error
Observe that the profile card is filled with data


etoledom (approved):
The card is now filled with data! 🎉

PR #730, opened 2025-05-01
About editor base networking

### LINKED ISSUE #708: Profile About fields edition endpoints
Work out the networking needed for editing the About fields.

This task would implement the endpoint itself, while the different body parameters for each form field would be implemented on each field task.

etoledom:
Closes #708 

### Description

This PR adds the `updateProfile` method to the `ProfileService`, to be able to perform patch operations on the profile.

So far, I've decided to make `UpdateProfileRequest` public to avoid internal duplications. But we can change that before merging if we decide.

I have also added a very simple UI for testing the request, it will show the response Profile's display name, or any error.

CleanShot 2025-05-01 at 13 59 39

### Testing Steps

- Use the provided test UI to try an update of the user's display name.
  - You can also close the QE and re-open to see the new name in the Profile card.
- Try an error like `Unauthorized` (modifying the given token)  to see the catch error.

### DIFF
--- a/Sources/Gravatar/Network/Services/ProfileService.swift
+++ b/Sources/Gravatar/Network/Services/ProfileService.swift
@@ -80,6 +80,23 @@ public struct ProfileService: ProfileFetching, Sendable {
             throw error.apiError()
         }
     }
+
+    /// Update profile information for the authenticated user
+    ///
+    /// Updates the profile information for the authenticated user. Only a subset of `Profile` fields are available for supported, so only the provided fields
+    /// will be updated. To unset a field, set it explicitly to an empty string.
+    ///
+    /// - Parameters:
+    ///   - props: The subset of data available for update. Only the provided fields will be updated.
+    ///   - token: Gravatar OAuth2 token.
+    /// - Returns: An asynchronously-delivered user profile with the fields updated.
+    public func updateProfile(with props: UpdateProfileRequest, token: String) async throws -> Profile {
+        var request = URLRequest(url: meProfileURL).settingAuthorizationHeaderField(with: token)
+        request.httpMethod = "PATCH"
+        request.httpBody = try? JSONEncoder().encode(props)
+        let response = try await fetch(with: request)
+        return response
+    }
 }
 
 extension ProfileService {
--- a/openapi/openapi.yaml
+++ b/openapi/openapi.yaml
@@ -5,9 +5,9 @@ info:
   description: Gravatar's public API endpoints
   contact:
     name: Gravatar
-    url: https://gravatar.com
+    url: 'https://gravatar.com'
 servers:
-  - url: https://api.gravatar.com/v3
+  - url: 'https://api.gravatar.com/v3'
     description: Production server
 tags:
   - name: profiles
@@ -16,6 +16,8 @@ tags:
     description: Operations about user avatars
   - name: qr-code
     description: Operations about QR codes
+  - name: experimental
+    description: Experimental operations that might be subject to change. Use with caution.
 components:
   headers:
     X-RateLimit-Limit:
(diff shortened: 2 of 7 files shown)


openapi/openapi.yaml
```diff
@@ -5,9 +5,9 @@ info:
   description: Gravatar's public API endpoints
   contact:
     name: Gravatar
-    url: https://gravatar.com
+    url: 'https://gravatar.com'
```
etoledom, line 8: Most of these changes are caused by a different JSON -> YAML converter.
The previous one used is now requiring subscription.

Should we use JSON directly? Yaml is easier to read, but we can have these problems.

Sources/Gravatar/Network/Services/ProfileService.swift
```diff
@@ -80,6 +80,23 @@ public struct ProfileService: ProfileFetching, Sendable {
...
+    ///   - props: The subset of data available for update. Only the provided fields will be updated.
+    ///   - token: Gravatar OAuth2 token.
+    /// - Returns: An asynchronously-delivered user profile with the fields updated.
+    public func updateProfile(with props: UpdateProfileRequest, token: String) async throws -> Profile {
+        var request = URLRequest(url: meProfileURL).settingAuthorizationHeaderField(with: token)
+        request.httpMethod = "PATCH"
+        request.httpBody = try? JSONEncoder().encode(props)
+        let response = try await fetch(with: request)
+        return response
+    }
```
pinarol, line 99: We need to wrap this with this so that we throw `APIError` like the other public functions.

```
        do {

        } catch {
            throw error.apiError()
        }
```

I believe we can benefit from using the new typed throws to avoid this kind of inconsistencies. But that's for another PR.

openapi/openapi.yaml
pinarol: Using JSON directly makes sense to me actually, since the only purpose of this file is documentation.

Sources/Gravatar/Network/Services/ProfileService.swift
pinarol: Oh well, I noticed another public function that misses this. :( I'll make a PR in the release branch. 

Edit: No I was confused, we are good in the release branch.

openapi/openapi.yaml
pinarol: Hold on. I remembered that we do some modifications in this file just before the stub generation so it's still good to observe those changes. So we can keep the yaml imo.

Sources/Gravatar/Network/Services/ProfileService.swift
etoledom: This is done by `fetch(with: request)` in this case. So the error returned is already an `APIError`.

> I believe we can benefit from using the new typed throws

💯

openapi/openapi.yaml
etoledom: > I remembered that we do some modifications in this file just before the stub generation

I don't remember this 🤔 
If you mean the access control modifier script, it runs on the already generated files. The yaml file at this stage is not needed.

etoledom: Anyway, we can decide this regardless of this PR 👍

Sources/Gravatar/Network/Services/ProfileService.swift
pinarol: Ooh ok I missed that.

openapi/openapi.yaml
pinarol: It's in the makefile

```
sed -i '' 's|components/schemas/Rating|components/schemas/AvatarRating|g' $(OPENAPI_DIR)/openapi.yaml
	sed -i '' 's| Rating:| AvatarRating:|g' $(OPENAPI_DIR)/openapi.yaml
```

We replace `Rating` as `AvatarRating` to avoid name conflicts.

pinarol (approved):
LGTM 👍

openapi/openapi.yaml
etoledom: Ooh I didn't know about this one 😬 
We can also adjust the make file to modify the JSON. But to think about it later 👍

Issue #712, opened 2025-04-21
QE About editing fields


etoledom:
Implement all missing About editing fields

Issue #711, opened 2025-04-21
Job Title field + improved form UI


etoledom:
This task includes adding:
- The job title field, 
- The different form sections (personal - professional), 
- The UI adornment around the field (box line) 
- Networking.
- Possibly also a new entry in editable fields setting.

PR #732, opened 2025-05-02
Add about info UI

### LINKED ISSUE #709: Display name form field + editable fields setting
This task adds the first form field, together with the setting option to choose which form fields are editable.

This task include 
- Display name form field
- Networking
- Save button
- The UI on the demo app to select the editable form fields, in both UIKit and SwiftUI

pinarol:
Closes #709 #711 #712

### Description

Adds the input fields and the ability to save changes.

### Testing Steps

Go to QuickEditor demo screens for both SwiftUI and UIKit
Select the "about" scope
Use the "Select input fields" button to select the fields you want to test
Open the QE 
Observe that the input fields are filled from the network
Change some fields and tap Save
Observe that the changes are reflected in the profile card (or check them via "View profile" button)

### DIFF
--- a/Sources/GravatarUI/QuickEditor/AboutInfoField.swift
+++ b/Sources/GravatarUI/QuickEditor/AboutInfoField.swift
@@ -49,6 +49,13 @@ public struct AboutInfoField: OptionSet, Sendable {
         .company,
     ]
 
+    var hasMultipleCategories: Bool {
+        (
+            (self.intersection(.personalFields).rawValue > 0 ? 1 : 0) +
+            (self.intersection(.professionalFields).rawValue > 0 ? 1 : 0)
+        ) > 1
+    }
+
     func localizedName() -> String {
         switch self {
         case .displayName:
--- a/Sources/GravatarUI/QuickEditor/AboutInfoField.swift
+++ b/Sources/GravatarUI/QuickEditor/AboutInfoField.swift
@@ -51,8 +51,8 @@ public struct AboutInfoField: OptionSet, Sendable {
 
     var hasMultipleCategories: Bool {
         (
-            (self.intersection(.personalFields).rawValue > 0 ? 1 : 0) +
-            (self.intersection(.professionalFields).rawValue > 0 ? 1 : 0)
+            (self.intersection(.personalFields).isEmpty ? 0 : 1) +
+                (self.intersection(.professionalFields).isEmpty ? 0 : 1)
         ) > 1
     }
 
--- a/Sources/GravatarUI/QuickEditor/AboutInfoField.swift
+++ b/Sources/GravatarUI/QuickEditor/AboutInfoField.swift
@@ -50,10 +50,10 @@ public struct AboutInfoField: OptionSet, Sendable {
     ]
 
     var hasMultipleCategories: Bool {
-        (
-            (self.intersection(.personalFields).isEmpty ? 0 : 1) +
-                (self.intersection(.professionalFields).isEmpty ? 0 : 1)
-        ) > 1
+        [
+            !self.intersection(.personalFields).isEmpty,
+            !self.intersection(.professionalFields).isEmpty,
+        ].hasMoreThanOneTrue
     }
 
     func localizedName() -> String {
--- a/Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
+++ b/Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
@@ -16,7 +16,9 @@ struct AboutEditorView: View {
         ScrollView {
             VStack(alignment: .leading, spacing: 0) {
                 personalInfoContent()
-                Spacer().frame(height: .DS.Padding.double)
+                if fields.hasMultipleCategories {
+                    Spacer().frame(height: .DS.Padding.double)
+                }
                 professionalInfoContent()
             }
             .padding(.DS.Padding.double)
@@ -41,7 +43,7 @@ struct AboutEditorView: View {
 
     @ViewBuilder
     private func personalInfoContent() -> some View {
-        if hasMultipleSections {
+        if fields.hasMultipleCategories {
             sectionHeader(title: Localized.personalSectionHeaderText)
         }
         if fields.contains(.displayName) {
@@ -81,7 +83,7 @@ struct AboutEditorView: View {
 
     @ViewBuilder
     private func professionalInfoContent() -> some View {
-        if hasMultipleSections {
+        if fields.hasMultipleCategories {
             sectionHeader(title: Localized.professionalSectionHeaderText)
         }
         if fields.contains(.jobTitle) {
@@ -141,11 +143,6 @@ struct AboutEditorView: View {
         }
         .padding(.vertical, .DS.Padding.single)
         .frame(maxWidth: .infinity)
-    }
-
-    private var hasMultipleSections: Bool {
-        (fields.contains(.personalFields) ? 1 : 0) +
-            (fields.contains(.professionalFields) ? 1 : 0) > 1
     }
 
     private enum Localized {
--- a/Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
+++ b/Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
@@ -9,7 +9,6 @@ struct AboutEditorView: View {
 
     @ObservedObject var model: AvatarPickerViewModel
     let fields: AboutInfoField
-    @StateObject private var inputFields: AboutInputFields = .init()
     @Environment(\.colorScheme) var colorScheme: ColorScheme
     @Environment(\.dynamicTypeSize) var dynamicTypeSize
 
@@ -32,7 +31,9 @@ struct AboutEditorView: View {
 
     private func saveButton() -> some View {
         Button {
-            print("")
+            Task {
+                await self.model.saveAboutInfo(for: fields)
+            }
         } label: {
             CTAButtonView(Localized.saveButtonTitle)
         }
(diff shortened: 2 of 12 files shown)


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -626,12 +592,11 @@ enum AvatarPicker {
...
     ]
     let selectedImageID = "5"
-    let profileModel = PreviewModel()
 
-    return AvatarPickerView<NoCustomEditor>(
+    AvatarPickerView<NoCustomEditor>(
         avatarImageModels: avatarImageModels,
         selectedImageID: selectedImageID,
-        profileModel: profileModel,
+        profileModel: nil,
```
pinarol, line 599: It's OK to remove this since we don't have the profile card inside `AvatarPickerView` anymore. So nothing to preview here even if this was not `nil`.

Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
```diff
@@ -0,0 +1,197 @@
+import SwiftUI
+
+struct AboutEditorView: View {
+    private enum Constants {
+        static let primaryFont: Font = .subheadline
+        static let sectionHeaderFont: Font = .subheadline.weight(.semibold)
```
pinarol, line 6: I am not sure if we should use smaller fonts here(like `.footnote`). The designs show 14pt but it doesn't correspond to a predefined type. I want to stick to the predefined ones. https://developer.apple.com/design/human-interface-guidelines/typography#Specifications

etoledom: Wouldn't `.headline` be a good option?
It's 14pt and semibold already 🤔 

CleanShot 2025-05-05 at 16 22 31@2x

etoledom: For the footer we don't have a 13pt preferred font, and I think it's ok if it's the `.footnote` preferred font at 12pt.

Sources/GravatarUI/QuickEditor/AboutInfoField.swift
```diff
@@ -49,6 +49,13 @@ public struct AboutInfoField: OptionSet, Sendable {
         .company,
     ]
 
+    var hasMultipleCategories: Bool {
+        (
+            (self.intersection(.personalFields).isEmpty ? 0 : 1) +
+                (self.intersection(.professionalFields).isEmpty ? 0 : 1)
+        ) > 1
+    }
```
etoledom: What about something like:
```swift
    var hasMultipleCategories: Bool {
        !self.intersection(.personalFields).isEmpty &&
            !self.intersection(.professionalFields).isEmpty
    }
```

etoledom (commented):
Looking good!

The behavior is all as expected 🎉 

I noticed a couple of design details that I belive could be improved:

---
When none of the fields is selected to be presented, the QE is presented completely empty. I think we could have a good default for this case, and not allow to present a completely empty QE. For example to show all the fields, or only name and about me fields.
CleanShot 2025-05-05 at 15 34 21

---
When scrolling of the About Me field, there's a padding where the text goes out of sight, which is smaller than the drawed frame. I'd expect that the frame si the edge.

This is a small detail though but it got my attention right away.

CleanShot 2025-05-05 at 15 32 14

On Android is similar but the padding is smaller and not so notorious:
(cc @AdamGrzybkowski )
CleanShot 2025-05-05 at 21 53 24@2x

Maybe we could ask a designer how important is this, but I believe that making the frame to be the edge would be ideal (maybe with a tinny drop shadow.

---

The main frame around the fields section is scrolling together with the scroll view, and the top and bottom edges of the frame are lost.

CleanShot 2025-05-05 at 22 02 31

I believe that the design intention is to make this frame to be the edge of the scrolling area:

(From figma)
CleanShot 2025-05-05 at 16 39 36@2x

I can't see the behavior in Android since this section is not yet scrollable (cc @AdamGrzybkowski again, sorry 😅 )

---

And finally just noting that the "3rd party" demo profile card doesn't get updated after the saving or closing the QE. 
If this is much more extra work, it's totaly good to have a different PR for this 👍

Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
pinarol: The default column is the middle one. I think you are looking at the xSmall column.

Screenshot 2025-05-06 at 13 52 41

pinarol:
> When none of the fields is selected to be presented, the QE is presented completely empty. I think we could have a good default for this case, and not allow to present a completely empty QE. For example to show all the fields, or only name and about me fields.

The default value in the public API is `.all`. If a developer manually deselects all fields, they're intentionally passing an empty array—which results in empty content, as expected. We shouldn’t try to override this behavior in ways that feel unintuitive. Personally, I’d be more surprised to see fields appear when I’ve explicitly passed an empty array—I’d likely assume it was a bug. This behavior is entirely under the developer’s control.

pinarol:
> When scrolling of the About Me field, there's a padding where the text goes out of sight, which is smaller than the drawed frame. I'd expect that the frame si the edge.

SwiftUI TextEditor does not let us set `insets`, we can only set a `padding`. And the padding will reduce the size of the scrollable area. So once a padding is set, that kind of space is inevitable. But we can get rid of the vertical padding and see how it looks, like:

```
                  .padding(.horizontal, .DS.Padding.single)
                  .padding(.vertical, 0)
```

Now the vertical space is zero. It looks a bit different from the designs but maybe it's not that bad.

Sources/GravatarUI/QuickEditor/AboutInfoField.swift
pinarol: That's a check for whether we have fields from "ALL categories" not just from "more than 1 category". It works because we have only 2 sections at the moment so ALL means 2 categories. But the logical meaning behind it is "ALL categories" and it would fail if we added a 3rd category. So I prefer keeping it as is, since it implements the "intended logic" logic better.

pinarol:
> The main frame around the fields section is scrolling together with the scroll view, and the top and bottom edges of the frame are lost.

That's the intended behavior. We want the button to be fixed at the bottom. 

> I believe that the design intention is to make this frame to be the edge of the scrolling area:

It already is the edge of the scrolling area? I think I didn't understand this one.

Edit: I think I got it. I pushed a new commit to put the borders outside the scrollable area. Let me know how it looks.

pinarol:
> And finally just noting that the "3rd party" demo profile card doesn't get updated after the saving or closing the QE.
If this is much more extra work, it's totaly good to have a different PR for this 👍

~~I was thinking about doing that on a separate PR since this one is blocking your side of the work.~~

Edit: It turned out to be a small change so I pushed it here.

Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
etoledom: Totally true! I was wondering how the numbers were so different 🤦‍♂️

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -245,15 +245,17 @@ class AvatarPickerViewModel: ObservableObject {
...
         do {
             let request = aboutInfoModel.updateProfileRequest(for: fields)
             let updatedProfile = try await profileService.updateProfile(with: request, token: authToken)
             self.profileResult = .success(updatedProfile)
+            return true
         } catch {
             // TODO: Handle errors properly.
             toastManager.showToast(error.localizedDescription, type: .error)
+            return false
         }
```
etoledom: What about we return the updated profile, and we pass it to the 3rd party app through the update handler?
This would avoid making an extra network call on their side.

Sources/GravatarUI/QuickEditor/AboutInfoField.swift
etoledom: Ok, I got it now 👍 

I think it's ok to leave it as it is, though I'm thinking how can we make it a bit easier to read at a glance.

What about something like this? 

```swift
    var hasMultipleCategories: Bool {
        hasMoreThanOneTrue([
            !self.intersection(.personalFields).isEmpty,
            !self.intersection(.professionalFields).isEmpty
        ])
    }

    func hasMoreThanOneTrue(_ bools: [Bool]) -> Bool { // Or in a [Bool] extension
        bools.count{ $0 } > 1
    }
```

etoledom (approved):
> <Text Editor Padding> Now the vertical space is zero. It looks a bit different from the designs but maybe it's not that bad.

Yes! I was referring to the verticals one, sorry for the confusion. Now it looks perfect 🙏 

> Edit: I think I got it. I pushed a new commit to put the borders outside the scrollable area. Let me know how it looks.

Yes! That's exactly what I meant. It looks very nice now, thank you for the update.

> The default value in the public API is .all. If a developer manually deselects all fields, they're intentionally passing an empty array—which results in empty content, as expected. 

It just feels wrong to me to allow this to happen. The dev might have a dynamic algorithm to choose the fields to display (who knows) and the empty page doesn't necessarily have to be an expected result. And if this happen, to the final user will look like a problem on Gravatar.

I'd at least add a warning somehow or some assertion to warn devs about opening the About editor completely empty.

This could be a broader conversation, so let's not block this PR before we decide 👍 

---

Nothing left that I'd consider blocking, so let's :shipit: at your discretion 🎉

pinarol:
> It just feels wrong to me to allow this to happen. The dev might have a dynamic algorithm to choose the fields to display (who knows) and the empty page doesn't necessarily have to be an expected result. And if this happen, to the final user will look like a problem on Gravatar.

Then it's a bug on the main app side and it should be fixed there. The `AboutInfoField` is an `OptionSet` of type `Int` so one can also pass a random value like `AboutInfoField(rawValue: 2344212343)`. The same case can be made here as well, like should we then show all the fields for example? We shouldn't I think.

Sources/GravatarUI/QuickEditor/AboutInfoField.swift
pinarol: Yes this looks nice. I'll update. 

(We are turning this to O(n) complexity from O(1) but with this tiny array size it should be OK.)

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
pinarol: Let's discuss this for both iOS and Android and make another PR if necessary. Keeping parity is important for this one.

PR #733, opened 2025-05-05
Adding buttons on the profile card to switch scopes

### LINKED ISSUE #713: "Avatar picker" / "About info" edition switch
Implement a way to switch context between Avatar picking and About info editing.

etoledom:
Closes #713 

### Description

This PR adds buttons to the profile card to switch scopes in the QE.

- A new scope has been added `. avatarAndAboutEditor `
- Added the buttons to the card to switch between avatar picker and about editor when this new scope is selected.
- Added the new scope to the demo app.

The layout configuration for the new scope is the same than the one for the Avatar picker, since it takes preference.
By default, the avatar picker is presented first. We can make this configurable.

https://github.com/user-attachments/assets/25534c66-272f-4314-8a76-874b2359285e

### Testing Steps

- Go to the Quick editor demos.
- Select the new scope `. avatarAndAboutEditor `.
  - Check that the switch between scopes, and each scope works as expected.
  - Check that the layout and configurations work as expected.


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -35,7 +35,7 @@ class AvatarPickerViewModel: ObservableObject {
     @Published private(set) var isAvatarsLoading: Bool = false
     @Published var avatarIdentifier: AvatarIdentifier?
     @Published var forceRefreshAvatar: Bool = false
-    @Published var profileModel: AvatarPickerProfileView.Model?
+    @Published var profileModel: AvatarPickerProfileView<EmptyView>.Model?
```
pinarol: We can just use `AvatarPickerProfileViewModel`

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerProfileView.swift
```diff
@@ -2,20 +2,8 @@ import Gravatar
...
-        var displayName: String
-        var location: String
-        var profileURL: URL?
-
-        var profileDetails: String? {
-            location.nilIfEmpty()
-        }
-    }
+struct AvatarPickerProfileView<AccessoryView>: View where AccessoryView: View {
+    typealias Model = AvatarPickerProfileViewModel
```
pinarol: No need for a typealias here imo, it makes things more confusing. We can just use `AvatarPickerProfileViewModel` directly where necessary.

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditorScopeOption.swift
```diff
@@ -42,6 +43,17 @@ public struct QuickEditorScopeOption {
...
     }
+
+    public static func avatarPickerAndAboutInfoEditor(
+        avatarPickerConfig: AvatarPickerConfiguration = .horizontalInstrinsicHeight,
+        aboutEditorFields fields: AboutInfoField = .all
+    ) -> Self {
+        .init(
+            scope: .avatarPickerAndAboutInfoEditor,
+            avatarPickerConfig: avatarPickerConfig,
+            aboutEditorConfig: .init(presentationStyle: .expandableMedium(), fields: fields)
```
pinarol: Instead of reusing `AvatarPickerConfiguration` and `AboutEditorConfiguration`, I think we should have a dedicated config for this mode. It's ambiguous which part is relevant this way, like, should I respect `aboutEditorConfig.presentationStyle` or `avatarPickerConfig.contentLayout`? It's also good for separating any future potential settings that will apply to only one mode. 

Something like: `AvatarAndAboutEditorConfiguration` or similar.

pinarol: I am also not a fan of initializing all the configs no matter what the scope is. If the scope is `.avatarPicker` then I shouldn't need to initialize other configs like `aboutEditorConfig`. I think we can still make use of associated enum objects **internally**  to avoid this. I'll try something. We don't need to keep this PR from getting merged though.

Sources/GravatarUI/SwiftUI/QuickEditorModalPresentationModifier.swift
```diff
@@ -123,6 +123,18 @@ extension ModalPresentationWithIntrinsicSize {
...
+            case .horizontal:
+                switch verticalSizeClass {
+                case .compact:
+                    false
+                default:
+                    true
+                }
+            case .vertical:
+                false
+            }
```
pinarol: This code repeats in the `.avatarPicker` case as well so we can reuse it by extracting to a separate function/block or sth.

pinarol (commented):
Looking good, just small comments.

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditorScopeOption.swift
etoledom: I agree, we can definitely improve this.
Initially I used optionals for the configs, but that was spreading too many optional checks everywhere.
As an internal implementation option, initializing everything works as long as every scope uses its corresponding config, but I see it can get confusing.

If we can improve this to avoid confusion and not allow for mistakes, that's great.

etoledom: In this particular case of multiple scopes though, I think is fine to initialise the configs of every used scope. 
We can use a third scope config publicly to avoid confusion, but internally we route to the actual config of each used scope to make easier the view's initialization using their corresponding config.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
etoledom: Ups, slipped from previous tests 👍

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditorScopeOption.swift
pinarol: I don't think it is fine. It creates ambiguity around what should be effective. Especially when multiple scopes are combined. Every used scope should have its own presentation style, when we re-use existing config objects then there are multiple presentation styles and we introduce an ambiguity around which presentation style should be used.

We already have one ambiguity currently: `aboutEditorConfig.presentationStyle` or `avatarPickerConfig.contentLayout`. Which one should be effective depends on an internal discussion. Our design can be self-explanatory instead.

pinarol: A nice way to avoid this type of ambiguities is to make use of associated values for each case. The enum can still remain internal.

pinarol:
Noticed one more thing:

We should allow for selecting input fields from demo apps for the `Avatar & About scope` as well.

Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditorScopeOption.swift
pinarol: I made a PR for this https://github.com/Automattic/Gravatar-SDK-iOS/pull/735

etoledom:
> Noticed one more thing:
> 
> We should allow for selecting input fields from demo apps for the `Avatar & About scope` as well.

Added the missing fields selector to the new scope in UIKit QE demo.

pinarol (approved):
👯 :shipit: 🎉

PR #734, opened 2025-05-07
Add snapshot tests for about info editor content

pinarol:
Closes BETS-33

### Description

Adds snapshot tests for about info editor content.

I added 2 types of snapshots. The ones that are named as "...IntrinsicHeight" shows the full content with intrinsic height so the scrolling won't be visible there. The other "fixedHeight" is the one with a normal phone height so we'll see a clipped content or empty space according to the content height.

### Testing Steps

CI green.


etoledom (approved):
Looks good!
CI 🍏 ✅

PR #735, opened 2025-05-07
Use associated values for scopes

pinarol:
Closes #

### Description

Proposal on: https://github.com/Automattic/Gravatar-SDK-iOS/pull/733

Updating the `QuickEditorScopeOption.Scope` to use associated values for configurations.

Also, we actually don't have to make config types public (like `AboutEditorConfiguration`, `AvatarPickerAndAboutEditorConfiguration` etc.). But I didn't want to touch that in this PR as I wasn't so sure.

### Testing Steps

Same with https://github.com/Automattic/Gravatar-SDK-iOS/pull/733


etoledom (approved):
Nice! Let's roll with this one 👍 
Thank you 🙏

PR #736, opened 2025-05-08
Adding alert when closing the editor with unsaved changes

### LINKED ISSUE #714: About editing: Detect unsaved changes
Implement a UX to save users from closing the QE with unsaved changes without noticing. (an alert, changing button titles, etc...)

etoledom:
Closes #714 

### Description

This PR adds an alert when the user tries to close the quick editor while there are unsaved changes.

The alert will be shown also in the "avatar & about" scope, while showing the avatar picker, if there are still unsaved changes in the about editor side.

CleanShot 2025-05-08 at 08 13 09

### Testing Steps

- Launch the demo app on the QE screen
- Select About Editor scope 
- Open the QE 
- Close the QE
 - It should close without warning 
- Open the QE and make some changes
- Close the QE 
  - An alert warning should appear
- Test both action buttons
- Repeat opening QE and making changes
- Save changes and close QE
  - No alert should be presented  
- Repeat opening the QE using the Avatar & About scope
- Make changes to the About section
- Switch to the Avatar picker section
- Close the QE
  - The alert should be presented.


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -381,25 +385,32 @@ class AvatarPickerViewModel: ObservableObject {
...
+            async let profile: () = fetchProfile()
 
-        // We need to await them otherwise network requests can be cancelled.
-        await avatars
-        await profile
+            // We need to await them otherwise network requests can be cancelled.
+            await avatars
+            await profile
+        }
     }
```
etoledom: This solves a problem where switching between avatar picker and about editor would remove the unsaved changes on the about editor side.

pinarol:
I can just swipe down and close the sheet without the alert, don't you think we need to disable that too if there are unsaved changes? I expect swiping down to be used way more often than tapping the Done button.

Sources/GravatarUI/SwiftUI/QuickEditorModalPresentationModifier.swift
```diff
@@ -24,6 +24,8 @@ struct QuickEditorModalPresentationModifier<ModalView: View>: ViewModifier, Moda
     @State private var presentationDetents: Set<PresentationDetent>
     @State private var prioritizeScrollOverResize: Bool = false
     @Environment(\.colorScheme) var colorScheme: ColorScheme
+    @State private var dismissAttempt: Bool = false
+    @State private var debounceWorkItem: DispatchWorkItem?
```
pinarol: Seem like, I forgot to remove this var `debounceWorkItem`. It's left from an experiment.

pinarol (approved):
Looks good 🎉 We can remove that variable and merge I think.

PR #737, opened 2025-05-08
Release 3.3.0

pinarol:
Closes #

### Description

Bumped up version to 3.3.0
Downloaded translations (We have a bit more translations than necessary for 3.3.0 but that should be OK 🤷 )

### Testing Steps

Smoke test the QE functionalities, try to cover some non-basic scenarios as well:

- Auth token provided externally
- Auth token internally created via OAuth 
- Network connectivity off
- Session expired
- Intro screen should be shown when the user is not logged in


andrewdmontgomery (approved):
👍

PR #738, opened 2025-05-08
Adding an initial loading state for the about editor

### LINKED ISSUE #715: About editing loading state
Implement an initial loading state for the about editing section.
It could be a shimmer skeleton or similar.

etoledom:
Closes #715 

### Description

This PR adds an initial loading indicator state to the About Editor View.

The loading indicator is the same type than the one used on the Avatar Picker.

CleanShot 2025-05-08 at 23 33 03

### Testing Steps

- Run the demo app.
- Open the Quick Editor demo screen.
- Select the About Editor scope.
- Present the QE.
  - Check that the loading indicator appears while the content is loading.
  - Check that the loading indicator will NOT appear while saving changes.


Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
```diff
@@ -14,6 +14,15 @@ struct AboutEditorView: View {
     var aboutUpdateHandler: (() -> Void)?
 
     var body: some View {
+        if model.profileResult == nil {
```
pinarol: We can't only use `profileResult == nil` condition here, no? We wouldn't want to show the loading indicator in case the service fails. We need to have a loading boolean. Can you check the logic on the avatar picker side? There, we avoid showing the loading indicator if the content is loaded once already.

Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
```diff
@@ -36,6 +45,15 @@ struct AboutEditorView: View {
             .padding(.bottom, .DS.Padding.double)
     }
 
+    private func loadingView() -> some View {
+        VStack {
+            ProgressView()
+                .progressViewStyle(CircularProgressViewStyle())
+                .controlSize(.regular)
```
pinarol: Can we add a top padding? 
                `.padding(.top, .DS.Padding.large)`

Currently it looks a bit too close to the top.
And it would match the avatar picker side also.

Can you also update AvatarPickerView's `private func avatarsLoadingView() -> some View {` to use `.padding(.top, .DS.Padding.large)` instead of `Spacer(minLength: .DS.Padding.large)`. Not sure why it was done that way in the first place. 🤔 We could even consider sharing the view.

Tests/GravatarUITests/AboutEditorViewTests.swift
```diff
@@ -114,6 +114,25 @@ struct AboutEditorViewTests {
         )
     }
 
+    @MainActor
+    @Test("Test about editor loading state snapshot")
+    func testLoadingState() async throws {
```
pinarol, line 119: It's good to have tests although I am not sure if the snapshot images can be generated consistently when there's an animating view in it. But anyway, we can consider saving in other forms if it becomes flaky.

Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
pinarol: Oh wait, does this hold the error object in case it fails? In that case it could work. But having a loading specific published variable is still the safer route imo for possible future changes.

Tests/GravatarUITests/AboutEditorViewTests.swift
etoledom: I had the same concern and runed it many times with a 100% success rate. 

> we can consider saving in other forms if it becomes flaky.

💯

Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
etoledom: Yes, `profileResult` is only nil once at the first load. But now I noticed that this is not good enought either.
We want the loading state againg when we retry after an error on the initial loading.

I've switches to use `isProfileLoading`. I had a confusion thinking that this would be true while saving, but that's not true and this property works well, and will be helpful also when retrying after an error 👍

Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
etoledom: I agree. With the `Separator` the behavior of the loading view was different in Avatars and in About screens, but now using padding they behave in the same way and we can safely share them 🎉

etoledom:
@pinarol - Thank you for the review. Ready for another look 👀

pinarol (approved):
Looks good! 🎉

PR #742, opened 2025-05-09
About Editor save button improvement

### LINKED ISSUE #739: About editor Save button improvements
This task includes:

- The Save Changes button remains disabled until any changes are made to the fields. 
- The Save Changes button shows a loading indicator during the save. 
  - Tapping should be disabled during this state.

etoledom:
Closes #739 

### Description

On this PR:

- Implement detection of unsaved changes in the About Editor.
- Fix an issue when unsaved changes were lost in the About Editor when switchings between About & Avatar picker.
- About Editor Save button is now disabled when:
  - There are no unsaved changes.
  - Save request is running.

CleanShot 2025-05-09 at 08 03 18

### Testing Steps

- Run the Demo app opening the QE demo screen
- Select the `About Editor` scope
- Open QE
  - Notice that the `Save` button should be disabled by default
- Make a change to any field
  - The `Save` button should now be enabled   
- Tap on Save
  - The button should be disabled with a loading indicator
  - When the network request finishes, the loading indicator should disappear and the button stays disabled.

### Test bug fix

- Select `Avatar & About` scope
- Go to About Editor
- Make some change to the fields without saving
- Switch to Avatar Picker and back
  - Check that the change is still there, and the button is active


Sources/GravatarUI/SwiftUI/AboutEditor/AboutEditorView.swift
```diff
@@ -7,6 +7,7 @@ struct AboutEditorView: View {
         static let footerFont: Font = .footnote
     }
 
+    @State var isLoading: Bool = false
```
pinarol: Can we make this @State var private? 

From Apple [docs](https://developer.apple.com/documentation/swiftui/state):

> Declare state as private to prevent setting it in a memberwise initializer, which can conflict with the storage management that SwiftUI provides:

pinarol: Can we give it a bit more specific name as it can be confused with the loading state for the data fetching? Like, `isSaving` or sth similar.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -37,11 +43,16 @@ class AvatarPickerViewModel: ObservableObject {
     @Published var forceRefreshAvatar: Bool = false
     @Published var profileModel: AvatarPickerProfileViewModel?
     @Published var aboutInfoModel: AboutInfoModel = .init()
+    private var updatedAboutInfoModel: AboutInfoModel = .init()
```
pinarol: Nit and totally optional: The naming here can be a bit confusing. `updatedAboutInfoModel` sounds like this is the user updated one while it's actually the fetched one from the backend. 🤔 I am also not sure we need a separate var for this. We could compare necessary fields with the existing `profileModel` as well. But I don't mind having it too.

pinarol (approved):
Looks good. 🎉
Leaving some small comments but I don't need to check again.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
etoledom: `profileModel` is `AvatarPickerProfileViewModel`, which has only 3 fields meant for the profile card.
It doesn't hold all the information necesarry to compare with `AboutInfoModel`. The name is a bit confusing, maybe we could rename to `profileCardViewModel`.

But I like this idea, and we can use the `Profile` instance stored in the `profileResult` variable.
Seems to work well 🎉

PR #746, opened 2025-05-15
About editor error handling

### LINKED ISSUE #710: About editing Error Handling
- Presenting a toast with an error message.

- UI state for error if needed.

- Initial loading error state.

etoledom:
Closes #710 

### Description

This PR adds error handling to the About Editor scopre:

- Initial loading error view are shared with Avatar Picker:
  - Session expired
  - Network error
  - Generic errors 

- Save error toast message is shared with avatar selection toast errors.

#### Initial loading:

- Test presenting the Quick Editor without internet in all 3 scopes
  - Test reloading the content after establishing internet connection, and switching modes in `Avatar & About` scope.
- Test presenting the Quick Editor with a wrong (or expired) token in all 3 scopes

Expected result:
- The related error view should be presented in both About Editor and Avatar Picker scopes and modes.
- When the messages mention `... update your avatar` in the Avatar Picker, it should say `... update your Profile` in About Editor scope.

*Note:* I believe that `update your profile` will be more generic across multiple future possible scopes, so I have chosen this one to be the default.

#### Saving error:
- Test saving about info changes without internet connection
  - Check that the toast appears as expected


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -71,20 +71,19 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
...
                         }
+
                         Spacer()
                             .frame(height: Constants.vStackVerticalSpacing)
                     }
                     .accumulateIntrinsicHeight()
                 }
-                .task {
-                    model.refresh(modelToRefresh: .avatarPickerModel)
-                }
```
etoledom, line 87: This is now triggered by QuickEditor. It will initially refresh all models.
This avoids re-refreshing every time we switch between Avatar and About in the mixed scope.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -370,6 +363,16 @@ class AvatarPickerViewModel: ObservableObject {
         self.gridResponseStatus = .failure(error)
     }
 
+    func showToast(for error: Error) {
```
pinarol: Since this method is only useful for avatar update errors (because it fallsback to Localized.avatarUpdateFail) we should either rename it or maybe make it an inner function of `postAvatarSelection(...)`

pinarol (commented):
Looks good overall, just a few things:

1. If the session is expired on `Save` button tap then we should show the session expired error rather than the toast. Currently it just shows a generic error "Ooop, something didn't quite work out bla bla". The avatar picker side shows the session expired error on an attempt to avatar change which is much better since at that moment the error is unrecoverable. 

2. Let's also add the profile update success toast as part of this PR, shall we? It should be in figma.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerViewModel.swift
etoledom: Oh good catch! I'll make it generic

etoledom:
@pinarol - 
1- Done ✅  - Good catch
2- Done ✅ 
3- I've made it generic to be able to reuse it

Ready for another look 👀

pinarol (approved):
Looks great 🎉

PR #748, opened 2025-05-15
Returning Profile after an update on about editor

### LINKED ISSUE #747: About Editor: Return Profile on about info update handler
The QuickEditor updateHandler should return the Profile model when the About Editor updates the profile.

etoledom:
Closes #747 

### Description

On this PR we add the Profile model to the about editor updateHandler callback.

I've decided to use the same strategy than [on Android](https://github.com/Automattic/Gravatar-SDK-Android/pull/629). This seems to be the simplest way to return the model only on a specific update type, without using enum associated values.
Which is to use a protocol, and concrete types for each type of update. And casting the protocol in a switch statement on the implementation side.

Enclosing the update structs in a `public enum QuickEditorUpdate` will help with discoverability of the options available:

CleanShot 2025-05-15 at 16 46 14@2x

### Testing Steps

- Open the Quick Editor in UIKit, Avatar Picker scope.
- Update the Avatar.
  - Check that the 3rd party avatar is updated
- Open the Quick Editor on About Editor scope 
- Update the display name 
  - Check that the display name on the 3rd party app side is updated.
- Repeat on SwiftUI side.


Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditorUpdate.swift
```diff
@@ -0,0 +1,32 @@
...
+///     case is QuickEditorUpdate.Avatar:
+///         // Handle avatar update
+///     case let update as QuickEditorUpdate.AboutInfo:
+///         // Handle profile update using `update.profile`
+///     default:
+///         break
+///     }
+/// }
+/// ```
+public protocol QuickEditorUpdateType {}
```
pinarol, line 20: `QuickEditorUpdateType` is only confirmed but it is not used from anywhere, which makes it dead code.

If we had such an API for example: 
```
func update(type: QuickEditorUpdateType) { ... }
```
it would have a purpose then. 

But currently we don't need it, so I think we should just remove it. 

You already defined `QuickEditorUpdate` which will act as a namespace/catalog for all the possible types. It is sufficient.

pinarol: Hold on, we do use `QuickEditorUpdateType`, it just doesn't appear in this PR's diff because it's an old type. Ok ignore my comment.

pinarol (approved):
Looks good 🎉

PR #750, opened 2025-05-15
Add detent calculations for about and avatar scope

pinarol:
Closes BETS-34

### Description

Adds detent calculations for about and avatar scope.
Adds intrinsic height calculations for the About editor.

**Some details:**

We still use `AvatarPickerContentLayout` to define the detents.

The difference for the about mode is that, if the content height is smaller than our default height, we use the intrinsic height directly in the sheet. If it's bigger, then we use 2 detents, one at the default height(same with avatar picker height) the other at the `.large`.

We can experiment with other ideas as well.

### Testing Steps

Test in both SwiftUI and UIKit demo pages
Test all content layout options and try with different number of about fields.

Let me know how it looks and fields.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/GravatarUI/QuickEditor/QuickEditorConfiguration.swift
```diff
@@ -42,15 +42,23 @@ public struct AboutEditorConfiguration: Sendable {
 
 /// Configuration which will be applied to the scope which allows switching between Avatar Picker and About info editor.
 public struct AvatarPickerAndAboutEditorConfiguration: Sendable {
+    public enum Mode: Sendable {
```
pinarol: I just moved this to here since it came natural to me but we can discuss different locations and also names.

Sources/GravatarUI/SwiftUI/QEDetent.swift
```diff
@@ -27,14 +28,51 @@ enum QEDetent {
...
+                switch multipleEditorMode {
+                case .avatarPicker:
+                    .init([.height(intrinsicHeight)])
+                case .aboutEditor:
+                    if intrinsicHeight >= QEModalPresentationConstants.bottomSheetEstimatedHeight {
+                        .init([.height(QEModalPresentationConstants.bottomSheetEstimatedHeight), .large])
+                    } else {
+                        .init([.height(intrinsicHeight)])
+                    }
+                }
```
pinarol, line 62: This is where we can update to try out different detent behaviors for the about mode of the avatar&about editor.

Sources/GravatarUI/QuickEditor/QuickEditorConfiguration.swift
pinarol: Update: We'll tackle this on a separate PR and I'll update this PR afterwards.

Sources/GravatarUI/SwiftUI/QEDetent.swift
```diff
@@ -27,14 +28,53 @@ enum QEDetent {
...
+        verticalSizeClass: UserInterfaceSizeClass?,
+        currentPage: QuickEditorPage
+    ) -> [QEDetent] {
+        switch presentation {
+        case .horizontal:
+            if verticalSizeClass == .compact {
+                // in landscape mode where the device height is small we display the full size sheet(which is
+                // also the default value of the detent).
+                .init([.large])
+            } else if case .avatarPickerAndAboutInfoEditor = scopeOption.scope {
```
etoledom: What's the reason for this scope check?
The check is already done by the caller 🤔

pinarol: No good reason, just confusion it seems. I'll update.

pinarol: Done.

etoledom (approved):
Looks good and works great 🎉

PR #752, opened 2025-05-19
Avoid scrolling of frame on Avatar Picker

### LINKED ISSUE #740: Update the Avatar Picker design to match About Editor scrolling frame
In the About Editor, the frame line is surrounding the scrollable area, but it doesn't scroll together with it.

We want to copy this behavior in the Avatar Picker

etoledom:
Closes #740 

### Description

Small PR which prevents the frame hearline around the Avatar Picker from scrolling together with the content.

CleanShot 2025-05-19 at 08 59 12

### Testing Steps

- Open the QE on horizontal layout mode
- Scroll the content.
  - Check that the hairline frame does not scroll together with the content.


Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -131,6 +131,13 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
                     Button(Localized.dismissButtonTitle, role: .cancel) {}
                 }
             }
+            // Display the frame outside of the scroll view around the content, but not around the loading or error states.
+            .if(!model.isAvatarsLoading && !model.grid.isEmpty) { content in
+                content
+                    .avatarPickerBorder(colorScheme: colorScheme)
+                    .padding(.horizontal, .DS.Padding.double)
+                    .padding(.bottom, .DS.Padding.medium)
```
pinarol: Unfortunately we can't just add a vertical padding outside of intrinsic height calculations. This will not be taken intro consideration when accumulating the intrinsic size thus will lead to wrong value. 

You can observe this by using a different color in accumulateIntrinsicHeight(), thus painting views to see which are taken into consideration. Things we can check:

- content of the scroll view should be equal to the scroll view height. If it scrolls, it means our intrinsic height is calculated to be less than it really is.
- every view needs to be painted in the vertical axis.

Currently the scrollview scrolls because of the `.DS.Padding.medium` we add here without accumulating it. We need to find a different way of doing this. 🤔

etoledom: Trying with Spacer now, to be able to accumulate it's height also.

Sources/GravatarUI/SwiftUI/AvatarPicker/AvatarPickerView.swift
```diff
@@ -131,7 +131,17 @@ struct AvatarPickerView<ImageEditor: ImageEditorView>: View {
                     Button(Localized.dismissButtonTitle, role: .cancel) {}
                 }
             }
-
+            // Display the frame outside of the scroll view around the content, but not around the loading or error states.
+            .if(!model.isAvatarsLoading && !model.grid.isEmpty) { content in
+                VStack {
```
pinarol: Can we add a `spacing: 0` here? It's not always necessary but sometimes the system just puts some default spacing when `nil` is passed and the rules of when that happens is not clear. In the past I had difficult time debugging the intrinsic height problems just because the system added some 8 pt spacing by default.

At the moment there's no system added spacing here so we are currently fine (probably because we use a Spacer in it?). Just suggesting it as something that can save us in the future if the contents change.

pinarol (approved):
Looks good. Just 1 nit and we can 🚢 No need for me to re-check. :shipit:

PR #753, opened 2025-05-19
Implement about detents

pinarol:
Closes BETS-47

### Description

Defines a generic `SheetPresentationStyle` and uses it to define "About" scope detents.

**intrinsicHeight**: Displays sheet based on the content height.
**automatic**: If the content height is smaller than the default sheet height then it applies intrinsicHeight, otherwise it applies 2 detents: default sheet height + large.
The rest is as usual.

### Testing Steps

Test with both UIKit and SwiftUI demo apps.
Select "About" scope
Pick different number of fields.
Pick different presentation styles.
Display the QE and check if anything looks odd.

### DIFF
--- a/Sources/GravatarUI/QuickEditor/SheetPresentationStyle.swift
+++ b/Sources/GravatarUI/QuickEditor/SheetPresentationStyle.swift
@@ -21,10 +21,14 @@ public struct SheetPresentationStyle: Sendable {
         .init(detentMode: .large)
     }
 
-    /// Medium height sheet that is expandable to full height. In compact height this is inactive and the sheet is displayed as full height.
-    /// - initialFraction: The fractional height of the sheet in its initial state.
-    /// - prioritizeScrollOverResize: A behavior that prioritizes scrolling the content of the sheet when
-    /// swiping, rather than resizing it. Note that this parameter is effective only for iOS 16.4 +.
+    /// A medium-height sheet presentation style that can expand to full height.
+    ///
+    /// In a compact height size class environment, the medium-height presentation is disabled and the sheet is presented at full height by default.
+    /// - Parameters:
+    ///   - initialFraction: The fractional height (relative to the full height) at which the sheet is initially presented.
+    ///   - prioritizeScrollOverResize: Determines whether scroll gestures within the sheet take precedence over resizing gestures. Available on iOS 16.4 and
+    /// later.
+    /// - Returns: A configured ``SheetPresentationStyle`` instance representing an expandable medium-height sheet.
     public static func expandableMedium(
         initialFraction: CGFloat = SheetPresentationStyle.expandableMediumInitialFraction,
         prioritizeScrollOverResize: Bool = false
--- a/Sources/GravatarUI/QuickEditor/SheetPresentationStyle.swift
+++ b/Sources/GravatarUI/QuickEditor/SheetPresentationStyle.swift
@@ -15,7 +15,8 @@ public struct SheetPresentationStyle: Sendable {
 
     let detentMode: DetentMode
 
-    /// Full height sheet.
+    /// A full-height sheet presentation style.
+    /// - Returns: A `SheetPresentationStyle` instance configured to present the sheet at full height.
     public static func large() -> SheetPresentationStyle {
         .init(detentMode: .large)
     }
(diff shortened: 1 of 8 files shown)


Sources/GravatarUI/QuickEditor/SheetPresentationStyle.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
+
+public struct SheetPresentationStyle: Sendable {
```
pinarol, line 4: Maybe this naming is too generic, but the options also feel quite generic and decoupled from the about editor. 🤔

etoledom: Looks good to me! Ideally we reuse it for most future scopes 👍

Sources/GravatarUI/QuickEditor/SheetPresentationStyle.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// Full height sheet.
+    public static func large() -> SheetPresentationStyle {
+        .init(detentMode: .large)
+    }
+
+    /// Medium height sheet that is expandable to full height. In compact height this is inactive and the sheet is displayed as full height.
+    /// - initialFraction: The fractional height of the sheet in its initial state.
+    /// - prioritizeScrollOverResize: A behavior that prioritizes scrolling the content of the sheet when
+    /// swiping, rather than resizing it. Note that this parameter is effective only for iOS 16.4 +.
+    public static func expandableMedium(
```
etoledom: We could specify `compact height size class`.

```suggestion
    /// A medium-height sheet presentation style that can expand to full height.
    ///
    /// In a compact height size class environment, the medium-height presentation is disabled and the sheet is presented at full height by default.
    /// - Parameters:
    ///   - initialFraction: The fractional height (relative to the full height) at which the sheet is initially presented.
    ///   - prioritizeScrollOverResize: Determines whether scroll gestures within the sheet take precedence over resizing gestures. Available on iOS 16.4 and
    /// later.
    /// - Returns: A configured ``SheetPresentationStyle`` instance representing an expandable medium-height sheet.
    public static func expandableMedium(
```

Sources/GravatarUI/QuickEditor/SheetPresentationStyle.swift
```diff
@@ -0,0 +1,65 @@
...
+            initialFraction: CGFloat = SheetPresentationStyle.expandableMediumInitialFraction,
+            prioritizeScrollOverResize: Bool = false
+        )
+        case intrinsicHeight
+        case automatic(prioritizeScrollOverResize: Bool = false)
+    }
+
+    let detentMode: DetentMode
+
+    /// Full height sheet.
```
etoledom:
```suggestion
    /// A full-height sheet presentation style.
    /// - Returns: A `SheetPresentationStyle` instance configured to present the sheet at full height.
```

Sources/GravatarUI/QuickEditor/SheetPresentationStyle.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// Represents a bottom sheet with intrinsic height.
+    ///
+    /// There are 2 size classes where this mode is inactive:
+    ///  - Compact height: The sheet is displayed in full height.
+    ///  - Regular width: The system ignores the intrinsic height and defaults to a full size sheet by the system.
+    public static func intrinsicHeight() -> SheetPresentationStyle {
+        .init(detentMode: .intrinsicHeight)
+    }
+
+    /// Applies `.intrinsicHeight` when the content height is below a defined threshold. Otherwise, applies `.expandableMedium()`.
```
etoledom:
```suggestion
    /// Returns a sheet presentation style that automatically chooses between `.intrinsicHeight` and `.expandableMedium()`,
    /// depending on the content height.
    ///
    /// If the content height is below a predefined threshold, `.intrinsicHeight` is used;
    /// otherwise, `.expandableMedium()` is applied.
    ///
    /// - Parameter prioritizeScrollOverResize: Determines whether scroll gestures within the sheet take precedence over resizing gestures.
    ///   This parameter is effective only on iOS 16.4 and later.
    ///
    /// - Returns: A `SheetPresentationStyle` instance that adapts the sheet height based on content size.
```

Sources/GravatarUI/QuickEditor/SheetPresentationStyle.swift
etoledom:
```suggestion
/// A value that describes how the Quick Editor sheet should be presented,
public struct SheetPresentationStyle: Sendable {
```

etoledom (approved):
Looking great and works perfect 🎉 

I worked out a bit the inline docs thinking on the public generated documentation.
Feel free to merge, modify, or dismiss 👍

pinarol:
Great suggestions 🎉

PR #754, opened 2025-05-19
Remove the rogue button

pinarol:
Closes GRA-79

### Description

Removing the rogue button. It causes weird background color in the profile card in dark mode because mode we use a background color with opacity there.

Also, the rogue button strengthened the profile card's shadow somehow so removing it turns the shadow back into its original state. So the snapshots needed update.

### Testing Steps

Check the profile card in dark mode
Observe that the weird background color is gone.


etoledom (approved):
Oh this looks like an unintended left over. Good catch! 🙏

PR #755, opened 2025-05-20
Update profile card background in dark mode

pinarol:
Closes GRA-105

### Description

`colorInvert()` doesn’t work properly in snapshot tests.  It must be requiring some sort of real rendering engine that’s only available in the real app. Thus, we are updating it to `secondarySystemBackground`. Here's how it looks:

<table>
  <tr>
    <th>Old</th>
    <th> New(full opacity) </th>
    <th> New(0.9 opacity)</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
    <td></td>
  </tr>
</table>

### Testing Steps

Check out QE's profile card background color in dark mode and in snapshot tests.


etoledom (approved):
Full opacity looks to me exactly as before, and now the snapshots represent better the UI 🎉

PR #756, opened 2025-05-20
Hide the profile card on compact height and keyboard present

### LINKED ISSUE #741: About Editor adaptation to landscape
We need to find a way to adapt the vertical space of the About Editor in landscape mode, since, specially for small devices, there's not enough space to make it a comfortable experience.

Posible solutions:
- Hide the email and profile card on landscape.
- Include the email and profile card in the scrollable area when in landscape.

Image

etoledom:
Closes #741 

### Description

Small PRs which will hide the Profile card in the case the available height is compact and the keyboard is present.

CleanShot 2025-05-20 at 16 25 40

### Testing Steps

- Open the Quick Editor with About Editor scope
- Start editing any field
- Rotate the device to make it landscape
  - Check that the profile view is not present
  - Check that there's enough space to see the fields which is being edited.


etoledom:
I'm considering also hiding the save button to have a clear view of the field (specially for the multiline one) 🤔

pinarol:
> I'm considering also hiding the save button to have a clear view of the field (specially for the multiline one) 🤔

We can definitely try.

pinarol (approved):
Looks good. Let me know if you want me to check again in case you also hide the Save button.

etoledom:
@pinarol  - I've added a commit to also hide the save button. It looks like this:

CleanShot 2025-05-20 at 20 07 07

pinarol:
That looks good to me as well.

PR #760, opened 2025-05-21
Adding bottom padding to error view on About editor

etoledom:
Closes BETS-48 

### Description

Adding padding under error view on the about editor.

Now both avatar and about error views have the same space at the bottom

I've chosen to add it only under the error view, as the content already looks well laid out

### Testing Steps

- Run the demo app on an old iPhone without notch (iPhone SE)
- Go to the Quick Editor demo screen
- Insert an invalid auth token (to force auth error)
- Open the Quick Editor with About Editor scope
  - Check that there's a space between the lower frame of the error and the screen.


pinarol (approved):
Looks good 👍

PR #763, opened 2025-05-27
UIKit sheet presentation from SwiftUI context

etoledom:
Closes BETS-50

### Description

This PR is aimed to replace the native SwiftUI sheet presentation with the UIKit sheet presentation.

Wins from this changes are:

- Native transition animation on SwiftUI context when the sheet height changes.
- Native detection of dismissal attempt by the user.
- Half and full detents support for pre-iOS16 

### To Achieve this the follow changes were made:

`QuickEditorViewController` is now used by both UIKit and SwiftUI contexts. It wraps `QuickEditor` view in a `UIHostingController` to be presented by UIKit framework.

#### From UIKit context:
Not many changes. It uses the public `QuickEditorPresenter` to present `QuickEditorViewController` in a bottom sheet.

A change was made where the `CustomImageEditorControllerRepresentable` conversion is handled by `QuickEditorPresenter` directly, in order to remove UIKit dependencies from the `QuickEditorViewController` init parameters.

#### From SwiftUI context:

There's a new `QuickEditorBottomSheetPresenterViewController` which will present modally (on a sheet) the`QuickEditorViewController`. This part is the trick to do the presentation work.

Then `QuickEditorBottomSheetPresenterViewControllerRepresentable` wraps it in a Representable, to be able to be inserted into the SwiftUI hierarchy.

Then, `QuickEditorBottomSheetPresenterViewControllerRepresentable` is initialised with the properties given by the public `View.gravatarQuickEditorSheet` extension, and passed to a new view modifier `QuickEditorBottomSheetViewControllerPresentationModifier`.

The new `QuickEditorBottomSheetViewControllerPresentationModifier` modifier will conditionally add or remove the presenter controller to the hierarchy according to `isPresented` property. And it will ensure that its dimensions are zero to keep it hidden. Every time the controller is added to the hierarchy, it will automatically present the QuickEditorController in a sheet presentation.

#### From SwiftUI context:

- Native transition animation on SwiftUI context when the sheet height changes.
- Native detection of dismissal attempt by the user.
- Custom image editor injection.
- System color scheme, and user specified color scheme.
- Update callbacks to parent app.
- All different presentation styles.

#### From UIKit context:

Keen eye to:
- Custom image editor injection.
- System color scheme, and user specified color scheme.

As the rest has been mostly kept as it was before, but still good to smoke test everything else.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/QuickEditorModalPresentationModifier.swift
+++ b/Sources/GravatarUI/SwiftUI/QuickEditorModalPresentationModifier.swift
@@ -13,108 +13,19 @@ enum QEModalPresentationConstants {
     static let bottomSheetMinHeight: CGFloat = 350
 }
 
-@available(iOS 16.0, *)
-struct QuickEditorModalPresentationModifier<ModalView: View>: ViewModifier, ModalPresentationWithIntrinsicSize {
-    fileprivate typealias Constants = QEModalPresentationConstants
-
+struct QuickEditorBottomSheetViewControllerPresentationModifier: ViewModifier {
     @Binding var isPresented: Bool
-    @State private var isPresentedInner: Bool
-    @State private var sheetHeight: CGFloat = Constants.bottomSheetEstimatedHeight
-    @State private(set) var verticalSizeClass: UserInterfaceSizeClass?
-    @State private var presentationDetents: Set<PresentationDetent>
-    @State private var prioritizeScrollOverResize: Bool = false
-    @Environment(\.colorScheme) var colorScheme: ColorScheme
-    @State private var dismissAttempt: Bool = false
-    @State private var currentPage: QuickEditorPage
-
-    let onDismiss: (() -> Void)?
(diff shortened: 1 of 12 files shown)


Sources/GravatarUI/SwiftUI/QuickEditorModalPresentationModifier.swift
```diff
@@ -13,108 +13,20 @@ enum QEModalPresentationConstants {
...
-                        dismissAttempt: $dismissAttempt,
-                        isPresented: $isPresented,
-                        isLargeDetentOnly: presentationDetents == [.large]
-                    )
-                    .environment(\.dismissAttempt, dismissAttempt)
+        content.if(isPresented) { content in
+            ZStack {
+                content
+                quickEditorPresenter
+                    .frame(width: 0, height: 0)
```
pinarol, line 26: So this is the trick! 🪄

pinarol (approved):
Masterpiece 🎉

PR #764, opened 2025-05-27
Move headers from URLSession level to request level

pinarol:
Closes #

### Description

Currently we don't add headers in case a custom `URLSession` is injected. We might be missing some usage analytics because of that. This PR proposes to change the point we add these headers so they are not overridden/lost.

### Testing Steps

Use a sniffer tool to check that the headers are added both for upload and other operations.


etoledom (approved):
Tested a custom URLSession passed to the Profile service:

- On trunk it won't add the `x-` headers
- With these changes, they are added 

:shipit: 🙏

PR #767, opened 2025-05-28
Fix vertical padding above the vertical avatar grid

pinarol:
Closes BETS-51

No space above the avatars grid:

### Description

Adding space so the "after" looks like this:

### Testing Steps

Check the spacing above the vertical avatar grid in the QE


etoledom (approved):
Perfect

PR #768, opened 2025-05-29
Disable fields on save

etoledom:
Closes GRA-124

### Description

This PR is meant to disable the About editor fields while the save request is ongoing.

I tried changing the style of the fields for a visual feedback but there's an issue with the text `TextEditor` on pre-iOS16

To be able to remove this white background from the text on pre-iOS16, we need to call:
`UITextView.appearance().backgroundColor = .clear`

Which doesn't seem good to do from an external library.

### Testing Steps

- Open the Quick Editor with About editor scope
- Save some changes
  - Check that the fields are disabled while the request is ongoing.


pinarol (approved):
LGTM 🎉

PR #776, opened 2025-06-03
Disable unwanted scroll bouncing

etoledom:
Closes BETS-52

### Description

This PR is aimed to remove the unwanted bouncing of the Avatar Picker when it's on horizontal layout.

CleanShot 2025-06-03 at 20 16 10

### Testing Steps

- Run the demo app
- Go to Quick Editor demo screen
- Present the quick editor with Avatar Picker scope in horizontal layout.
  - Check that there is no bouncing on the avatars section. 
  - Check that scroll is enabled on landscape.
- Smoke-test other layouts and configurations.


pinarol:
Do you notice a difference when swiping down the sheet? It doesn’t react immediately when you drag down, and then when it does react, it jumps. 


https://github.com/user-attachments/assets/d2f7ac5f-f997-4f8b-aff1-6fc19cb7b384

pinarol (approved):
Seems like my issue is related to number of avatars and not this change. So let’s 🚀

Issue #675, opened 2025-03-21
Accessibility: Description blocks


etoledom:
It's good to group blocks of information as one single accessibility element. This makes navigating the simpler and easier to reason with:

<img src="https://github.com/user-attachments/assets/0d93ac60-ca7b-46b5-b8b9-0d12a84ef5b5" width="40%">

These two sections described on the screenshot could be grouped and read as one element.

- "John Appleseed. - Location - . Double tap to open profile on a web view.

- "Avatars (or "Your avatars" I like more in this case). Description".

PR #779, opened 2025-06-06
Accessibility mark iii

etoledom:
Closes https://github.com/Automattic/Gravatar-SDK-iOS/issues/675

### Description

This PR makes the page switch buttons on the profile card accessible with a label for Voice Over.
- Note that the design itself is not good for accessibility, and grouping all the card together will make it difficult for the system to read out the image description. So I decided to go with the simpler route and name the buttons appropriately, even with the confusion of the position change for voice over users.

We also mix the textfield title and value together as one accessibility element in the About Editor page, which is the recommendation for these cases.

### Testing Steps

- Run the demo app
- Go to Quick Editor demo
- Present the quick editor with the Avatar & About scope
- Turn ON Voice Over
  - Check that Voice Over reads the label on the page switch buttons.
  - Check that on the About Editor page, Voice Over reads the title and the fields content together, and does not interact with the title label.

### DIFF
--- a/Demo/Demo/Gravatar-Demo/DemoQuickEditorViewController.swift
+++ b/Demo/Demo/Gravatar-Demo/DemoQuickEditorViewController.swift
@@ -48,7 +48,7 @@ final class DemoQuickEditorViewController: BaseFormViewController {
     )
 
     lazy var initialPageButton = ButtonLabelField(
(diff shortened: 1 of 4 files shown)


pinarol (approved):
Looks and works good 🎉

PR #781, opened 2025-06-12
Remove profile card on iPhone SE 1st Gen with keyboard enabled

etoledom:
Closes GRA-179

### Description

This PR address a problem where the About fields wouldn't be visible on an iPhone SE 1st gen with the keyboard enabled.
Now the Profile card is removed on this case.

**Note:** We have decided to also include other iPhone SE and iPhone 8 to this logic.

CleanShot 2025-06-12 at 13 03 17@2x

### Testing Steps

- Run the demo app on an iPhone SE 1st gen 
- Open the Quick Editor on About Editor scope
- Edit a field with the keyboard present
  - Check that the field is visible


pinarol (approved):
Looks good 🎉

PR #782, opened 2025-06-13
Requested changes from PR review

etoledom:
Closes GRA-193

### Description

- This PR removes the email from the Quick Editor header when the token is given externally (not using internal OAuth flow).
- Changing the arrow icon from the `View profile` button to an external link icon.
- The Profile card maximum text size is `.xxxl` and it won't grow for accessibility sizes.
- We have removed the profile card for text sizes of `.accessibility3` and more.
- On Avatar & About scope, the default height of the About editor is now intrinsic height.

### Testing Steps

- Run the demo app
- Navigate to the Quick Editor demo screen
- Open the Quick Editor with the token field blank
  - Check that the email is visible
- Close the Quick Editor  
- Add a valid token to the token field
- Open the Quick Editor
  - Check that the email is not visible
- Check that the View Profile button is displaying the new external link icon instead of the arrow.


pinarol (approved):
Looks good 🎉

PR #784, opened 2025-06-13
Using device orientation to hide profile card

etoledom:
Closes GRA-194

### Description

This PR fixes a problem on iPad mini when the device is rotated in Landscape mode.
Now the About fields are visible after removing the Profile view on landscape, when the keyboard is present.

image

### Testing Steps

- Run the demo app on the iPad mini
- Present the Quick Editor on About Editor scope
- Edit a field
- Turn to Landscape
  - Check that the Profile view is hidden, and the fields are visible.

### DIFF
--- a/Sources/GravatarUI/SwiftUI/DeviceOrientationPropertyWrapper.swift
+++ b/Sources/GravatarUI/SwiftUI/DeviceOrientationPropertyWrapper.swift
@@ -1,0 +1,47 @@
+import SwiftUI
+import Combine
+
+enum Orientation {
+    case landscape
+    case portrait
+    case unknown
+}
+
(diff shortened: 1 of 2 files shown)


Sources/GravatarUI/SwiftUI/DeviceOrientationPropertyWrapper.swift
```diff
@@ -0,0 +1,48 @@
+import Combine
+import SwiftUI
+
+enum Orientation {
+    case landscape
+    case portrait
+    case unknown
+}
+
+@propertyWrapper
```
pinarol, line 10: Nice!

pinarol (approved):
Elegant solution and it works good 🎉

etoledom:
Thank you!

PR #791, opened 2025-06-25
Update openapi spec

etoledom:
Closes GRA-220

### Description

Updating the OpenApi spec to bring UserID and User Login fields needed on Gravatar App.
These new fields are only present for the Authenticated Profile fetch request.

The Profile fetch demo screen has been updated with an optional token field to make this request and see the new fields:

### Testing Steps

- Run the demo app
- Go to Fetch Profile demo screen
- Paste a valid OAuth2 access token into the token field
- Fetch the profile
  - Check on the results that the userId and userLogin fields are returned as expected


pinarol (approved):
I see the new fields. LGTM 👍

PR #792, opened 2025-07-01
Use ProfileID instead of Email for image upload option

etoledom:
Closes GRA-439

### Description

On this PR we deprecate `AvatarSelection` in favor of `AvatarUploadSelectionPolicy`.

The reason being `AvatarSelection` requires an Email, which is not always available, specially when authenticating through OAuth. `AvatarUploadSelectionPolicy` requires `ProfileIdentifier` which can be instantiated with an Email or a Email Hash. The hash can be retrieved directly from the Profile object.

### Testing Steps

- Smoke test Image upload demo screen
- Smoke test uploading avatars in the Quick Editor demo

### DIFF
--- a/Sources/Gravatar/Options/AvatarSelection.swift
+++ b/Sources/Gravatar/Options/AvatarSelection.swift
@@ -41,6 +41,7 @@ public struct AvatarUploadSelectionPolicy: Equatable, Sendable {
     public static func selectUploadedImage(for profileID: ProfileIdentifier) -> AvatarUploadSelectionPolicy {
         .init(policy: .selectUploadedImage(for: profileID))
     }
+
     // Set the uploaded image as the avatar for the profile only if there was no other avatar previously selected.
     public static func selectUploadedImageIfNoneSelected(for profileID: ProfileIdentifier) -> AvatarUploadSelectionPolicy {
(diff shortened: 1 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/Gravatar/Options/AvatarSelection.swift
```diff
@@ -1,14 +1,14 @@
 /// Defines how to handle avatar selection after uploading a new avatar
 public enum AvatarSelection: Equatable, Sendable {
     case preserveSelection
-    case selectUploadedImage(for: Email)
-    case selectUploadedImageIfNoneSelected(for: Email)
+    case selectUploadedImage(for: ProfileIdentifier)
+    case selectUploadedImageIfNoneSelected(for: ProfileIdentifier)
```
etoledom: We should request a `ProfileIdentifier` instead of a profile email directly, since is not given that the email will be available, but the hash should.

pinarol (approved):
Looks and works good 🎉

PR #821, opened 2026-07-11
Decrypt demo secrets outside the repo via a8c-secrets

mokagio:
Part of [AINFRA-2640](https://linear.app/a8c/issue/AINFRA-2640) (folds in the [AINFRA-1549](https://linear.app/a8c/issue/AINFRA-1549) mechanism swap).

### Description

The demo app's `configure` setup decrypted `.configure-files/Secrets.swift.enc` straight into `Demo/Demo/Secrets.swift` — plaintext credentials inside the working tree, reachable by agents and one `.gitignore` typo away from being committed (the concern behind AINFRA-2138).

This swaps the decrypt mechanism to `a8c-secrets` (matching `day-one-apple` and `woocommerce-ios`) and moves the plaintext out of the checkout: `a8c-secrets decrypt` writes to `~/.a8c-secrets/<repo-id>/`, and the demo target's build phase copies it into `$(DERIVED_FILE_DIR)`, never the repo. `Demo/Demo/Secrets.tpl` stays the fallback so external contributors still build the demo.

One deliberate divergence from the `day-one-apple` reference: the build phase reads `repo-id` from `.a8c-secrets/repo-id` rather than hardcoding it, so the decrypt path can't drift on casing and degrades gracefully to the template before provisioning lands.

**Provisioning follow-ups (not in this PR, needed before real secrets flow):**

- Commit `.a8c-secrets/{repo-id,keys.pub,Secrets.swift.age}` — needs the plaintext and key generation (`a8c-secrets setup init` + `encrypt`).
- Set `A8C_SECRETS_IDENTITY` in Buildkite wherever `CONFIGURE_ENCRYPTION_KEY` was provisioned before.

Until then CI stays green: with no `.age` files, `a8c-secrets decrypt` is a no-op and the demo builds from the template.

### Testing Steps

- `make build-demo` — full Demo build-for-testing passes; the "Generate Secrets.swift" phase runs, falls back to the template (no local secrets), and `Demo/Demo/Secrets.swift` is never written into the checkout.
- CI's `🛠️ Build Demo` step exercises the same path through `make build-demo-for-distribution`.


mokagio:
Closing in favour of splitting this into two independent PRs, matching the AINFRA-2640 issue's own framing (*where* the plaintext lands vs. *what* decrypts it):

1. **Decrypt into DerivedData + external-contributor support** (AINFRA-2640) — keep `.configure`, just move its destination out of the repo checkout and copy into `$(DERIVED_FILE_DIR)` via a build phase, with a template + gitignored local override for OSS contributors. No mechanism change.
2. **a8c-secrets over `.configure`** (AINFRA-1549) — swap the decrypt mechanism; re-points the build phase's source path.

This PR combined both plus an unrelated Xcode 26.6 bump. Reworking as (1) first in a new PR. The DerivedData build-phase design validated here carries over unchanged.

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

PR #822, opened 2026-07-11
Decrypt demo secrets into DerivedData, outside the repo

mokagio:
Part of [AINFRA-2640](https://linear.app/a8c/issue/AINFRA-2640).

Adopts the new approach of decrypting secrets in the Derived Data folder, which we assume developers have configured to be stored outside of the repo (Xcode's default location), already in used in other projects [such as WooCommerce iOS](https://github.com/woocommerce/woocommerce-ios/pull/16936).

<details>
<summary>AI-generated details</summary>

### Description

`.configure` decrypted the demo credentials straight into `Demo/Demo/Secrets.swift` — plaintext inside the working tree, reachable by agents and one `.gitignore` typo from being committed (AINFRA-2138). This points the `.configure` destination outside the checkout (`~/.configure/Gravatar-SDK-iOS/secrets/Secrets.swift`) and adds a build phase that copies it into `$(DERIVED_FILE_DIR)` — the pattern `wordpress-ios` already uses. The decrypt **mechanism is unchanged** (still `.configure` / `configure_apply`); swapping it for a8c-secrets (AINFRA-1549) is a separate PR that only re-points the build phase's source path.

The compiled `Secrets.swift` comes from one of two sources: the decrypted file, or, failing that, a gitignored `Secrets.external-contributors.swift` that external contributors fill with their own credentials. That external file is seeded from the committed `Secrets.template.swift` (empty defaults) on first build, so the demo always builds without any secrets.

Because the compiled file now lives in `$(DERIVED_FILE_DIR)`, Xcode draws its `Secrets.swift` reference red and `Open Quickly` can't find it. That is expected, not a dead reference — a committed `README.md` in the new `Generated` group explains it. Deleting the reference stops the demo compiling.

### Testing Steps

- `make build-demo` — build-for-testing passes; the "Generate Secrets.swift" phase runs, seeds the gitignored external file from the template (no local secrets), and `Demo/Demo/Secrets.swift` is never written into the checkout.
- Internal contributors: `make setup-secrets` decrypts to `~/.configure/…`; the next build picks it up.
- CI's `🛠️ Build Demo` exercises the same path via `make build-demo-for-distribution` (unchanged `configure_apply` decrypt).

</details>


Scripts/build-phases/generate-secrets.xcfilelist
```diff
@@ -0,0 +1,5 @@
+# Input files for the "Generate Secrets.swift" build phase. Declaring the
+# sources lets Xcode's build system skip the phase when nothing changed.
+${SRCROOT}/Demo/Secrets.template.swift
+${SRCROOT}/Demo/Secrets.external-contributors.swift
+${HOME}/.configure/Gravatar-SDK-iOS/secrets/Secrets.swift
```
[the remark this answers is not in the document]
mokagio: `${}` works also.

Demo/Demo/Secrets.template.swift
```diff
@@ -0,0 +1,15 @@
+// Template for the demo app's secrets. Do not edit — the build phase copies this
+// into DerivedData (or, with real credentials, materializes them there).
```
mokagio:
```suggestion
// Template for the app's secrets.
//
// Only edit this file if `Secrets` needs to change in the app.
```

Demo/Demo/Secrets.template.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+// Internal contributors: run `make setup-secrets` (or `bundle exec fastlane run
+// configure_apply`) to decrypt the real credentials, which land outside the repo
+// under ~/.configure. External contributors: the first build copies this file to
+// the gitignored `Demo/Demo/Secrets.external-contributors.swift`; put your own
+// credentials from https://gravatar.com/developers/applications there. Otherwise
+// the demo builds with the empty defaults below.
```
mokagio:
```suggestion
// External contributors: The first build copies this file to the .gitignored `Demo/Demo/Secrets.external-contributors.swift`.
// Edit that file with your own credentials from https://gravatar.com/developers/applications.
// Otherwise the app will build with the empty defaults below.
//
// Internal contributors: Run `make setup-secrets` before building to decrypt the first-party secrets instead.
```

Scripts/build-phases/generate-secrets.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Materialize the demo app's Secrets.swift into DerivedData so the decrypted
+# credentials never land in the repo checkout (AINFRA-2640). The compiled file
+# comes from one of two sources, in order:
```
mokagio:
```suggestion
# Materialize the demo app's Secrets.swift into DerivedData so the decrypted
# credentials never land in the repo checkout.
#
# The compiled file comes from one of two sources, in order:
```

Scripts/build-phases/generate-secrets.sh
```diff
@@ -0,0 +1,41 @@
...
+#
+#   1. ~/.configure/Gravatar-SDK-iOS/secrets/Secrets.swift — internal
+#      contributors, decrypted by `configure_apply` outside the repo.
```
mokagio:
```suggestion
#   1. ${OUT_OF_REPO_SECRETS_ROOT}/Secrets.swift — internal
#      contributors, decrypted by `configure_apply` outside the repo.
```

Scripts/build-phases/generate-secrets.sh
```diff
@@ -0,0 +1,41 @@
...
+#   2. Demo/Demo/Secrets.external-contributors.swift — gitignored, so external
+#      contributors can paste their own credentials into a clearly named file
+#      that can't be committed. On first build it is seeded from the committed
+#      Demo/Demo/Secrets.template.swift (nil/empty defaults), so the demo always
+#      builds without any secrets.
```
mokagio:
```suggestion
#   2. Demo/Demo/Secrets.external-contributors.swift — gitignored, so external
#      contributors can paste their own credentials into a clearly named file
#      that can't be committed. On first build it is seeded from the committed
#      Demo/Demo/Secrets.template.swift.
```

Scripts/build-phases/generate-secrets.sh
```diff
@@ -0,0 +1,41 @@
...
+
+if [ -f "$SECRETS_FILE" ]; then
+    echo "Applying demo secrets from ${SECRETS_FILE}"
+    cp "$SECRETS_FILE" "$DESTINATION"
+    exit 0
+fi
+
+# The decrypted secret is absent — warn every time, even when a fallback exists,
+# so internal contributors notice a missing/undecrypted secret.
+echo "warning: decrypted demo secrets not found under ~/.configure. Internal contributors: run 'make setup-secrets'."
```
mokagio:
```suggestion
echo "warning: decrypted secrets not found under ~/.configure. Internal contributors: run 'make setup-secrets'."
```

Scripts/build-phases/generate-secrets.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Seed the external-contributors file from the template on first build. It is
+# gitignored, so edits persist across builds and can't be committed.
```
mokagio:
```suggestion
# Seed the external-contributors file from the template on first build.
```

Sources/GravatarUI/GravatarUI.docc/GettingStarted.md
```diff
@@ -38,7 +38,7 @@ While some features are ready to be tested out of the box, others require additi
 2. To have access to the Quick Editor through OAuth, set the following `Callback URI`:
   - `https://gravatar.com/iosdemo/oauth/callback` 
 3. Copy the `API key`, `Client ID`, and `Callback URI` to be pasted on the demo app.
-4. After building the Demo app for the first time, a file `Secrets.swift` is going to be automatically created. Fill up the constants of this file with the secrets from the previous step.
+4. Internal contributors: run `make setup-secrets` to decrypt the demo credentials (stored outside the repo under `~/.configure`). External contributors: build the demo once — it seeds a gitignored `Demo/Demo/Secrets.external-contributors.swift` from the template — then fill in the values from the previous step there.
```
mokagio:
```suggestion
4. Internal contributors: run `make setup-secrets` to decrypt the credentials outside the repo. External contributors: build the app once and it will seed a gitignored `Demo/Demo/Secrets.external-contributors.swift` from a template, then fill in the values from the previous step there.
```

.gitignore
```diff
@@ -41,4 +41,8 @@ vendor/swiftgen
 
 # Other
 openapi/GravatarOpenAPIClient/Sources/
+Demo/Demo/Secrets.external-contributors.swift
+# Precaution: drop this once pre-2026-07 checkouts are out of circulation.
+# Nothing writes this path anymore — before AINFRA-2640 `.configure` decrypted
+# plaintext straight here, so an old checkout may still hold a stale copy.
```
mokagio:
```suggestion
# See AINFRA-2640.
```

Demo/Demo/Generated/README.md
```diff
@@ -0,0 +1,9 @@
+# Generated Sources
+
+`Secrets.swift` is shown in red in Xcode and `Open Quickly` cannot find it.
+That is expected: it does not exist in the repository.
+
+The `Generate Secrets.swift` build phase writes it into the demo target's `$(DERIVED_FILE_DIR)` on every build, from the decrypted secrets under `~/.configure/Gravatar-SDK-iOS/secrets`, or, failing those, from `Secrets.external-contributors.swift` (seeded from the committed `Secrets.template.swift` on first build).
```
AliSoftware: I wonder if it wouldn't be wiser to remove the `~/.configure/…` path to avoid leaking that path to AI agents that would read this `README.md` and might be tempted to read it (and even if they might ask for permission first, we all know how easy it is to approve AI prompts without thinking twice), and replace it with something like `from the secrets decrypted by \`configure_apply\`` instead?

(I guess this could be addressed later, once we migrate this repo to `a8c-secrets` and will have to revisit this paragraph anyway)

Scripts/build-phases/generate-secrets.xcfilelist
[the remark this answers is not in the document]
AliSoftware: It always annoyed me that Xcode used `$(…)` while shell uses `${…}`, why make up a new syntax when a commonly known and widespread one already exists… 🤷‍♂️  And that has also always made me wonder if there was any subtle behavior difference between `$(…)` vs `${…}` in Xcode…

Scripts/build-phases/generate-secrets.sh
```diff
@@ -0,0 +1,40 @@
...
+fi
+
+# The decrypted secret is absent — warn every time, even when a fallback exists,
+# so internal contributors notice a missing/undecrypted secret.
+echo "warning: decrypted secrets not found under ~/.configure. Internal contributors: run 'make setup-secrets'."
+
+# Seed the external-contributors file from the template on first build.
+if [ ! -f "$EXTERNAL" ]; then
+    echo "Seeding ${EXTERNAL} from the template — external contributors: add your own credentials there."
+    cp "$TEMPLATE" "$EXTERNAL"
```
AliSoftware: I think this might not be visible enough, not only because it does a plain `echo` without `warning:` or `error:` that will thus just be lost in the logs that nobody looks at, but also because the build will succeed anyway.

I'd suggest 2 possible solutions instead:
1. Either `echo "error: …"` and `exit 1` in this case, to make the first build fail and make this error / suggestion more visible and increase the nudge for external contributors to fill the template with their secrets before retrying the build. The subsequent builds will not show the error nor fail due to the `exit 1` because the template would have already been created (with empty values) but at least the nudge on first build couldn't be missed even if they didn't read the README
2. Or simply remove the `cp "$TEMPLATE" "$EXTERNAL"` from here and error until the user has copied the template and filled their secrets, i.e. `echo "error: If you are an external contributor, copy the "$TEMPLATE" file to "$EXTERNAL", fill in your own secrets, and try again"` + `exit 1`

The more I think about it the more I feel like option 2 is more explicit to ensure external contributors wouldn't risk skipping / missing that step.

```bash
# Internal Contributor: the decrypted secrets is present, copy it
if [ -f "$SECRETS_FILE" ]; then
    echo "Applying demo secrets from ${SECRETS_FILE}"
    cp "$SECRETS_FILE" "$DESTINATION"
    exit 0
fi

# External Contributor: the manually-filled external-contributor secrets is present, copy it
if [ -f "$EXTERNAL" ]; then
  cp "$EXTERNAL" "$DESTINATION"
  exit 0
fi

# Neither internal nor external secret file exists — error with instructions
echo "error: Decrypted secrets not found. If you're an internal contributors from Automattic, run 'make setup-secrets' first. If you are an external contributor, copy the file from '${TEMPLATE}' to '${EXTERNAL}' and add your own credentials into it."
exit 1
```

Scripts/build-phases/generate-secrets.sh
```diff
@@ -0,0 +1,40 @@
...
+#      contributors, decrypted by `configure_apply` outside the repo.
+#   2. Demo/Demo/Secrets.external-contributors.swift — gitignored, so external
+#      contributors can paste their own credentials into a clearly named file
+#      that can't be committed. On first build it is seeded from the committed
+#      Demo/Demo/Secrets.template.swift.
+
+SECRETS_FILE="${HOME}/.configure/Gravatar-SDK-iOS/secrets/Secrets.swift"
+TEMPLATE="${SRCROOT}/Demo/Secrets.template.swift"
+EXTERNAL="${SRCROOT}/Demo/Secrets.external-contributors.swift"
+DESTINATION="${SCRIPT_OUTPUT_FILE_0}"
```
AliSoftware, line 19: I have not tested this, but I wonder, does this resolve correctly even when we're using `.xcfilelist` like here—as opposed to filling the individual output files in the build phase? Or do we need to use `${SCRIPT_OUTPUT_FILE_LIST_0}` instead?

See https://stackoverflow.com/q/53771040 — though this answer might be outdated, maybe this behavior changed in more recent Xcode versions and `SCRIPT_OUTPUT_FILE_*` are now filled even by values coming from `.xcfilelist` files?

AliSoftware (commented):
Left some suggestions to improve the build phase for better signaling (especially to external contributors)

Demo/Demo/Generated/README.md
mokagio: > (I guess this could be addressed later, once we migrate this repo to a8c-secrets and will have to revisit this paragraph anyway)

I'm leaning toward this. _Later is never_, but... this is a low activity repo, meaning even if we have the line it's unlikely to be read by AI agents running on our devs machines, and we are focusing on migrating to `a8c-secrets`, meaning _later_ is likely to be _soon_.

Scripts/build-phases/generate-secrets.sh
mokagio: I like option 2, too.

Scripts/build-phases/generate-secrets.sh
mokagio: > when we're using `.xcfilelist` like here

This PR uses `.xcfilelist` only for the input files. See diff above in `project.pbxproj` with values in `inputFileListPaths` and `outputPaths`, with `inputPaths` and `outputFileListPaths` as empty lists.

But, regardless, from reading [the docs](https://developer.apple.com/documentation/xcode/running-custom-scripts-during-a-build) I'd say that `SCRIPT_OUTPUT_FILE_[#]` is the correct option for `outputPaths` and/or `outputFileListPaths`:


SCRIPT_OUTPUT_FILE_COUNT | The total number of files described as script outputs.
-- | --
SCRIPT_OUTPUT_FILE_[#] | Environment variables that contain the paths to the script’s output files. Xcode creates an environment variable for each output file, starting with SCRIPT_OUTPUT_FILE_0 and increasing the number value sequentially for each subsequent file.
SCRIPT_OUTPUT_FILE_LIST_COUNT | The total number of file lists described as script outputs.
SCRIPT_OUTPUT_FILE_LIST_[#] | Environment variables that contain the paths to the script’s output file lists. Xcode creates an environment variable for each output file list, starting with SCRIPT_OUTPUT_FILE_LIST_0 and increasing the number value sequentially for each subsequent file list.

Scripts/build-phases/generate-secrets.sh
mokagio: Implemented in 7eb5624b3830795419fbb29a5ac68d1848f15e36

Scripts/build-phases/generate-secrets.sh
AliSoftware: > This PR uses .xcfilelist only for the input files. See diff above in project.pbxproj with values in inputFileListPaths and outputPaths, with inputPaths and outputFileListPaths as empty lists.

Ah! I missed that subtlety, good point 👍 

> But, regardless, from reading [the docs](https://developer.apple.com/documentation/xcode/running-custom-scripts-during-a-build) I'd say that SCRIPT_OUTPUT_FILE_[#] is the correct option for outputPaths and/or outputFileListPaths:

Tbh I'm not convinced that this table from the docs nor the docs paragraphs is settling the question clearly. I still think that the wording they use is still ambiguous as to if `SCRIPT_OUTPUT_FILE_[#]` includes files set via `.xcfilelist` too vs only paths to the scripts' output files set in "Output Files" but not "Output Files Lists"…
It depends if in their wording of "output files" they use for `SCRIPT_OUTPUT_FILE_[#]` documentation means "the aggregated list of all output files of the script regardless of how that list was built internally—via aggregating "Output Files" + the individual paths listed in all the "Output Files Lists" files—or if it just means "the files set in "Output Files" section", and I think their wording leave the door open to interpret it either way…

Anyway, this won't impact us here since as you mention even if the script indeed uses `SCRIPT_OUTPUT_FILE_0`, it happens that unlike the case of input files, we do use "Output Files" not "Output Files Lists" for this script build phase anyway.

AliSoftware (approved):

